PR #46, opened 2020-10-23
Import typography

six7:


phantasmagoriadigital:
LOVE FIGMA TOKENS! Was trying to add letter spacing (and similar things) and bumped into this, seems to have been on hold for a while and would be really helpful if we could define letter spacings as well. Would be interested to help with the project as well.

six7:
Didnt have too much time on this PR, its basically working but there are some small hickups when it comes to user expectation when running typography import.

Will try to squeeze a basic working state out of it just so we can move on with this, we can always improve on it once released 👍

PR #55, opened 2020-11-13
JSON Folder Edit Feature

MarcelloPaiva:


six7:
Thanks again for the PR @MarcelloPaiva - merging this! One thing I was wondering if it would make sense was if we should maybe remove the key from the tokens being shown, so that the user cannot break the code by e.g. removing the trailing comma at the end of the block:

so instead of showing

`spacing: {
  xs: 4,
}`

we could show
`xs: 4,`

MarcelloPaiva:
I see, yeah I think that could work well. We would just have to change up some of the regular expressions that I created

six7:
noticed a bug that prevents me from moving on with this @MarcelloPaiva - could you maybe try your luck on it?

if a token has nested tokens like in this example below it will only match to the first `}` and stop there:

`colors: {
  gray: {
    '100': '#ccc'
  },
 white: '#fff#
}`

i realize this PR is already merged, but basically what would help me would be the regex that solves this scenario (note it could be nested multiple levels deep). i'm the worst at regex 😅

PR #112, opened 2021-04-17
Add boxShadow Fixes #29

### LINKED ISSUE #29: Add shadow tokens
Add a box-shadow token.

For the start a single value could suffice, but in the future it would be neat if we could combine multiple shadows

scttdavs:
Thanks for this plugin! Our team is starting to use it and is really excited about it. The one thing it is missing from what we'd like is just a box-shadow token. So here is my shot at it 😀 Let me know if I've missing anything. I'm open to naming also, just went with boxShadow for now.

image

### DIFF
--- a/src/config/default.json
+++ b/src/config/default.json
@@ -27,11 +27,13 @@
     },
     "boxShadow": {
         "dark": {
-            "x": 5,
-            "y": 5,
-            "spread": 3,
-            "color": "rgba(0, 0, 0, 0.5)",
-            "blur": 5
+            "value": {
+                "x": 5,
+                "y": 5,
+                "spread": 3,
+                "color": "rgba(0, 0, 0, 0.5)",
+                "blur": 5
+            }
         }
     },
     "colors": {
--- a/src/config/default.json
+++ b/src/config/default.json
@@ -25,6 +25,9 @@
         "md": 4,
         "lg": 8
     },
+    "boxShadow": {
+        "none": "none"
+    },
     "colors": {
         "gray": {
             "100": "#f7fafc",
--- a/src/config/default.json
+++ b/src/config/default.json
@@ -26,7 +26,13 @@
         "lg": 8
     },
     "boxShadow": {
-        "none": "none"
+        "dark": {
+            "x": 5,
+            "y": 5,
+            "spread": 3,
+            "color": "rgba(0, 0, 0, 0.5)",
+            "blur": 5
+        }
     },
     "colors": {
         "gray": {
--- a/src/config/tokens.js
+++ b/src/config/tokens.js
@@ -11,6 +11,7 @@ export default {
     colors: ['fill', 'border'],
     opacity: ['opacity'],
     borderWidth: ['borderWidth'],
+    boxShadow: ['boxShadow'],
     typography: ['typography'],
     lineHeights: ['lineHeights'],
     fontSizes: ['fontSizes'],
--- a/src/plugin/updateNode.ts
+++ b/src/plugin/updateNode.ts
@@ -30,6 +30,13 @@ export default async function setValuesOnNode(node, values, data) {
         }
     }
 
+    // BOX SHADOW
+    if (values.boxShadow) {
+        if (typeof node.boxShadow !== 'undefined') {
+            node.boxShadow = values.boxShadow
+        }
+    }
+
     // BORDER WIDTH
     if (values.borderWidth) {
         // Has to be larger than 0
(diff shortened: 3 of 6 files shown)


src/config/tokens.js
```diff
@@ -11,6 +11,7 @@ export default {
     colors: ['fill', 'border'],
     opacity: ['opacity'],
     borderWidth: ['borderWidth'],
+    boxShadow: ['boxShadow'],
```
scttdavs, line 14: I'm not exactly sure what this file is for, but I think I probably did this wrong.

src/plugin/updateNode.ts
```diff
@@ -30,6 +30,31 @@ export default async function setValuesOnNode(node, values, data) {
...
+                type: 'DROP_SHADOW',
+                visible: true,
+                blendMode: 'NORMAL',
+                color: {r, g, b, a: opacity},
+                offset: {x: Number(x), y: Number(y)},
+                radius: Number(blur),
+                spread: Number(spread),
+            };
+
+            effects.push(effect);
```
six7, line 53: I'd rather replace the whole effects array with the new effect. Because right now this leads to a bug: any time someone updates tokens, a new boxShadow effect is pushed to all tokens that contain a reference to this token.

src/config/default.json
```diff
@@ -25,6 +25,15 @@
...
     },
+    "boxShadow": {
+        "dark": {
+            "x": 5,
+            "y": 5,
+            "spread": 3,
+            "color": "rgba(0, 0, 0, 0.5)",
+            "blur": 5
+        }
+    },
```
six7, line 36: This will lead to the UI showing 4 different tokens (the plugin basically checks for a `value` prop in a token, if it finds that, it is understood to be a single token rather than multiple). An exception are typography tokens, which have some logic involved due to wanting to support old no-value tokens. But I'd say for new tokens like `boxShadow` we can force the new syntax.

So, adding the `value` prop will make sure that the token is displayed as a single token.

```suggestion
    "boxShadow": {
        "dark": {
          "value": {
            "x": 5,
            "y": 5,
            "spread": 3,
            "color": "rgba(0, 0, 0, 0.5)",
            "blur": 5
          }
        }
    },
```

src/config/tokens.js
six7: You're right, this file is not needed anymore (the properties file is need to tell the plugin what keys to look at). Will remove in a future update.

six7 (changes requested):
Awesome work @scttdavs ! Thanks for this.

Some things that I'm still unsure of, but I think we can solve these in another issue:
- Would it make sense to allow entering box-shadows in a css-like syntax? e.g. `0 0 4px -4px rgba(0,0,0,0.5)`? I really like that users are able to enter different values for each property. And if we're looking to be conform with style dictionary I think that can be solved by a transformer later.
- I think it would be great if users could create and update effect styles from within the plugin, similar to how we do it for color and typography tokens. I realize this is a bigger effort (as we don't have effect style tokens yet) so I'd be fine with moving that functionality to it's own issue.

Once you resolved the two comments I'd say we'd be good to go! 🎉

src/plugin/updateNode.ts
scttdavs: good catch! I filtered out the `DROP_SHADOW` effect instead, that way any other effects that happen to be applied will not be lost. Seemed a little safer that way.

scttdavs:
> Would it make sense to allow entering box-shadows in a css-like syntax? e.g. 0 0 4px -4px rgba(0,0,0,0.5)? I really like that users are able to enter different values for each property. And if we're looking to be conform with style dictionary I think that can be solved by a transformer later.
I think it would be great if users could create and update effect styles from within the plugin, similar to how we do it for color and typography tokens. I realize this is a bigger effort (as we don't have effect style tokens yet) so I'd be fine with moving that functionality to it's own issue.

Hmm, yes, I thought about the css-like declaration, but thought it may be too code-like for some designers, and thought keeping the Figma language and inputs that they would be used to would be simpler. I agree, that other one sounds like it could be a good improvement for a separate ticket. This PR would at least open things up and allow folks to start using boxShadows :D

src/config/default.json
scttdavs: 👍 another good catch!

six7 (approved):
Awesome, thank you for your contribution 🙏

six7:
Will merge this and include it in the next version 👍

six7:
version 51 just got released with boxShadow support!

PR #132, opened 2021-05-21
added abbility to create global palette of values

jhajduk-sage:
This code gives possibiblity to create global palette set, which can be referenced in other sets.
It happens by creating set with "palette" name. It then gives you possibiblity to reference to it by setting value to $palette.colors.blue.400

Creating such global generic set accessible from all sets allows you to create defined scale of values to use in other sets, which should help with keeping consistency.

### DIFF
--- a/src/app/components/TokenData.tsx
+++ b/src/app/components/TokenData.tsx
@@ -149,6 +149,11 @@ export default class TokenData {
                 }, [])
             );
 
+            // Allows for referencing values form 'palette' set in each other set.
+            if ('palette' in this.tokens) {
+                this.mergedTokens.palette = JSON.parse(this.tokens.palette.values);
+            }
+
             this.setAllAliases();
         }
     }


six7:
First of all thanks for the contribution @jhajduk-sage !

I do not yet see the benefit of this, users could just create a token set, call it global and then reference the colors by specifying `$colors.blue.400`, right? I'd like the plugin to be as flexible as possible (even more so in the future), so introducing a static name for a global palette seems counterintuitive.

However, I think what would be best if we allowed users to reference values from within a specific token set, so that you could create an alias to `global.colors.blue.400`, referencing a token `colors.blue.400` living in the `global` token set (however users would be able to reference any token in any token set.

jhajduk-sage:
Giving possibility to reference values from specific token set was my goal actually ;). However this what I came up with was simpliest and fastest approach ;).

However i totally agree with you. I will try to work out better, more flexible solution.

Thanks for answering! :)

PR #146, opened 2021-06-10
Next

### LINKED ISSUE #99: Add new autolayout properties (left, right, top, bottom)
We should add the following autolayout properties that Figma introduced a while ago:

https://www.figma.com/plugin-docs/api/FrameNode/#paddingleft

```
paddingLeft
paddingRight
paddingBottom
paddingTop
```

I would suggest leaving the old vertical and horizontalPadding properties in code (so that the plugin works with older tokens set) but the interface only allows setting new tokens.

### LINKED ISSUE #123: Improve performance by extending 'Update/Apply' functionality to selection / disabling auto update + apply to selection
Especially when it comes to themes, users may want to apply changes only to their current selection.

For example, a user might have a light and a dark theme. Changing themes currently triggers an update to the whole page or the whole document. Users might want to duplicate a frame, then apply the dark theme on that frame and see changes side-by-side.

Currently I plan on integrating new options:

- Apply to selection
- Apply to page (existing)
- Apply to document (existing)
---
- Update on change
- Update styles

image

Update on change: If on, this equals current behaviour of updating the page/document whenever a token gets changed. **This leads to long loading times if you're on a page with many layers!** Being able to toggle that behaviour off allows users to quickly change tokens, and then apply these tokens to their designs when they feel like it.

Update styles: Currently all changes trigger updates in styles (as long as styles are local). Being able to toggle that off would also allow for granular changes to tokens, instead of always triggering a change of styles.

### LINKED ISSUE #125: Importing outdated styles
I'm running into an issue where Figma Tokes plugin seems to keep old outdated metadata around. Even after clearing all tokens and then reimporting from Figma, some text styles are still using old values, in my case a font that isn't even available anymore, old letterSpacing, etc ... where Figma clearly has different values for these styles. Even worse, the plugin seems to be changing the actual styles after applying them through the plugin. This makes me believe the plugin is getting this data from some hidden metadata on those layers/styles. Maybe there is a way to just clean all of this cruft when using the "clear" button?

### LINKED ISSUE #137: Support for different FigmaTokens titles (minor fix)
**Nowadays I need follow the naming structure of FigmaTokens to my job works fine.**

When I'm working in Figma and try propose alternative titles for tokens, I have some problems... For example:

I have Box Shadow tokens....

If I change the name in Json file for box-shadow, the plugin doesn't work in Figma, because it doesn't understand the tokens in JSON and hidden the values._

**My suggestion is use some common variations of tokens.** 

Example: Box Permission to rename for shadow, BoxShadow, boxShadow and Boxshadow (commun cases)

### LINKED ISSUE #138: Performance issues with increasing numbers of tokens
I am experiencing wait times of around 1 minute every time i add or change something in Figma Tokens, due to the app updating. This occurs both with JSON files synced to JSONbin and when i uncheck that for local storage.

Is there anything that can be done about it? Maybe an option to update the file manually instead of after every change?

We have a substantial amount of tokens, both global ones and component specific ones which reference the global ones for their values. If we have to go for a simpler scheme it will be less valuable to us.

Hopefully something can be done?

### LINKED ISSUE #162: Pause sync to jSONbin (feature request)
Would it be possible to pause syncing to JSONbin and manually push to remote? 

Currently every change I make to a token results in a a remote update, adding to the number of requests in my JSONbin plan.

### LINKED ISSUE #167: JSON file doesn't show Alias Tokens
EXPORT ALIAS

When I reference Tokens (Alias tokens) and try to export them, I only see raw values in the JSON file.
Does the plugin supports different types of tokens? (ex: Global/Alias/...)

six7:
- Complete rewrite of state management
- Adds ability to change updateMode to manual (fixes #138)
- Allows you to create your own token naming structure by introducing token types (fixes #137)
- Allows users to change updateMode to selection, document or page (fixes #123)
- Adds import token dialog with diff between old and new values (fixes #125)
- Adds new auto layout properties (fixes #99 and #161)
- Fixes hsla preview (#158)
- Adds updateRemote option allowing you to pause sync to remote (fixes #162)
- Fixes alias output to style dictionary (fixes #167)


src/plugin/tokenHelpers.ts
```diff
@@ -0,0 +1,66 @@
...
+import {SingleTokenObject} from 'Types/tokens';
+import {mergeDeep} from './helpers';
+
+export function findAllAliases(tokens) {
+    return tokens.filter((token) => checkIfAlias(token, tokens));
+}
+
+export function reduceToValues(tokens) {
+    const reducedTokens = Object.entries(tokens).reduce((prev, group) => {
+        prev.push({[group[0]]: group[1].values});
```
WardMe, line 13:
```suggestion
        prev.push({[group[0]]: group[1]});
```
Due to the type change, this returns an error when updating json

PR #185, opened 2021-08-02
Fixing zero values

WardMe:
Currently the plugin doesn't support zero `0` values for tokens, the check `if (values.opacity)` for example skips the update because `0 === false` and then it won't update the styles.

I've updated the checks to `typeof prop !== 'undefined'` so it also works for 0 values and added an extra value check for properties that don't allow 0 values in Figma (ie. width, height, border width, …) and convert the 0 to 0.001.

This way I can set my opacity to 0 or hide elements.

What do you think?

### DIFF
--- a/src/plugin/figmaTransforms/generic.test.ts
+++ b/src/plugin/figmaTransforms/generic.test.ts
@@ -1,4 +1,4 @@
-import {convertTypographyNumberToFigma} from './generic';
+import {convertTypographyNumberToFigma, fakeZeroForFigma} from './generic';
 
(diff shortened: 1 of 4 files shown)


six7 (approved):
This is awesome, thank you for your work! 🥳

six7:
Another thing I just noticed is that for sizing this seems to not work yet, as Figma requires a value larger than 0.01 (we set it to 0.001) - I'm thinking this is a bug on Figma's end. Why would they not allow 0.001 values?

PR #190, opened 2021-08-04
Adding a token filter

WardMe:
Our token library is getting bigger and bigger because of component tokens and it was getting kind of hard to find the right tokens in the long list. So I've created a quick filter at the top of the token list to make it easier to find the right tokens.

It currently only searches for an exact string in the name of the token, but could be expanded in the future to support more options.

What do you think?


WardMe:
Example of what it does:
https://user-images.githubusercontent.com/3016302/128192632-5bb83d19-2c88-46f3-ba38-5cb11682217a.mov

six7:
I love it! Let me take a look at it tomorrow in-depth but I really like this!

six7:
I created a PR that merges some changes into yours, take a look and let me know what you think!

Also, I think it would be awesome if we could `Cmd + F` to toggle visibility of the field as well, but I think we can do that at a later point as well.

WardMe:
Looks good! Thanks!

six7:
This just got shipped in Version 57. Thank you for this! 👏

PR #225, opened 2021-08-27
Reordering boxShadow values

benhami1ton:
Moved the blur key to between y and spread. If the edit token modal is pulling this index order to generate the fields, it will allow the user to input values in the same sequence as [CSS syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#values) of "box-shadow: h-offset v-offset blur spread color;". This will also match the tab order of Figma's drop shadow values.

### DIFF
--- a/src/config/tokenTypes.js
+++ b/src/config/tokenTypes.js
@@ -72,9 +72,9 @@ export default {
             value: {
                 x: '',
                 y: '',
+                blur: '',
                 spread: '',
                 color: '',
-                blur: '',
             },
         },
     },


six7:
Thank you for this 👍 

I'd like to move to a string-based representation of box shadows in the future so that we can have multiple box shadows such as 
`["0 0 2px 4px #ccccccaf", "0 2px 1px 4px #eeeeeeaf"]` - but until then this will be a nice addition!

benhami1ton:
That'll be awesome. I've got tokens for each umbra, penumbra, and ambient shadow, having a way to merge those into one would be super helpful.

six7:
Your PR just got released with version 62. Thank you for your contribution 🙏

PR #244, opened 2021-09-07
Share tokens with other plugins

jhajduk-sage:
# Overview
This contribution introduces storing information about tokens used for specific layers in shared plugins space, so other plugins would use it.

# Motivation
Developers need to know, where to use which token. One of possibilities to solve this is to let them inspect tokens in Figma. This however creates risk of corruption of Design Tokens, for example by mistakenly changing their structure, or values.

## Inspecting token values
That's why I created this dead simple inspector plugin (https://github.com/jhajduk-sage/figma-tokens-inspector) for tokens applied with Figma Tokens plugin. To make it work, design tokens information are needed to be stored in shared 'tokens' space.

I decided to add `setSharedlugnData` next to `setPluginData` (and not instead of) to be sure, that any accidental changes to the shared plugin data won't crash Figma Tokens Plugin.

## Other plugins
This shared data can also be used in other plugins for example to generate Design Token usage documentation. Together with tokens list which is already stored in shared space, this can introduce whole lot of possibilities.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ cypress/videos
 .env
 .env.production
 package
+.idea
--- a/src/plugin/pluginData.ts
+++ b/src/plugin/pluginData.ts
@@ -101,16 +101,19 @@ export function removePluginData(nodes, key?) {
         } finally {
             if (key) {
                 node.setPluginData(key, '');
+                node.setSharedPluginData('tokens', key, '');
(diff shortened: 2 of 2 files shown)


six7:
Thank you for that, @jhajduk-sage ! I just released version 63 which comes with this.

I think this is great, as you mentioned this opens up a lot of possibilities. 🙏

PR #292, opened 2021-10-27
Deep equality check on imported typography styles

tpict:
At the moment, typography styles are always reported as having changed after clicking the "import" button:
image
This is because the equality check for token values was only implemented for strings. I've added `lodash.isEqual` as a dependency to do a deep comparison of the `value` property, whatever type it may be. The diff under "Existing tokens" will only show up if some part of the style changed.


six7:
Thanks again, merging this - will release in version 67!

PR #306, opened 2021-11-05
Box shadows

### LINKED ISSUE #281: [Tracking]: Box Shadow improvements
Box Shadows aren't properly supported yet, we should change that.

A couple of issue that are open:

- [ ] https://github.com/six7/figma-tokens/issues/163
- [x] https://github.com/six7/figma-tokens/issues/174
- [x] https://github.com/six7/figma-tokens/issues/234
- [x] https://github.com/six7/figma-tokens/issues/312

Currently we store box shadows as an object:
```
"boxShadow": {
    "dark": {
      "value": {
        "x": 5,
        "y": 5,
        "spread": 3,
        "color": "rgba(0, 0, 0, 0.5)",
        "blur": 5
      },
      "type": "boxShadow"
    }
  },
```

This doesn't really allow us to do something like multiple shadows.

Instead, I would propose to create boxShadows with an array syntax: `[5px 5px 5px 3px rgba(0, 0, 0, 0.5), 5px 5px 5px 3px rgba(0, 0, 0, 0.5)]` which would allow us to have multiple shadows. We could still pick up the values in the UI and possibly provide a nicer experience.

Also, this should allow us to import and export shadow styles more easily.

six7:
This is a Draft PR on Box Shadow improvements.

- [x] Pull Shadow Styles
- [x] Update Shadow Styles
- [x] Check Apply behaviour
- [x] Add support for multiple shadows
- [x] Add support for innerShadow (in addition to existing dropShadow)
- [x] Integrate prettier EditShadowTokenForm
- [x] Add cypress test for editing a box shadow token (single and multiple)
- [x] Add test for setEffectValuesOnTarget.ts
- [x] Add test for filterMatchingStyles.ts
- [x] Update tests for setValuesOnNode.ts
- [x] Update tests for tokenHelpers.ts
- [x] Update tests for updateStyles.ts 

Fixes #281


acstll:
May I try and help with the remaining "Add tests" task? I see there's tests only for the files in the `plugin` and `utils` folder, is that so?

six7:
> May I try and help with the remaining "Add tests" task? I see there's tests only for the files in the `plugin` and `utils` folder, is that so?

Oh sure, that would be great! I just pushed another commit.

Right, not everything has coverage right now - tests live (except cypress) right next to their function files. Ideally we'd have a cypress test for editing a box shadow token and adding multiple.

I'd love to release this on Friday/weekend - if you dont manage to get to it by then now worries - I could take care of it as well by then 👍

acstll:
Amazing, thanks for all the info 👍

acstll:
I'm not gonna be able to get this done by tomorrow, so I prefer not to block your plans (thanks for the clarity!). I have spent some time getting familiar with the codebase (and also learning cypress) but I need a bit more time to be able to confidently contribute :)

I think I found a bug, while testing the feature manually: when you have multiple values for a shadow token, the inputs lose focus on every keystroke. I believe this is caused by the `id` used for the keys in the inputs array being generated on every render/value update:

https://github.com/six7/figma-tokens/blob/814584ec8ee6ddd76e502a9eba8d0cf9b0066d4e/src/app/components/BoxShadowInput.tsx#L206

To solve this, I would temporarily add the generated id to the value object (in `ShadowTokenSingleValue`?), to make sure it stays the same on every update. And then remove it before persisting the data. But I'm not sure this is the best approach. (I couldn't figure out how or where the data gets persisted).

six7:
> I'm not gonna be able to get this done by tomorrow, so I prefer not to block your plans (thanks for the clarity!). I have spent some time getting familiar with the codebase (and also learning cypress) but I need a bit more time to be able to confidently contribute :)
> 
> I think I found a bug, while testing the feature manually: when you have multiple values for a shadow token, the inputs lose focus on every keystroke. I believe this is caused by the `id` used for the keys in the inputs array being generated on every render/value update:
> 
> https://github.com/six7/figma-tokens/blob/814584ec8ee6ddd76e502a9eba8d0cf9b0066d4e/src/app/components/BoxShadowInput.tsx#L206
> 
> To solve this, I would temporarily add the generated id to the value object (in `ShadowTokenSingleValue`?), to make sure it stays the same on every update. And then remove it before persisting the data. But I'm not sure this is the best approach. (I couldn't figure out how or where the data gets persisted).

Awesome, thank you! I'll get to it. I think this is due to me removing `currentEditToken` which would've kept data changed locally. Thanks for trying it out 🙏

PR #332, opened 2021-11-23
Preserve parent during conversion

jdwillemse:
Resolves [!301](https://github.com/six7/figma-tokens/issues/301)

I do not know if this has further knock-on effects so please advise.

### DIFF
--- a/src/utils/convertTokensObjectToResolved.test.ts
+++ b/src/utils/convertTokensObjectToResolved.test.ts
@@ -44,36 +44,42 @@ describe('convertTokensObjectToResolved', () => {
         };
 
         expect(convertTokensObjectToResolved(tokens)).toEqual({
-            colors: {
-                red: {
-                    value: '#ff0000',
-                    type: 'color',
-                },
-                blue: {
-                    value: '#0000ff',
-                    type: 'color',
-                },
-                primary: {
-                    value: '#ff0000',
-                    type: 'color',
+            theme: {
+                colors: {
+                    primary: {
+                        value: '#ff0000',
+                        type: 'color',
+                    },
                 },
             },
-            sizing: {
-                base: {
-                    value: 2,
-                    type: 'sizing',
+            options: {
+                colors: {
+                    red: {
+                        value: '#ff0000',
+                        type: 'color',
+                    },
+                    blue: {
+                        value: '#0000ff',
+                        type: 'color',
+                    },
                 },
-                scale: {
-                    value: 1.5,
-                    type: 'sizing',
-                },
-                small: {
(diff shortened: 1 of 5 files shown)


six7:
Thank you for this! I just checked, the function responsible for the exported output is `formatTokens` - we should make the change there. `convertTokensObjectToResolved` and `convertTokensToGroupedObject` are both responsible for the token-transformer package itself, which we don't want to touch for this issue.

```
const path = `${[activeTokenSet]}.${name}`;
```

Also, I think for the `parseTokenValues` file we shouldn't make that change, as this one's about the root level (like global: ... or light: ...) - and we only want to allow object or array for now.

Let me know if you want to make those changes or if I should - I could start from your basis!

jdwillemse:
I'm not sure if I'm misunderstanding you. 

As I understand it the problem lies in the transformation plugin, not in the export plugin. The export has the correct root keys but during conversion, these are removed. `formatTokens` does not seem to be used in the conversion plugin.

The root keys are used as theme keys and so without them, there are conflicting styles that have the same names.

Can you elaborate on how you see the solution?

six7:
> I'm not sure if I'm misunderstanding you.
> 
> As I understand it the problem lies in the transformation plugin, not in the export plugin. The export has the correct root keys but during conversion, these are removed. `formatTokens` does not seem to be used in the conversion plugin.
> 
> The root keys are used as theme keys and so without them, there are conflicting styles that have the same names.
> 
> Can you elaborate on how you see the solution?

The original issue was more about the Export function within the plugin not preserving the parent key, sorry if I explained incorrectly :(

The way I see it:
- Plugin should export the whole token object, that includes the root key, as
- the root keys are required in the token-transformer package which uses it to determine what themes the user wants to include
- The token-transformer package itself removes those root keys (as it should, but possibly as you mentioned in your issue comment maybe as a parameter so users could include those root keys)

That's why I think to solve this issue we should just keep the parent in the formatTokens function. If we were to include a parameter for the token-transformer package to preserve parents then this would be the way to go.

I hope that makes sense! And again, sorry for not being more transparent in the first place.

jdwillemse:
I think this resolves the issue mentioned in the original ticket. It does not solve the token-transform problem though. I've created a new ticket for that issue https://github.com/six7/figma-tokens/issues/350

six7:
> 

Totally, that's another issue that we'd need to solve. Ideally (as a followup) we'd export all token sets, not just the active one.

Thank you for this! I just edited the test file and removed the first part of the token name to make the test a bit easier to read (my fault in the beginning).

Will merge 🙏

PR #381, opened 2021-12-19
feat: Performance V0 - use nodemanager

LiamMartens:


LiamMartens:
@six7 quite a few changes - also because I started typing any untyped methods and data. Still a WIP and might not be 100% correct everywhere - but this doesn't have an effect on the runtime code anyways (but will make the life of any future developer a lot easier :smile:)

Aside from this the real relevant changes are in the NodeManager and in the plugin/node.ts and plugin/pluginData.ts files - where the updating of nodes happens.

The plugin does not hang at all anymore - and generally when applying new tokens/updating tokens the action run almost instantly (once the NodeManager has the map ready in the background)

I am interested in running a test on a large document with many tokens however - do you have any example figma doc available by any chance?

six7:
Sweeet! Lots of changes, wondering if it would make sense to separate work? I'll pause my work on other items as we'd encounter merge hell otherwise.

I did notice performance is now pretty bad when just using the plugin in large documents with many layers (e.g. Ant Design Open Source) without even using the Apply function (probably because of the background task?)

Also it seems Inspect is broken as it doesn't show any tokens even on layers that have tokens applied.

phantasmagoriadigital:
Apart from the above issues, I have been trying for quite a while.
- I run into a funny issue where tokens applied by the current version are
not being updated by the plugin from this PR.
- On multiple occasions, the plugin got stuck altogether, which is probably
a WIP issue. Figma needed to be restarted
- Occasionally, the busy loader just keeps going on, but file remains
workable
- performance strangely is not consistent, same single page file, when
clearing all styles and reapplying through the plugin takes between
15(fastest) and 28(slowest) seconds on the same operation.
- Been recording on Loom, I can share those privately if you'd like to
review.

LiamMartens:
@phantasmagoriadigital definitely helpful to have some the loom if you have it -- also to have some example design files. @six7  You mentioned Ant Design OS - they use tokens in the file or?

LiamMartens:
@phantasmagoriadigital thanks I got it
@six7 I can see what you mean - I need to have a think about working with files this big on document level..

phantasmagoriadigital:
@LiamMartens - For us also, many files are as big as Ant Design. We are now
planning to move most of our Design Systems to a 'one file per component'
structure, but the one thing currently holding back is the issue with
remote styles not working with Figma Tokens. So currently we maintain the
colors (styles) and components in one library, which isn't ideal anyway.
But many of our users have HUGE files with MASSIVE components where
500-1000 variants are no exception.

six7:
> @phantasmagoriadigital definitely helpful to have some the loom if you have it -- also to have some example design files. @six7 You mentioned Ant Design OS - they use tokens in the file or?

They're not using tokens in their files but they have a huge amount of layers so I thought this would be a good base to test. I do see some perf impact in these. Generally UI Kit authors tend to have just 1 file as you can then easily publish to Community (and distribution is easier)

src/app/components/TokensBottomBar.tsx
```diff
@@ -4,35 +4,56 @@ import { track } from '@/utils/analytics';
...
+            Create Styles
+          </Button>
+          <Button variant="primary" onClick={handleUpdate}>
+            {confirmTimeout ? 'Confirm' : 'Update'}
+          </Button>
+        </div>
       </div>
+      {!!confirmTimeout && (
+        <div className="p-2 rounded-sm">
+          Are you sure you want to run this action document wide? This operation can take more than 30 minutes on very large documents
```
six7, line 54: you could use the `useConfirm()` hook which will display an overlay asking the user, depending on what the user answered we could proceed

LiamMartens: Did do that in other branch - closing this for now

PR #396, opened 2022-01-12
feat: options for token-transformer & expandTypography option for style-dictionary compatibility

jeffreydelooff:
Currently the **token-transformer** CLI automatically expands typography types to:

```
"typographyName": {
  "fontFamily": {
    "value": "Roboto"
  },
  "fontSize": {
    "value": 18
  }
}
```
For style-dictionary compatibility it would be easier to have the following format (especially for custom formats):
```
"typographyName": {
  "type": "typography",
  "value": {
    "fontFamily": "Roboto",
    "fontSize": 18
  }
}
```
so we can use the **type** to group/filter typographies together.

This PR adds:
* yargs set-up for token-transformer inputs & validation
* possibility to add options for transformations (here I introduced expandTypography so users can opt-out of this auto expansion)

### DIFF
--- a/src/utils/convertTokensObjectToResolved.ts
+++ b/src/utils/convertTokensObjectToResolved.ts
@@ -8,7 +8,9 @@ export default function convertTokensObjectToResolved(
     tokens,
     usedSets = [],
     excludedSets = [],
-    options: TransformerOptions
+    options: TransformerOptions = {
+        expandTypography: true,
+    }
 ) {
     // Parse tokens into array structure
     const parsed = parseTokenValues(tokens);
--- a/src/utils/convertTokensObjectToResolved.ts
+++ b/src/utils/convertTokensObjectToResolved.ts
@@ -1,9 +1,15 @@
 import {mergeTokenGroups, resolveTokenValues} from '@/plugin/tokenHelpers';
+import {TransformerOptions} from './types';
 import convertTokensToGroupedObject from './convertTokensToGroupedObject';
 import parseTokenValues from './parseTokenValues';
 
 // Takes Figma Tokens input, resolves all aliases while respecting user's theme choice and outputs an object with resolved tokens, ready to be consumed by style dictionary.
-export default function convertTokensObjectToResolved(tokens, usedSets = [], excludedSets = []) {
+export default function convertTokensObjectToResolved(
+    tokens,
+    usedSets = [],
+    excludedSets = [],
+    options: TransformerOptions
+) {
     // Parse tokens into array structure
     const parsed = parseTokenValues(tokens);
     // Merge to one giant array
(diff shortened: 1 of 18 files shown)


webpack-transform.config.js
```diff
@@ -29,6 +29,11 @@ module.exports = (env, argv) => ({
             '@types': path.resolve(__dirname, 'types'),
             '@': path.resolve(__dirname, 'src'),
         },
+        fallback: {
```
six7, line 32: This one fails for me. The plugin is still running webpack 4. A quick search seemed to indicate that resolve.fallback is part of webpack 5? Can we go without this? I could imagine an upgrade to webpack to be more than just a version bump

src/utils/convertTokensObjectToResolved.ts
```diff
@@ -1,16 +1,24 @@
...
 import parseTokenValues from './parseTokenValues';
 
 // Takes Figma Tokens input, resolves all aliases while respecting user's theme choice and outputs an object with resolved tokens, ready to be consumed by style dictionary.
-export default function convertTokensObjectToResolved(tokens, usedSets = [], excludedSets = []) {
+export default function convertTokensObjectToResolved(
+    tokens,
+    usedSets = [],
+    excludedSets = [],
+    options: TransformerOptions = {
+        expandTypography: true,
```
six7, line 12: I'm wondering if we should make the new way the default and introduce a breaking change?  I've heard quite a bit of feedback that the current behavior is not expected so I'd be 👍 for changing this to expandTypography: false

token-transformer/README.md
```diff
@@ -3,18 +3,24 @@
...
 
 Output: Filename of output
 
 Sets: Sets to be used, comma-seperated
 
-Excludes: Sets that should not be part of the export (e.g. a global color scale)
\ No newline at end of file
+Excludes: Sets that should not be part of the export (e.g. a global color scale)
+
+--expandTypography: true|false in order to opt-out of automatic expansion of typography types (default: true)
```
six7, line 26: same, let's use default: false

six7 (changes requested):
Other than those changes this looks good! 🎉 Thank you for this!

Could we get those changes in? I was able to run it by removing the `fallback` param to try it out, but I guess that's needed for tests?

webpack-transform.config.js
jeffreydelooff: Indeed, I think it already worked without this because of v4. I have removed it.

src/utils/convertTokensObjectToResolved.ts
jeffreydelooff: Yes, you are right. It feels also the recommended way to not expand it at all. 👍

six7 (approved):
Nice! LGTM! 🎉  Will release later today

avendiart:
@six7 has this been released? `0.0.16` still seems to always expand by default.

six7:
> @six7 has this been released? `0.0.16` still seems to always expand by default.

There's been a bug in this recent version that @jeffreydelooff [fixed](https://github.com/six7/figma-tokens/pull/404) today - I just released it and should work now 🙏

PR #404, opened 2022-01-17
fix: boolean check being removed by webpack

jeffreydelooff:
Webpack apparently removed the **expandTypography** check from `convertTokensToGroupedObject.js`. Not sure why, but probably because the default value was set to false.

Minified/obfuscated code before fix:
```
if(delete e.name,delete e.rawValue,delete e.internal__Parent,"typography"===e.type)
```

After fix:
```
if(delete s.name,delete s.rawValue,delete s.internal__Parent,r&&"typography"===s.type)
```


token-transformer/index.js
```diff
@@ -76,16 +76,17 @@ const transform = () => {
             expandTypography,
         };
 
-        log(`transforming tokens from input: ${input}\n`);
-        log(`using sets: ${sets.length > 0 ? sets : '[]'}\n`);
-        log(`using excludes: ${excludes.length > 0 ? excludes : '[]'}\n`);
+        log(`transforming tokens from input: ${input}`);
+        log(`using sets: ${sets.length > 0 ? sets : '[]'}`);
+        log(`using excludes: ${excludes.length > 0 ? excludes : '[]'}`);
+        log(`using options: { expandTypography: ${expandTypography}}`);
```
six7, line 82:
```suggestion
        log(`using options: { expandTypography: ${expandTypography} }`);
```

six7 (approved):
Good to merge once suggestion has been applied 🙏 

Thanks!

Issue #423, opened 2022-01-24
Export modal: Add option to "resolve values"


A143447:
**Is your feature request related to a problem? Please describe.**
We have several categories of tokens that build on each other: core, bridging and components. (Components referring to bridging and bridging referring to core).
In the Figma Tokens plugin, JSON tab, when all the categories are selected, 'All token sets' is selected and export is clicked, the tokens are exported with the referring token name rather than the actual resolved value being calculated. It would be helpful to have the option of resolving tokens to their actual value as the next step for the exported JSON is to generate iOS/Android etc specific tokens using Style Dictionary. (Style Dictionary requires that the actual value be given). Can you please assist?

This image is the exported Component (referring to bridging - that refers to Core). The expected value of the token is 8 - as per the subsequent image:
![image](https://user-images.githubusercontent.com/39042077/150704642-e57f0853-d717-4813-8a18-843dc6d54ad0.png)
![image](https://user-images.githubusercontent.com/39042077/150712891-1ac9d85a-d092-4c5f-bb6f-efae5cdf07c7.png)

**Describe the solution you'd like**
As part of the export process, a new option to resolve tokens to their actual value could be given. This functionality (not option) was available in version 52 of the plugin.

**Describe alternatives you've considered**
The only alternative that we have considered is to have a local version of the figma plugin version 52 available for the developers to use when exporting the tokens. This version allows the developers to select multiple categories of tokens and the export has the values resolved to the lowest level. 

**Additional context**
This request is related to https://github.com/six7/figma-tokens/issues/189

six7:
This shouldn't be too difficult to add, thanks for opening this!

We'd want another option in export `resolve values` - something similar to what we'd want to do for the token-transformer package.

A143447:
yes, `resolve values` sounds perfect. Thank you so much for being open to this feedback

six7:
FYI: Version 18 of token transformer now features a way to `--preserveRawValue=true` giving you a dedicated rawValue field on each token. This doesn't solve this issue yet, but should provide another way of accessing those unresolved values.

SamIam4Hyma:
@six7 Suggest to close b/c it's stale unless I'm misunderstanding

PR #455, opened 2022-02-11
Add the ability to preserve rawValue on token transform

zephraph:
This PR adds the option `--preserveRawValue` to `token-transformer` which does as it suggests... preserves the `rawValue` field in the token output. This value is very useful when theming because it preserves the context of what a given value referenced which can be useful in custom formatters.

As a use-case example, image you had this token config

```js
colors: {
  base: {
    green: {
      '500': {
        value: '#48d597',
        type: 'color',
      },
    },
  },
},
main: {
  theme: {
    accent: {
      "1": {
        value: "{base.green.500}",
        type: "color"
      }
    }
  },
  content: {
    accent: {
      default: {
        value: "{theme.accent.1}",
        type: "color"
      },
    }
  }
}
```

It's fairly clear from looking at the config that there's a relationship between `content-accent`, `theme-accent-1`, and `base-green-500`. If you're building a theming setup, preserving that relationship via css-variables maximizes the flexibility you have to make broad changes by touching the fewest points. That's essentially what I wanted to do.

### DIFF
--- a/src/utils/__snapshots__/convertTokensObjectToResolved.test.ts.snap
+++ b/src/utils/__snapshots__/convertTokensObjectToResolved.test.ts.snap
@@ -68,6 +68,28 @@ Object {
 }
 `;
 
+exports[`convertTokensObjectToResolved preserves rawValue when requested 1`] = `
+Object {
+  "colors": Object {
+    "background": Object {
+      "rawValue": "$colors.black",
+      "type": "color",
+      "value": "#000000",
+    },
+    "black": Object {
+      "rawValue": "#000000",
+      "type": "color",
+      "value": "#000000",
+    },
+    "white": Object {
+      "rawValue": "#ffffff",
+      "type": "color",
+      "value": "#ffffff",
+    },
+  },
+}
+`;
+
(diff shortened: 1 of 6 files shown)


six7:
🔥 This is great! Will take a closer look over the weekend, thank you for this 🙏

zephraph:
An alternative to this would be to just not delete the `rawValue` in the first place. It'd add bloat to folks' existing configs if they check those in, but it could be worth having generally. I'm more than happy w/ this approach though.

six7:
Could you update the tests in `src/utils/convertTokensObjectToResolved.test.ts` to match changes and include a scenario for when the setting is on?

Other than that this is awesome, thank you! 🙏 

I wanted to work on something similar: `--resolveReferences=false` which would then insert raw values in the `value` field (and fix #423), but I think both of these options are useful, especially as `--preserveRawValues` could be used to enhance token descriptions with the raw values or something else, really up to the user what they do with it 👍

six7 (approved):
Nice, thank you! Will release this later today. 🎉

PR #484, opened 2022-02-23
Add throwErrorWhenNotResolved option to the token-transformer

borbit:
Designers frequently make changes to the design tokens and occasionally make mistakes in tokens' values. As a result, we end up with "failed to resolve" tokens that can easily be missed. I didn't find an existing solution for the problem and put up this PR as a proposal. These changes will give a possibility to spot a problem on the stage of transforming tokens.


six7 (changes requested):
Again, thank you for this! 🎉 

I noticed errors trying to resolve values for `typography` and `boxShadow` are currently not picked up, but those are not even being picked up by `failedToResolve` in the plugin itself. We should fix it at that level, for this feature it's fine 👍 

A test would be good in `src/utils/convertTokensObjectToResolved.test.ts`

borbit:
Our country 🇺🇦 is under attack by russia now, so I'll update my PR a bit later. Thank you for understanding @six7 ❤️

PR #491, opened 2022-02-27
fix: added mask input field feature so that the user can control the …

srinivas0210:
…input field accessibility


src/app/components/Input.tsx
```diff
@@ -40,35 +54,52 @@ const Input: React.FC<Props> = ({
...
+          min={min}
+          max={max}
+          step={step}
+          data-custom={custom}
+          ref={inputRef}
+          placeholder={placeholder}
+        />
+
+        {isMasked && (
+          <StyledIcon onClick={handleVisibility}>{show ? <IconVisibility /> : <IconVisibilityOff />}</StyledIcon>
```
six7, line 98: The only concern I have here is that this is not accessible. Can you wrap it in a `<button>` so users can tab into it?

six7 (changes requested):
Other than the comment this is great! ❤️ If you could wrap it in a button (while keeping visuals as is) we'd be good to go 👍

six7 (approved):
🎉

PR #495, opened 2022-02-28
Add resolveTokens to token-transformer

six7:


jakobe:
I guess this can be closed now that #639 has been merged

PR #500, opened 2022-03-03
fix: setting showEmptyGroups is stored in clientstorage

### LINKED ISSUE #469: Hide/show empty groups, remember last setting
### Discussed in https://github.com/six7/figma-tokens/discussions/454

<div type='discussions-op-text'>

<sup>Originally posted by **kbradley** February 11, 2022</sup>
I'm finding that every time I close the Plugin dialog and then reopen later it doesn't remember my last setting to "Hide empty groups" so I have to set it again each time to eliminate visual clutter. It would be great if the plugin would remember the last setting you used for this.</div>

srinivas0210:
fixes #469


src/utils/uiSettings.ts
```diff
@@ -45,7 +48,8 @@ export async function getUISettings(): Promise<SavedSettings> {
         ignoreFirstPartForStyles,
         inspectDeep,
       };
-      notifyUISettings(settings);
+
+      notifyUISettings(settings, showEmptyGroups);
```
six7, line 52: Could we move showEmptyGroups into the settings const?

src/utils/uiSettings.ts
```diff
@@ -4,7 +4,10 @@ import { notifyUISettings, notifyUI, SavedSettings } from '@/plugin/notifiers';
 // update credentials
 export async function updateUISettings(uiSettings) {
   try {
-    await figma.clientStorage.setAsync('uiSettings', JSON.stringify(uiSettings));
```
six7, line 7: Is there a reason for this change? Just trying to understand 👍

src/plugin/notifiers.ts
```diff
@@ -62,16 +62,19 @@ export type SavedSettings = {
...
+    width,
+    height,
+    updateMode,
+    updateOnChange,
+    updateStyles,
+    ignoreFirstPartForStyles,
+    updateRemote = true,
+    inspectDeep,
+  }: SavedSettings,
+  showEmptyGroups: boolean,
```
six7, line 76: Could we move showEmptyGroups to SavedSettings type?

six7 (changes requested):
Few comments, I'll try it out later today but codewise other than the few comments all cool 😎

six7:
Tested in Figma and works great 👍

src/plugin/notifiers.ts
six7: Yeah let's leave it like that then. My original idea was to just move `hideEmptyGroups` to the settings state, which would have avoided that.

PR #501, opened 2022-03-03
Add expandShadow parameter for boxShadow types

iShavgula-TacTill:
Adds a possibility to expand boxShadow tokens just like the typography tokens.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,39 @@ cypress/videos
 .env.production
 package
 .idea
+
+
+
+# Created by https://www.toptal.com/developers/gitignore/api/macos
+# Edit at https://www.toptal.com/developers/gitignore?templates=macos
+
+### macOS ###
+# General
+.DS_Store
+.AppleDouble
+.LSOverride
+
+# Icon must end with two \r
+Icon
+
+
+# Thumbnails
+._*
+
+# Files that might appear in the root of a volume
+.DocumentRevisions-V100
+.fseventsd
+.Spotlight-V100
+.TemporaryItems
+.Trashes
+.VolumeIcon.icns
+.com.apple.timemachine.donotpresent
+
+# Directories potentially created on remote AFP share
+.AppleDB
+.AppleDesktop
+Network Trash Folder
+Temporary Items
+.apdisk
+
+# End of https://www.toptal.com/developers/gitignore/api/macos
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -9,3 +9,39 @@ cypress/videos
 .env.production
 package
 .idea
+
+
+
+# Created by https://www.toptal.com/developers/gitignore/api/macos
```
six7, line 15: Let's not change gitignore

Other than that this PR looks great!

six7 (changes requested):

.gitignore
iShavgula-TacTill: Good to hear !

I used this [platform](https://www.toptal.com/developers/gitignore) to add OS X specific paths. Otherwise, I'd get many `.DS_Store` files, one in every folder.

iShavgula-TacTill: I reverted the changes for the .gitignore file.

six7 (changes requested):
I noticed it doesn't yet work for shadow tokens that have multiple shadows setup (with an array as a value):

```
{
  "boxShadow": {
    "default": {
      "value": [
        {
          "x": 5,
          "y": 5,
          "spread": 3,
          "color": "rgba(0, 0, 0, 0.15)",
          "blur": 5,
          "type": "dropShadow"
        },
        {
          "x": 4,
          "y": 4,
          "spread": 6,
          "color": "#00000033",
          "blur": 5,
          "type": "innerShadow"
        }
      ],
      "type": "boxShadow"
    },
    "single": {
      "value": {
        "x": 5,
        "y": 5,
        "spread": 3,
        "color": "rgba(0, 0, 0, 0.15)",
        "blur": 5,
        "type": "dropShadow"
      },
      "type": "boxShadow"
    }
  }
}
```

iShavgula-TacTill:
Hah, I just noticed that `+` button, didn't know that was possible.

six7:
> Hah, I just noticed that `+` button, didn't know that was possible.

😀 I think it's one of the more common scenarios as most shadows have multiple levels of depth, so I'd love it if we found a way to make that work

iShavgula-TacTill:
Hey,
I just pushed changes that I made earlier, I was waiting for a confirmation from you on Slack.
Please, let me know if there are other changes that to be done.

six7 (changes requested):
Nice! Sorry, must've missed the Slack message. Works as expected 👍. Can you add a test case for single value tokens and multi value tokens to `src/utils/convertTokensObjectToResolved.test.ts` ? Then we'd be good to merge! 🎉

PR #505, opened 2022-03-04
Introduce new setting: Ignore instances

### LINKED ISSUE #503: Setting: Restrict token connection to main component
We should introduce a setting that would allow users to restrict token settings to only layers that are not part of an instance / are the instance layer itself

That would allow users to only keep tokens on their mainComponents, which is often useful.

### Discussed in https://github.com/six7/figma-tokens/discussions/488

<div type='discussions-op-text'>

<sup>Originally posted by **engegu** February 24, 2022</sup>
We use Figma Tokens a lot in constructing components for our design system. This means that instances our user pulls from the library has the Figma Tokens "metadata" on them. Is there a way to prevent this? </div>

six7:
fixes #503

### DIFF
--- a/src/app/components/Settings.tsx
+++ b/src/app/components/Settings.tsx
@@ -7,11 +7,15 @@ import { RootState, Dispatch } from '../store';
 import Label from './Label';
 
 function SyncSettings() {
-  const { ignoreFirstPartForStyles } = useSelector((state: RootState) => state.settings);
+  const { ignoreFirstPartForStyles, ignoreInstances } = useSelector((state: RootState) => state.settings);
   const dispatch = useDispatch<Dispatch>();
 
   const handleIgnoreChange = (bool) => {
     dispatch.settings.setIgnoreFirstPartForStyles(bool);
+  };
+
+  const handleIgnoreInstances = (bool) => {
+    dispatch.settings.setIgnoreInstances(bool);
   };
 
   return (
@@ -40,6 +44,20 @@ function SyncSettings() {
               <code className="p-1 -m-1 font-bold">primary/500</code>
               . Since
               Version 53 you no longer need to prepend type to token names.
+            </div>
+          </div>
+        </div>
+        <div className="space-y-4">
+          <Heading>Config</Heading>
+          <div className="space-y-2">
+            <div className="flex items-center space-x-2">
+              <Checkbox
+                id="ignoreInstances"
+                checked={ignoreInstances}
+                defaultChecked={ignoreInstances}
+                onCheckedChange={handleIgnoreInstances}
+              />
+              <Label htmlFor="ignoreInstances">Skips layers that are instances or part of instances</Label>
             </div>
           </div>
         </div>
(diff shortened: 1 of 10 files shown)


src/app/components/Settings.tsx
```diff
@@ -7,13 +7,17 @@ import { RootState, Dispatch } from '../store';
 import Label from './Label';
 
 function SyncSettings() {
-  const { ignoreFirstPartForStyles } = useSelector((state: RootState) => state.settings);
+  const { ignoreFirstPartForStyles, ignoreInstances } = useSelector((state: RootState) => state.settings);
   const dispatch = useDispatch<Dispatch>();
 
   const handleIgnoreChange = (bool) => {
```
LiamMartens, line 13: Could you just update these methods to callbacks and with complete typing whilst you're at it

```
  const handleIgnoreChange = useCallback((bool: boolean) => {
    dispatch.settings.setIgnoreFirstPartForStyles(bool);
  }, [dispatch]);

  const handleIgnoreInstances = useCallback((bool: boolean) => {
    dispatch.settings.setIgnoreInstances(bool);
  }, [dispatch]);
```

src/app/components/Settings.tsx
```diff
@@ -43,6 +47,20 @@ function SyncSettings() {
             </div>
           </div>
         </div>
+        <div className="space-y-4">
+          <Heading>Config</Heading>
+          <div className="space-y-2">
+            <div className="flex items-center space-x-2">
+              <Checkbox
+                id="ignoreInstances"
+                checked={ignoreInstances}
```
LiamMartens, line 56: Could you update this with `checked={!!ignoreFirstPartForStyles}`. Currently it causes a typing error.

src/plugin/controller.ts
```diff
@@ -52,14 +53,15 @@ figma.on('close', () => {
 });
 
 async function sendSelectionChange(): Promise<SelectionContent | null> {
-  const nodes = inspectDeep && shouldSendSelectionValues ? (await defaultNodeManager.findNodesWithData({ updateMode: UpdateMode.SELECTION })).map((node) => node.node) : Array.from(figma.currentPage.selection);
+  const nodes = inspectDeep && shouldSendSelectionValues ? (await defaultNodeManager.findNodesWithData({ updateMode: UpdateMode.SELECTION, ignoreInstances: shouldIgnoreInstances })).map((node) => node.node) : Array.from(figma.currentPage.selection);
```
LiamMartens, line 56: This is becoming very long, can we just add some newlines
```
  const nodes = (inspectDeep && shouldSendSelectionValues)
    ? (await defaultNodeManager.findNodesWithData({
      updateMode: UpdateMode.SELECTION,
      ignoreInstances: shouldIgnoreInstances,
    })).map((node) => node.node)
    : Array.from(figma.currentPage.selection);
```

src/plugin/controller.ts
```diff
@@ -52,14 +53,15 @@ figma.on('close', () => {
...
   const currentSelectionLength = figma.currentPage.selection.length;
+  console.log('sendSelectionChange', shouldIgnoreInstances, nodes);
```
LiamMartens, line 58: This can be removed

LiamMartens (changes requested):

PR #541, opened 2022-03-22
feat: drag and drop for tokens to reorder

srinivas0210:


src/app/styles/tailwind.css
```diff
@@ -462,6 +462,24 @@ input.input {
   border: 1px solid var(--black);
 }
 
+.drag-item {
+  display: none;
```
six7, line 466:
```suggestion
```

display: none causes layout shift, let's not use that

src/app/styles/tailwind.css
```diff
@@ -462,6 +462,24 @@ input.input {
...
+}
+
+.drag-over-item {
+  margin-left: 8px;
```
six7, line 470:
```suggestion
    transform: translateX(8px);
```

using transform doesn't shift the layout as much

six7 (changes requested):

PR #573, opened 2022-04-01
feat: token listing performance

LiamMartens:
@six7 I am aware this is a huge PR again.. but to properly fix some of the performance/rendering issues I had to cover quite a few areas.


src/app/components/createTokenObj.tsx
```diff
@@ -56,13 +60,20 @@ export function appendTypeToToken(token) {
 // Creates a tokens object so that tokens are displayed in groups in the UI.
 export function createTokensObject(tokens: SingleToken[], tokenFilter = '') {
   if (tokens.length > 0) {
-    const obj = tokens.reduce((acc, cur) => {
+    const obj = tokens.reduce<CreateTokensObjectResult>((acc, cur) => {
       if (tokenFilter === '' || cur.name?.toLowerCase().search(tokenFilter?.toLowerCase()) >= 0) {
-        const hasTypeProp = cur.type && cur.type !== '' && cur.type !== 'undefined';
+        // @TODO check if we need to do a "type" check.
+        // Why is this here? Will there be instances where tokens don't have a type?
```
six7, line 66: Yes, some users might not have specified one with the json.

six7 (changes requested):
Code-wise this is fine, however after testing it for a while the following no longer work:

- Drag and drop when displayType is set to List doesnt work at all
- Drag and drop now seems to allow to move items to other groups, but only visually. (the blue line)
- Deleting a token set doesn't seem to change anything both in tree and list mode
- Deleting a token group doesnt seem to change anything 
- Deleting a token doesnt seem to change anything
- Import styles doesnt seem to work anymore
- Inspector: Remap tokens doesnt seem to work anymore


I don't know where, but clicking a token set sometimes lead to `Cannot set properties of undefined (setting 'values')`

image

PR #575, opened 2022-04-02
FT-27/autocomplete

srinivas0210:


src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+  resolvedTokens,
+  showAutoSuggest,
+  setShowAutoSuggest,
+}) => {
+  const [mappedValues, setMappedValues] = React.useState([]);
+  const [currentFocus, setCurrentFocus] = React.useState(-1);
+  const filteredValue = inputValue.replace(/[^a-zA-Z0-9.]/g, ''); // removing non-alphanumberic except . from the input value
+  const focused = document.querySelector('.autosuggest-item-focused'); // get the element with this className, this is helpful for autoscroll upto focused element
+
+  const handleKeyUp = (e: { preventDefault?: any; which?: any }) => {
```
LiamMartens, line 29: Any functions inside a functional component should be properly wrapped with `useCallback` for performance reason.
It would also be great to type these events correctly

In this case I believe it would be `(e: React.KeyboardEvent<HTMLDivElement>)`

src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+  children,
+  inputValue,
+  setInputValue,
+  resolvedTokens,
+  showAutoSuggest,
+  setShowAutoSuggest,
+}) => {
+  const [mappedValues, setMappedValues] = React.useState([]);
+  const [currentFocus, setCurrentFocus] = React.useState(-1);
+  const filteredValue = inputValue.replace(/[^a-zA-Z0-9.]/g, ''); // removing non-alphanumberic except . from the input value
```
LiamMartens, line 26: Can we `useMemo` this `useMemo(() => inputValue.replace(/[^a-zA-Z0-9.]/g, ''), [inputValue])`

src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const focused = document.querySelector('.autosuggest-item-focused'); // get the element with this className, this is helpful for autoscroll upto focused element
```
LiamMartens, line 27: Is there any possibility to implement this without using a `querySelector`? This is generally bad practice within React

src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const handleKeyUp = (e: { preventDefault?: any; which?: any }) => {
+    const { which } = e;
+
+    if (which === 38 && currentFocus > 0) {
+      e.preventDefault();
+      setCurrentFocus(currentFocus - 1);
+    }
+  };
+
+  const handleKeyDown = (e: { preventDefault?: any; which?: any }) => {
```
LiamMartens, line 38: Same comment as before - make this a `useCallback`

src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+      e.preventDefault();
+      setCurrentFocus(currentFocus + 1);
+    }
+    if (which === 13) {
+      e.preventDefault();
+      setInputValue(`{${mappedValues[currentFocus].name}}`);
+    }
+  };
+
+  const getHighlightedText = (text: string, highlight: string) => {
```
LiamMartens, line 51: This should also be a `useMemo`

src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+    );
+  };
+
+  useEffect(() => {
+    if (inputValue.includes('}')) {
+      setShowAutoSuggest(false);
+      setMappedValues([]);
+      return;
+    }
+    const filteredOnType = resolvedTokens.filter((token: any) => token?.type === type);
```
LiamMartens, line 72: Not sure if avoidable in this case but let's refrain from using `any` for any new code when possible

src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+  useEffect(() => {
+    if (focused) focused.scrollIntoViewIfNeeded(true);
+  }, [focused]);
+
+  return (
+    <div className="relative autosuggest" onKeyUp={handleKeyUp} onKeyDown={handleKeyDown}>
+      {children}
+      {mappedValues && mappedValues.length > 0 && (
+        <div className="absolute w-full overflow-y-scroll cursor-pointer z-10 dropdown">
+          {mappedValues.map((token: any, index: number) => (
```
LiamMartens, line 96: Also here - let's try to type this correctly

src/app/components/Autosuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <div
+              key={index}
+              onClick={() => setInputValue(`{${token.name}}`)}
```
LiamMartens, line 99: This should also be a `useCallback`

src/app/components/EditTokenForm.tsx
```diff
@@ -231,43 +256,53 @@ function EditTokenForm({ resolvedTokens }: Props) {
...
+                      className="block w-4 h-4 rounded-sm cursor-pointer shadow-border shadow-gray-300 focus:shadow-focus focus:shadow-primary-400"
+                      style={{ background: internalEditToken.value, fontSize: 0 }}
+                      onClick={handleToggleInputHelper}
+                    >
+                      {internalEditToken.value}
+                    </button>
+                  )
+                }
+                suffix={(
+                  <StyledInputSuffix onClick={() => setShowAutoSuggest(!showAutoSuggest)}>
```
LiamMartens, line 292: Make the `onClick` a `useCallback`

src/app/styles/autosuggest.css
```diff
@@ -0,0 +1,50 @@
+@import url('./main.css');
+
+.dropdown {
```
LiamMartens, line 3: Is there any way we can style this with stitches instead?

LiamMartens (changes requested):

src/app/components/DownshiftInput.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+interface DownShiftProps {
+  type: string;
+  label?: string;
+  error?: string;
+  value: string;
+  placeholder?: string;
+  prefix?: React.ReactNode;
+  suffix?: React.ReactNode;
+  resolvedTokens: any;
```
LiamMartens, line 73: We need to get out of the `any` hell. This should be `ResolveTokenValuesResult`. It may not be 100% correct but it's better than any

src/app/components/DownshiftInput.tsx
```diff
@@ -0,0 +1,173 @@
...
+  setInputValue: Function;
```
LiamMartens, line 74: We should not use `Function`. It is too generic and could accept any kind of paramter and return any kind of value. It should be a properly typed function definition ie `(param: string) => void`

src/app/components/EditTokenForm.tsx
```diff
@@ -256,45 +256,38 @@ function EditTokenForm({ resolvedTokens }: Props) {
...
+                    style={{ background: internalEditToken.value, fontSize: 0 }}
+                    onClick={handleToggleInputHelper}
+                  >
+                    {internalEditToken.value}
+                  </button>
+                )
+              }
+              suffix={(
+                <StyledInputSuffix
+                  onClick={React.useCallback(() => setShowAutoSuggest(!showAutoSuggest), [showAutoSuggest])}
```
LiamMartens, line 285: I appreciate the useCallback but it shoulnd't be used within the render return. Please move it to a separate variable before the render code

src/app/components/DownshiftInput.tsx
```diff
@@ -0,0 +1,173 @@
...
+                .filter(
+                  (token: SingleToken) => !filteredValue || token.name.toLowerCase().includes(filteredValue.toLowerCase()),
+                )
+                .filter((token: SingleToken) => token?.type === type)
+                .map((token: SingleToken, index: number) => (
+                  <StyledItem
+                    className="dropdown-item"
+                    {...getItemProps({ key: token.name, index, item: token })}
+                    css={{
+                      backgroundColor: highlightedIndex === index ? '#18a0fb' : 'white',
```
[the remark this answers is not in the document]
srinivas0210: Hey..I missed this 👍

src/app/components/DownshiftInput.tsx
```diff
@@ -0,0 +1,173 @@
...
+                onChange: handleChange,
+              })}
+            />
+            {!!suffix && <button {...getToggleButtonProps()}>{suffix}</button>}
+          </Box>
+
+          {showAutoSuggest || (value.includes('{') && !value.includes('}')) ? (
+            <StyledDropdown>
+              {resolvedTokens
+                .filter(
```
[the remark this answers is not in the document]
srinivas0210: sure, makes sense 👍

srinivas0210:
Hey @LiamMartens , Thanks for reviewing my PR, I have just resolved your comments, Please review 😃

six7:
Do you need any help with this one?

PR #579, opened 2022-04-06
FT-4 sync with gitlab, pull from gitlab, upload tokens, used @gitbreak/bro…

swordEdge:
…wser as a gitlab client


six7 (commented):
thought i'd give you some comments while you're still working on this. this looks great so far! 🎉

src/app/components/SyncSettings.tsx
```diff
@@ -35,7 +35,7 @@ const SyncSettings = () => {
     setShowEditStorageModalVisible(true);
   };
 
-  const selectedRemoteProvider = () => [StorageProviderType.JSONBIN, StorageProviderType.GITHUB, StorageProviderType.URL].includes(
+  const selectedRemoteProvider = () => [StorageProviderType.JSONBIN, StorageProviderType.GITHUB, StorageProviderType.GITLAB, StorageProviderType.URL].includes(
```
LiamMartens, line 38: Could we just update this with a `useMemo` hook whilst we're working on this.
[useMemo](https://reactjs.org/docs/hooks-reference.html#usememo)

src/app/components/SyncSettings.tsx
```diff
@@ -175,6 +191,21 @@ const SyncSettings = () => {
                 text="GitHub"
                 id={StorageProviderType.GITHUB}
               />
+              <ProviderSelector
+                isActive={localApiState?.provider === StorageProviderType.GITLAB}
+                isStored={storageType?.provider === StorageProviderType.GITLAB}
+                onClick={() => {
```
LiamMartens, line 197: Let's make use of React's useCallback hook so we don't create a new function reference every cycle

LiamMartens (changes requested):
Left 2 comments

swordEdge:
@six7 @LiamMartens 

> Errors on Node.js CI / Create Package, Node.js CI / Tests

The dependency(gitbeaker/browser) is relying on Node.js 16+, but Vercel currently does not have support yet, as we are dependent on AWS.
https://github.com/vercel/community/discussions/37

six7:
> 

I think it's fine. Vercel is only required for previews, we should not make the plugin dependable on that.

Any updates on the failings tests? Would love to get this merged.

LiamMartens (approved):

LiamMartens:
@swordEdge just need to update the`node.js.yml` workflow to use Node 14 and we should be all good

PR #580, opened 2022-04-06
View token layers & Select All - FT-40 / FT-41

roppazvan:
https://github.com/six7/figma-tokens/issues/513
https://github.com/six7/figma-tokens/issues/519

* created a new TokenLayers component
* Added radix icons for the layer listing
* Added select all functionality


src/app/components/inspector/TokenNodes.tsx
```diff
@@ -0,0 +1,90 @@
...
+              transition: 'background 200ms ease',
+              '&:hover': {
+                background: 'rgba(196, 196, 196, 0.15)',
+              },
+            }}
+          >
+            <Box css={{ color: '$fgSubtle', marginRight: '$3' }}>
+              <IconLayers />
+            </Box>
+            <Box css={{ color: '#575757' }}>{nodes.length}</Box>
```
six7, line 83:
```suggestion
            <Box css={{ color: '$textMuted' }}>{nodes.length}</Box>
```

src/app/components/inspector/TokenNodes.tsx
```diff
@@ -0,0 +1,90 @@
...
+            css={{
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'center',
+              padding: '$2 $3',
+              borderRadius: '6px',
+              cursor: 'pointer',
+              transition: 'background 200ms ease',
+              '&:hover': {
+                background: 'rgba(196, 196, 196, 0.15)',
```
six7, line 76:
```suggestion
                background: '$bgSubtle',
```

sorry, that was my bad in the design

src/app/components/inspector/TokenNodes.tsx
```diff
@@ -0,0 +1,90 @@
...
+        }}
+      >
+        <DropdownMenu.Trigger>
+          <Box
+            css={{
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'center',
+              padding: '$2 $3',
+              borderRadius: '6px',
```
six7, line 72:
```suggestion
              borderRadius: '$default',
```

src/app/components/inspector/getNodeIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+  return (
+    icon && (
+      <Box
+        css={{
+          marginRight: '$3',
+          display: 'flex',
+          alignItems: 'center',
+          svg: {
+            fill: '#fff',
```
six7, line 55:
```suggestion
            fill: '$contextMenuForeground',
```

i dont know if stitches picks up svg fill - could be we'll have to change that. however we'd want to use tokens here as we might introduce new themes in the future so let's keep raw hex values out of this

six7 (changes requested):
🎉 just some minor color usage comments. i can't run the plugin currently (seems to be connected to the upgraded dropdown package)

esthercheran:
FT-40 & FT-41

src/app/components/inspector/TokenNodes.tsx
```diff
@@ -0,0 +1,123 @@
...
+            color: '$contextMenuForeground',
+            cursor: 'pointer',
+            padding: '$1 $4',
+            '&:hover': {
+              backgroundColor: '$interaction',
+              color: '$onInteraction',
+            },
+          }}
+        >
+          {getNodeIcon(type)}
```
LiamMartens, line 37: Would be great if we can useMemo this so it doesn't get re-evaluated each time

src/app/components/inspector/TokenNodes.tsx
```diff
@@ -0,0 +1,123 @@
...
+const VISIBLE_VIEWPORT_NODES = 10;
+const CONTAINER_PADDING = 8;
+
+export default function TokenNodes({ nodes }: { nodes: NodeInfo[] }) {
+  function selectAllNodes() {
+    const nodeIds = nodes.map(({ id }) => id);
+    selectNodes(nodeIds);
+  }
+
+  function getNode({ id, name, type }: NodeInfo) {
```
LiamMartens, line 22: Can we just split this into it's own component. It is bad practice to create new component/functions like this because it can cause performance issues.

src/app/components/inspector/TokenNodes.tsx
```diff
@@ -0,0 +1,123 @@
...
+    return (
+      <DropdownMenu.Item key={id} onSelect={() => goToNodeId(id)}>
```
LiamMartens, line 24: When split into its own component the onSelect can also be attached to a useCallback hook, again to prevent any unwanted re-renders due to reference changes

src/app/components/inspector/getNodeIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+  BoxIcon,
+  MinusIcon,
+  ValueIcon,
+} from '@radix-ui/react-icons';
+import Box from '../Box';
+
+const SVG_WIDTH = 12;
+const SVG_HEIGHT = 12;
+
+function getNodeIcon(type: NodeType) {
```
LiamMartens, line 17: this should either move to utils or be transformed into a proper component. Now it's in the components folder appearing to be a component but it isnt.

src/types/messages.tsx
```diff
@@ -244,22 +245,28 @@ export type ChangedTabsToPluginMessage = {
   requiresSelectionValues: boolean;
 };
 
+export type SelectNodes = {
```
LiamMartens, line 248: Can we change this to `SelectNodesPluginMessage`, only because that's also the name of all the other types

LiamMartens (changes requested):

LiamMartens (approved):

PR #594, opened 2022-04-10
feat: [FT-6] Undo actions

LiamMartens:


src/app/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,19 @@
+import { styled } from '@/stitches.config';
```
six7, line 1: You could use the `<Stack>` component that was introduced recently instead of this one, which allows you to use flex with a certain `gap` but can be used just like the Flex in here.

src/app/components/NavbarUndoButton/NavbarUndoButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const NavbarUndoButton: React.FC = () => {
+  const canUndo = useCanUndo();
+
+  const handleUndo = useCallback(() => {
+    UndoableEnhancerState.undo();
+  }, []);
+
+  if (canUndo) {
+    return (
+      <StyledUndoButton type="button" onClick={handleUndo}>
```
six7, line 26: Can we add a tooltip to this one?

Depending on if it's possible, could we:
`Undo %actionName` where action name is the name of the action? e.g. `create token`, `delete token`, `duplicate token`?

If not, just `Undo previous action`? 

I feel like the first one would give the user clarity as to what they're undoing, as some probably expect that any action would be undoable and this could help them understand what's happening.

src/app/components/icons/IconUndo.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type Props = React.SVGAttributes<SVGSVGElement>;
```
six7, line 3: No need to create a tsx component anymore, since a couple versions ago you can just use
`import IconUndo from '@/icons/undo.svg';`

six7 (changes requested):
Some minor thing, overall this feels great to use!

PR #603, opened 2022-04-13
Ft 44 combine sync and settings tab

srinivas0210:


six7 (commented):
Can you do a Search for `syncsettings` and replace that with `settings`? We currently set tab to syncsettings when a file has github sync but the user doesn't have the correct credentials

src/app/components/Navbar.tsx
```diff
@@ -60,11 +60,10 @@ export const Navbar: React.FC = () => {
         transform: 'translateY(-1px)',
       }}
     >
-      <Stack gap={0} direction="row" align="center" justify="between" css={{ width: '100%' }}>
+<Stack gap={0} direction="row" align="center" justify="between" css={{ width: '100%' }}>
```
six7, line 63: This seems to be a mistake?

src/app/components/Settings.tsx
```diff
@@ -1,39 +1,230 @@
...
+  const localApiState = useSelector(localApiStateSelector);
+  const storageType = useSelector(storageTypeSelector);
+  const apiProviders = useSelector(apiProvidersSelector);
+
+  const { setStorageType } = useStorage();
 
-  const handleIgnoreChange = React.useCallback((bool: boolean) => {
-    dispatch.settings.setIgnoreFirstPartForStyles(bool);
-  }, [dispatch.settings]);
+  const [confirmModalVisible, showConfirmModal] = React.useState(false);
```
six7, line 34: Can you keep Syncsettings as their own component and import it here?? That way this files does not get as big.

six7 (changes requested):
Some comments and remember to change tab for StartScreen.

Also, make sure to remove any remnants of SyncSettings, especially in constants / Tabs

src/app/components/Settings.tsx
```diff
@@ -1,39 +1,230 @@
...
+  const [editStorageItemModalVisible, setShowEditStorageModalVisible] = React.useState(Boolean(localApiState.new));
+  const [createStorageItemModalVisible, setShowCreateStorageModalVisible] = React.useState(false);
+
+  const handleEditClick = (provider) => {
```
LiamMartens, line 38: These function handlers should be `useCallback`'s to prevent unnecessary re-renders

src/app/components/Settings.tsx
```diff
@@ -1,39 +1,230 @@
...
+    track('Edit Credentials');
+    dispatch.uiState.setLocalApiState(provider);
+    setShowEditStorageModalVisible(true);
+  };
+
+  const selectedRemoteProvider = () => [StorageProviderType.JSONBIN, StorageProviderType.GITHUB, StorageProviderType.URL].includes(
```
LiamMartens, line 44: This can probably be a static `useMemo` instead. We don't need to re-evaluate this each time

src/app/components/Settings.tsx
```diff
@@ -1,39 +1,230 @@
...
+    localApiState?.provider as StorageProviderType,
+  );
+
+  const storedApiProviders = () => apiProviders.filter((item) => item.provider === localApiState?.provider);
```
LiamMartens, line 48: Same with this

src/app/components/Settings.tsx
```diff
@@ -1,39 +1,230 @@
...
+
+  const storageProviderText = () => {
```
LiamMartens, line 50: And this one as well

src/app/components/Settings.tsx
```diff
@@ -1,39 +1,230 @@
...
-              defaultChecked={ignoreFirstPartForStyles}
-              onCheckedChange={handleIgnoreChange}
-            />
-            <Label htmlFor="ignoreFirstPartForStyles">Ignore first part of token name for styles</Label>
+    <Box css={{ display: 'flex', flexDirection: 'column', flexGrow: 1 }} className="content scroll-container">
+      {confirmModalVisible && (
+        <ConfirmLocalStorageModal
+          isOpen={confirmModalVisible}
+          onClose={showConfirmModal}
+          onSuccess={() => {
```
LiamMartens, line 99: This should be a `useCallback`

src/app/components/Settings.tsx
```diff
@@ -1,39 +1,230 @@
...
+              shouldSetInDocument: true,
+            });
+            showConfirmModal(false);
+          }}
+        />
+      )}
+      {editStorageItemModalVisible && (
+        <EditStorageItemModal
+          isOpen={editStorageItemModalVisible}
+          onClose={() => setShowEditStorageModalVisible(false)}
```
LiamMartens, line 112: Any handlers should be `useCallback`

LiamMartens (changes requested):

PR #604, opened 2022-04-13
feat: [FT-64] Use token set as source

LiamMartens:


src/plugin/node.ts
```diff
@@ -114,7 +115,7 @@ export function selectNodes(ids: string[]) {
 
 export async function updateNodes(
   entries: readonly NodeManagerNode[],
-  tokens: Map<string, TokenArrayGroup[number]>,
+  tokens: Map<string, TokenArrayGrou[number]>,
```
six7, line 118: this feels like a mistake?

src/utils/convertTokensObjectToResolved.ts
```diff
@@ -18,7 +19,13 @@ export default function convertTokensObjectToResolved(
   // Parse tokens into array structure
   const parsed = parseTokenValues(tokens);
   // Merge to one giant array
-  const merged = mergeTokenGroups(parsed, usedSets);
+  const merged = mergeTokenGroups(
+    parsed,
+    // @README as I understand this function is only used in the utils/transform file
+    // which in turn is only used for a local script -- in which case for now we do not
```
six7, line 25: correct, the token-transformer has a way to exclude token sets which is fine for now. users basically first define all inputs and later define which sets should be excluded from output

six7 (changes requested):
Just this one mini comment that feels like a mistake, other than this good to merge!

LiamMartens:
@six7 great! I fixed the typing so we can go ahead and merge this so we can continue with the themes :+1:

six7 (approved):

six7:
Nice! 🎉

mstllc:
@LiamMartens what does this `Treat as source` option actually do? 😄

six7:
> @LiamMartens what does this `Treat as source` option actually do? 😄 

It doesnt include the tokens in the actual output, like if you wouldn't want your global tokens to create styles for you, as all you care about might be just the theme specific semantic tokens

PR #605, opened 2022-04-14
Ft 43 git sync introduce branch switch 341

swordEdge:


six7 (commented):
Bit of functional feedback:
- Can we remove the down arrow next to the branch name? The branch name with a hover background would be enough.
- When a user has to enter a new branch name (e.g. by using create a new branch from..) can we focus the input? So the user doesnt have to click into that?
- I tried creating a new branch, but it seems like it created 2

image

- The interaction when a user has local changes feels a bit weird right now as two modals pop up after another. I'll think about this some more, but I feel like we should consolidate that to 1 modal

swordEdge:
> * Can we remove the down arrow next to the branch name? The branch name with a hover background would be enough.
This is fixed.

> * I tried creating a new branch, but it seems like it created 2
This is fixed.

> * The interaction when a user has local changes feels a bit weird right now as two modals pop up after another. I'll think about this some more, but I feel like we should consolidate that to 1 modal
Waiting for instruction.
@six7

six7:
> _Waiting for instruction._
> 
> 
> 
> @six7 

Can we make it like this: 
You have unsaved changes. [Cancel] [Discard changes].
`Cancel`: Cancels the action, user stays on current branch, still has their local changes
`Discard changes`: New branch is immediately checked out and pulled, no second modal.

swordEdge:
> Can we make it like this: You have unsaved changes. [Cancel] [Discard changes]. `Cancel`: Cancels the action, user stays on current branch, still has their local changes `Discard changes`: New branch is immediately checked out and pulled, no second modal.

src/app/components/Footer.tsx
```diff
@@ -1,23 +1,86 @@
...
-        Version
-        {' '}
-        {pjs.plugin_version}
-      </Box>
+      <Stack direction="row">
+        {localApiState.branch && (
+        <>
+          <BranchSelector currentBranch={localApiState.branch} />
+          <Tooltip variant="top" label={`Pull from ${transformProviderName(storageType.provider)}`}>
+            <button onClick={() => pullTokens({})} type="button" className="button button-ghost">
```
six7, line 58: can we make sure to also specify usedTokenSet here? we did this before, having that here makes sure that the token sets the user had active before stay active. right now they get reset.

src/app/components/BranchSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const onBranchSelected = async (branch: string) => {
+    if (checkForChanges() && await askUserIfPushChanges()) {
+      await pushTokens();
+    }
+
+    setMenuOpened(false);
+    setCurrentBranch(branch);
+
+    if (await askUserIfPull()) await pullTokens({ context: { ...apiData, branch } });
```
six7, line 101: same: can we make sure to also specify usedTokenSet here? we did this before, having that here makes sure that the token sets the user had active before stay active. right now they get reset.

six7:
Thanks for the changes! Feeling much better that way 🔥

src/app/components/Input.tsx
```diff
@@ -159,6 +161,7 @@ const Input: React.FC<Props> = ({
           name={name}
           onChange={onChange}
           required={required}
+          autoFocus={autofocus}
```
six7, line 164: I'm not sure autofocus works right now, it's not focusing for me.

src/app/components/BranchSwitchMenu.tsx
```diff
@@ -0,0 +1,117 @@
+import * as BranchSwitchMenuPrimitive from '@radix-ui/react-dropdown-menu';
+import { styled } from '@/stitches.config';
+
+const itemStyles = {
```
six7, line 4: could this just use a DropdownMenu? Or do we have to recreate all those styles?

swordEdge: > could this just use a DropdownMenu? Or do we have to recreate all those styles?

I've tried to use the existing DropdownMenu but the submenu appeared under the trigger item and also many styles were not good fit to the DESIGN of the branch switching menu.

src/app/components/Footer.tsx
swordEdge: > can we make sure to also specify usedTokenSet here? we did this before, having that here makes sure that the token sets the user had active before stay active. right now they get reset.

Fixed!

src/app/components/BranchSelector.tsx
swordEdge: > same: can we make sure to also specify usedTokenSet here? we did this before, having that here makes sure that the token sets the user had active before stay active. right now they get reset.

Fixed!

swordEdge:
> Can we make it like this: You have unsaved changes. [Cancel] [Discard changes]. `Cancel`: Cancels the action, user stays on current branch, still has their local changes `Discard changes`: New branch is immediately checked out and pulled, no second modal.

Fixed!

src/app/components/BranchSwitchMenu.tsx
swordEdge: @six7 do you have any other ideas?

six7: Got it, other than that it's 👍 from my end! Would you be able to add tests? IMO we should write tests for any new features we're building so we can increase test coverage. Once Liam has reviewed we can merge.

src/app/components/BranchSelector.tsx
```diff
@@ -0,0 +1,184 @@
...
+  }, [localApiState.branch, setCurrentBranch]);
+
+  const checkForChanges = React.useCallback(() => {
+    if (lastSyncedState !== JSON.stringify(convertTokensToObject(tokens), null, 2)) {
+      return true;
+    }
+    return false;
+  }, [lastSyncedState, tokens]);
+
+  async function askUserIfPull(): Promise<boolean> {
```
LiamMartens, line 54: This should be a `useCallback`

src/app/components/BranchSelector.tsx
```diff
@@ -0,0 +1,184 @@
...
+    const { result } = await confirm({
+      text: 'Pull from GitHub?',
+      description: 'Your repo already contains tokens, do you want to pull these now?',
+    });
+    return result;
+  }
+
+  async function askUserIfPushChanges(): Promise<boolean> {
```
LiamMartens, line 62: Same with any other function created inside this component.

src/app/components/BranchSelector.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    </BranchSwitchMenuRadioItem>
+                  ))}
+              </BranchSwitchMenuRadioGroup>
+              <BranchSwitchMenu>
+                <BranchSwitchMenuTrigger>
+                  Create new branch from
+                  <ChevronRightIcon />
+                </BranchSwitchMenuTrigger>
+                <BranchSwitchMenuContent side="left">
+                  {checkForChanges()
```
LiamMartens, line 155: Would be better if we can memo this effect otherwise it will run on ever render

src/app/components/BranchSelector.tsx
```diff
@@ -0,0 +1,184 @@
...
+            <BranchSwitchMenuMainTrigger>
+              <GitBranchIcon size={16} />
+              <span>{currentBranch}</span>
+            </BranchSwitchMenuMainTrigger>
+
+            <BranchSwitchMenuContent side="top" sideOffset={5}>
+              <BranchSwitchMenuRadioGroup value={currentBranch}>
+                {branchState.branches.length > 0
+                  && branchState.branches.map((branch, index) => (
+                    <BranchSwitchMenuRadioItem key={index} value={branch} onSelect={() => onBranchSelected(branch)}>
```
LiamMartens, line 140: Move this to a `useCallback` as well so we're not creating new functions each time. Probably better to move these entries into their own smaller component anyways to separate concerns.
(the code that's inside the `map`)

src/app/components/BranchSelector.tsx
```diff
@@ -0,0 +1,184 @@
...
+                </BranchSwitchMenuTrigger>
+                <BranchSwitchMenuContent side="left">
+                  {checkForChanges()
+                  && (
+                  <BranchSwitchMenuItem onSelect={createBranchByChange}>
+                    Current changes
+                  </BranchSwitchMenuItem>
+                  )}
+                  {branchState.branches.length > 0 && branchState.branches.map((branch, index) => (
+                    <BranchSwitchMenuItem key={index} onSelect={() => createNewBranchFrom(branch)}>
```
LiamMartens, line 162: Same comment on this; `useCallback` it and move to it's own small component

src/app/components/BranchSwitchMenu.tsx
LiamMartens: @six7 we may want to reconcile some of this from design up - so things can be reused more (not necessary now)

src/app/components/Footer.tsx
```diff
@@ -1,23 +1,87 @@
...
+  const { pullTokens, pushTokens } = useRemoteTokens();
+
+  const checkForChanges = React.useCallback(() => {
+    if (lastSyncedState !== JSON.stringify(convertTokensToObject(tokens), null, 2)) {
+      return true;
+    }
+    return false;
+  }, [lastSyncedState, tokens]);
+
+  const transformProviderName = (provider: StorageProviderType) => {
```
LiamMartens, line 36: make this a `useCallback` like the `checkForChanges` function

src/app/components/Footer.tsx
```diff
@@ -1,23 +1,87 @@
...
-        Version
-        {' '}
-        {pjs.plugin_version}
-      </Box>
+      <Stack direction="row">
+        {localApiState.branch && (
+        <>
+          <BranchSelector currentBranch={localApiState.branch} />
+          <Tooltip variant="top" label={`Pull from ${transformProviderName(storageType.provider)}`}>
+            <button onClick={() => pullTokens({ usedTokenSet })} type="button" className="button button-ghost">
```
LiamMartens, line 59: Any handlers should be `useCallback`'ed

src/app/components/Footer.tsx
```diff
@@ -1,23 +1,87 @@
...
+              <DownloadIcon />
+            </button>
+          </Tooltip>
+          <Tooltip variant="top" label={`Push to ${transformProviderName(storageType.provider)}`}>
+            <button
+              onClick={() => pushTokens()}
```
LiamMartens, line 65: Same comment about the handlers

src/app/components/Footer.tsx
```diff
@@ -1,23 +1,87 @@
...
+              type="button"
+              className="relative button button-ghost"
+              disabled={editProhibited}
+            >
+              {checkForChanges() && <div className="absolute top-0 right-0 w-2 h-2 rounded-full bg-primary-500" />}
```
LiamMartens, line 70: Same comment as in some other file - if we can memo result of this function that would be great so we can not re-run the logic on each cycle.

src/app/components/Initiator.tsx
```diff
@@ -99,6 +101,17 @@ export function Initiator() {
               track('Fetched from remote', { provider: credentials.provider });
               if (!credentials.internalId) track('missingInternalId', { provider: credentials.provider });
 
+              const {
+                id, provider, secret, baseUrl,
+              } = credentials;
+              const [owner, repo] = id.split('/');
+              if (provider === StorageProviderType.GITHUB) {
+                const branches = await fetchGithubBranches({
```
LiamMartens, line 109: @six7 not for now but we should think about a more central way of managing all our remote/async data in the future. It's very scattered at the moment which can cause more bugs in the long run due to it's disconnected nature

src/app/components/Input.tsx
LiamMartens: not sure if this has been resolved yet?

src/app/components/modals/CreateBranchModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const inputRef = React.useRef(null);
+
+  /* @lifecycle
+  ** set focus on input
+  */
+  React.useEffect(() => {
+    inputRef.current?.focus();
+  }, [inputRef.current]);
+
+  const handleCreateNewClick = async () => {
```
LiamMartens, line 41: Let's make sure we do the hooks on new code

src/app/components/modals/CreateBranchModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  isOpen, onClose, onSuccess, startBranch, isCurrentChanges,
+}: Props) {
+  const { addNewBranch, pushTokens, fetchBranches } = useRemoteTokens();
+
+  const localApiState = useSelector(localApiStateSelector);
+  const apiData = useSelector(apiSelector);
+
+  const [formFields, setFormFields] = React.useState({});
+  const [hasErrored, setHasErrored] = React.useState(false);
+  const inputRef = React.useRef(null);
```
LiamMartens, line 32: This should be typed correctly. In this case it should be `useRef<HTMLInputElement | null>`

src/app/components/modals/CreateBranchModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+export default function CreateBranchModal({
+  isOpen, onClose, onSuccess, startBranch, isCurrentChanges,
+}: Props) {
+  const { addNewBranch, pushTokens, fetchBranches } = useRemoteTokens();
+
+  const localApiState = useSelector(localApiStateSelector);
+  const apiData = useSelector(apiSelector);
+
+  const [formFields, setFormFields] = React.useState({});
```
LiamMartens, line 30: This should also be typed correctly. Let's make sure the next developer knows exactly what fields could end up in here

src/app/components/modals/CreateBranchModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+    if (response) {
+      onSuccess(branch, branches);
+    } else {
+      setHasErrored(true);
+    }
+
+    if (isCurrentChanges) await pushTokens({ ...apiData, branch });
+  };
+
+  const handleChange = (e: any) => {
```
LiamMartens, line 63: The event should not use `any` - this should be `(e: React.ChangeEvent<HTMLInputElement>)`

src/app/components/modals/CreateBranchModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+    setFormFields({ ...formFields, [e.target.name]: e.target.value });
+  };
+
+  const handleSubmit = (e: any) => {
```
LiamMartens, line 67: This can also be typed `(e: React.FormEvent<HTMLFormElement>)`

src/app/components/modals/CreateBranchModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+    e.preventDefault();
+    handleCreateNewClick();
+  };
+
+  return (
+    <Modal large isOpen={isOpen} close={() => onClose(false)}>
```
LiamMartens, line 73: Let's move the handler to a `useCallback` as well

LiamMartens (changes requested):

src/app/components/BranchSelector.tsx
swordEdge: Fixed!

src/app/components/BranchSelector.tsx
swordEdge: Fixed!

src/app/components/BranchSelector.tsx
swordEdge: Fixed!

src/app/components/BranchSelector.tsx
swordEdge: Fixed!

src/app/components/Footer.tsx
swordEdge: Fixed!

src/app/components/Footer.tsx
swordEdge: Fixed!

src/app/components/Footer.tsx
swordEdge: Fixed!

src/app/components/modals/CreateBranchModal.tsx
swordEdge: fixed!

src/app/components/modals/CreateBranchModal.tsx
swordEdge: Fixed!

src/app/components/modals/CreateBranchModal.tsx
swordEdge: Fixed!

src/app/components/modals/CreateBranchModal.tsx
swordEdge: Fixed!

src/app/components/modals/CreateBranchModal.tsx
swordEdge: Fixed!

src/app/components/Footer.tsx
swordEdge: Fixed!

src/app/components/Input.tsx
swordEdge: Resolved!

src/app/components/BranchSelector.tsx
```diff
@@ -82,11 +76,12 @@ export default function BranchSelector() {
   };
 
   const createNewBranchFrom = async (branch: string) => {
```
LiamMartens, line 78: As mentioned before - all functions created inside components should be either `useCallback` or `useMemo`

src/app/components/BranchSelector.tsx
```diff
@@ -123,6 +118,29 @@ export default function BranchSelector() {
     dispatch.uiState.setLocalApiState({ ...localApiState, branch });
   };
 
+  const BranchSwitchMenuRadioElement = ({ branch, index } : { branch: string, index:number }) => {
```
LiamMartens, line 121: If possible - let's move these to their own component files

LiamMartens (changes requested):

src/app/components/BranchSelector.tsx
swordEdge: I think it's impossible to move these to their own component files because the function outside was used in it. @LiamMartens

src/app/components/BranchSelector.tsx
swordEdge: image

**createNewBranchFrom** function was used once in a function which is wrapped by useCallback. Do we still need to redefine with useCallback or useMemo? @LiamMartens

LiamMartens: Yes - really this should be separated because of the same reason as the `BranchSwitchMenuRadioElement` variable

six7 (approved):

six7:
@swordEdge changed base to `next` - there's one more conflict for Navbar, and the two comments from Liam - other than that let's get this merged once that is done.

src/app/components/BranchSelector.tsx
swordEdge: > It would be better to separate it, even if it is in this same file and pass the function as props. This way the function reference doesn't change and will never cause a re-render

> fixed

src/app/components/BranchSelector.tsx
swordEdge: fixed

PR #611, opened 2022-04-16
feat: FT-39 manage themes

LiamMartens:


src/app/components/Button.tsx
```diff
@@ -1,84 +1,111 @@
...
-  switch (size) {
-    case 'large':
-      sizeClass = 'button-large';
-      break;
-    default:
-      break;
-  }
+  }, [id, onClick]);
+
+  const variantClass = React.useMemo(() => (
```
six7, line 55: Could we move away from variant classes and instead use Stitches' variants? That should be way easier to work with.

src/app/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,17 @@
...
+import Text from '../Text';
+import { StitchesCSS } from '@/types';
+
+type Props = React.HTMLAttributes<HTMLDivElement> & {
+  css?: StitchesCSS
+  title: string
+  subtitle?: string
+};
+
+export const EmptyState: React.FC<Props> = ({ title, subtitle, ...props }) => (
```
six7, line 12: There's the `Blankslate` component already, could we either consolidate this with the other one or just use the other one?

six7 (commented):
This is great already! I added two minor comments

src/app/components/Button.tsx
LiamMartens: @six7 yea the button situation is something I want to ask you - I don't even think the variant classes work anymore so I wanted to check

src/app/components/EmptyState/EmptyState.tsx
LiamMartens: @six7 of course, ill have a look - I didnt realize it existed

src/app/components/Button.tsx
six7: i'm currently integrating a loading state into the <Button> - i can refactor it separately, no need to do that in this PR. i'll take your changes from this PR

six7:
One thing that would be helping users:
- When hitting `Create theme` could we take the current token sets as a basis? So whatever is set there, is set in the Create theme dialog? That way a user would not have to recreate everything from scratch when they create a new theme, but could use what they have currently as a basis. E.g. going from dark to light mode would mean only swapping one token set.

LiamMartens:
@six7 I updated the remote storage providers to support additional files for themes etc.. also re-worked it to be more generic and interchangeable for the future. -- I want to get this functionally checked and merged into next without the figma style ID logic because it's getting pretty large

six7 (approved):

PR #612, opened 2022-04-16
Ft 25 deep linking aliases 457

swordEdge:


six7:
Is this the right branch? The changes contain the git branch switcher, not deep linking aliases?

swordEdge:
> Is this the right branch? The changes contain the git branch switcher, not deep linking aliases?

Yes, It's right. This branch is just created from `switching branch` branch, that's why.

six7:
> > Is this the right branch? The changes contain the git branch switcher, not deep linking aliases?
> 
> Yes, It's right. This branch is just created from `switching branch` branch, that's why.

It'd be better if all our branches were based off of `main`, not of a branch that is currently in development, especially as the other one is not connected to this feature at all. If that other branch turns out non-mergeable we won't be able to merge this one either.

swordEdge:
I am gonna send another PR regarding FT-25 so I close this PR.

PR #618, opened 2022-04-18
Feat: minimise plugin

srinivas0210:


six7:
Neat! What do you think of changing the width of the plugin window as well to further reduce size? Maybe just to the icon width and a bit padding?

src/app/components/App.tsx
```diff
@@ -15,12 +17,46 @@ import ConfirmDialog from './ConfirmDialog';
...
+  }, [dispatch, windowSize]);
 
-  return (
+  return isPluginminimized ? (
+    <StyledButton type="button" onClick={handleResize}>
+      <MaximiseIcon />
+    </StyledButton>
+  ) : (
     <Box css={{ backgroundColor: '$bgDefault' }}>
       <Initiator />
```
six7, line 61: We need to make sure `<Initiator>` is always rendered. Right now it triggers a complete reload of the page as the component is being remounted, which results in `Fetching remote tokens...` when you maximize the plugin.

Instead, place the maximize button right after `LoadingBar`, conditionally rendering either the maximize button or the `<Box...>`

six7 (changes requested):
I noticed that we seem to fetch remote tokens when you click maximize? Let's not do that. This comes from `<Initiator> currently being rendered conditionally.

six7 (approved):

src/app/components/Navbar.tsx
```diff
@@ -18,8 +19,10 @@ import { Tabs } from '@/constants/Tabs';
 import Stack from './Stack';
 import { TabButton } from './TabButton';
 import { NavbarUndoButton } from './NavbarUndoButton';
+import Minimize from '../assets/minimize.svg';
+import useMinimizeWindow from './useMinimizeWindow';
 
-const transformProviderName = (provider: StorageProviderType) => {
+const transformProviderName = (provider) => {
```
six7, line 25: why remove the type?

src/app/components/Navbar.tsx
```diff
@@ -39,8 +42,22 @@ export const Navbar: React.FC = () => {
...
   const usedTokenSet = useSelector(usedTokenSetSelector);
+  const { handleResize } = useMinimizeWindow();
+
   const { pullTokens, pushTokens } = useRemoteTokens();
 
+  const StyledButton = styled('button', {
+    all: 'unset',
+    border: 'none',
+    padding: '$1',
+    marginRight: '10px',
```
six7, line 53:
```suggestion
    marginRight: '$3',
```

src/app/components/PluginResizer.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const StyledButton = styled('button', {
+  all: 'unset',
+  border: 'none',
+  padding: '$2',
+  marginLeft: '$4',
+  borderRadius: '$button',
+  cursor: 'pointer',
+  '&:hover, &:focus': {
+    boxShadow: 'none',
```
six7, line 14: if we remove focus styles we should provide another visual cue. otherwise if you use the keyboard only you will not see what's focused.

i'd say change backgroundcolor to `$bgSubtle`

src/app/components/Navbar.tsx
```diff
@@ -39,8 +42,22 @@ export const Navbar: React.FC = () => {
...
+    borderRadius: '$button',
+    cursor: 'pointer',
+    '&:hover, &:focus': {
+      boxShadow: 'none',
```
six7, line 57: same as below, add backgroundColor: '$bgSubtle'

src/app/store/models/settings.tsx
```diff
@@ -57,11 +59,23 @@ export const settings = createModel<RootModel>()({
...
         ...state,
         uiWindow: {
+          ...state.uiWindow,
           width: payload.width,
           height: payload.height,
         },
       };
     },
+    setMinimizePluginWindow(state, payload: { isMinimized: boolean; width: number; height: number }) {
+      track('Stored Window Size', { width: payload.width, height: payload.height });
```
six7, line 69:
```suggestion
      track('Minimized plugin');
```

six7 (changes requested):
Noticed a few more things that we should change

src/app/components/Navbar.tsx
```diff
@@ -103,6 +120,9 @@ export const Navbar: React.FC = () => {
             </Tooltip>
           </>
         )}
+        <StyledButton type="button" onClick={handleResize}>
```
six7, line 123: also, let's add a <Tooltip> that wraps this, much like we did for `NavbarUndoButton` - with a text `Minimize plugin` : `Maximize plugin`

six7 (approved):

PR #619, opened 2022-04-19
fix: token sorting bug

LiamMartens:


src/app/components/TokenListing.tsx
```diff
@@ -154,4 +154,11 @@ const TokenListing: React.FC<Props> = ({
   );
 };
 
-export default React.memo(TokenListing, isEqual);
+// @README the memo props check used to be a deep equals
+// but because the token sorting is done based on the order of an object
+// it comes as back as equals since object key order is disregarded in lodash's
+// isEqual check.
+// @TODO we should probably not rely on object key order for sorting anyways
```
six7, line 161: agree, the w3c draft itself discarded array based tokens which would've been easier to sort, however much harder to maintain.

any solution we come up with should be as easy as just reordering an object though.

six7 (approved):
Nice!

PR #621, opened 2022-04-19
deep linking aliases

swordEdge:


six7:
Can you add tests for this scenario? The alias files should already have some tests setup, let's extend them.

six7:
fyi: theres some failing tests @swordEdge

swordEdge:
> Can you add tests for this scenario? The alias files should already have some tests setup, let's extend them.

done!

swordEdge:
> * Don't show the red dot

fixed!
> * ideally, it would show the fully resolved value, not $16. In that case, 16px.

fixed!

src/utils/alias/getAliasValue.ts
```diff
@@ -3,6 +3,7 @@ import { convertToRgb } from '../color';
 import { findReferences } from '../findReferences';
 import { isSingleTokenValueObject } from '../is';
 import { checkAndEvaluateMath } from '../math';
+import { theme } from '@/stitches.config';
```
six7, line 6: i think this is a mistake? the stitches config is not required here

src/utils/alias/getAliasValue.ts
```diff
@@ -38,7 +39,11 @@ export function getAliasValue(token: SingleToken | string | number, tokens: Sing
       tokenReferences.forEach((reference, index) => {
         const resolvedReference = resolvedReferences[index];
         const stringValue = String(resolvedReference);
-        const resolved = checkAndEvaluateMath(stringValue);
+        let resolved = checkAndEvaluateMath(stringValue);
+        if (String(resolved).startsWith("$")) {
+          const sizeProperty = String(resolved).slice(1, String(resolved).length);
+            resolved = theme.space[`${sizeProperty}`].value;
```
six7, line 45: why is this here? our stitches theme config is definitely not something our users would want to alias?

six7 (changes requested):
i think there's some unintended changes here, let's not include our stitches theme config in user's aliases

i also noticed it currently still shows the red dot if the user used the $ notation, the preview doesnt work as well if the user used the $ notation.

image
image
image
image

src/utils/alias/getAliasValue.ts
swordEdge: Fixed!

src/utils/alias/getAliasValue.ts
swordEdge: Fixed!

swordEdge:
> i think there's some unintended changes here, let's not include our stitches theme config in user's aliases

Fixed!

swordEdge:
> Can you add tests for this scenario? The alias files should already have some tests setup, let's extend them.

Done!

src/utils/alias/getAliasValue.ts
```diff
@@ -9,22 +9,58 @@ export function getAliasValue(token: SingleToken | string | number, tokens: Sing
...
+          else 
+            nameToLookFor = ref.substring(1);
+
+          if (
+            (typeof token === 'object' && nameToLookFor === token.name) || 
+            nameToLookFor === token
+          ) 
+            return null;
+
+          type TokenNameNodeType = string | undefined;
```
six7, line 33: This can probably move to the top of the file or to its own file?

six7 (approved):
Looking good! I had one comment regarding where that type is being declared. Other than that this works great!

LiamMartens (approved):

src/utils/alias/getAliasValue.ts
swordEdge: Fixed!

PR #622, opened 2022-04-20
feat(Navbar): Combine Sync and Settings Tab

srinivas0210:
* Combine Sync and Settings Tab
* Removed Sync Tab
* Now on, we redirect users to 'settings' tab instead of 'syncsettings' tab

### DIFF
--- a/src/app/components/App.tsx
+++ b/src/app/components/App.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
-import SyncSettings from './SyncSettings';
 import Settings from './Settings';
 import Inspector from './Inspector';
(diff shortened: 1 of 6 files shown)


six7 (changes requested):
If you search for 'syncsettings' there's one more appearance in `src/constants/Tabs.ts` which we'd need to remove.

Also, Padding seems to be off now. 
image

Remove the padding from `<SyncSettings>` as that's being taken care of in the parent wrapper

The scrollbar looks different, this is because the class `content scroll-container` is responsible for handling the style of the scrollbar, you only set `overflowY: auto`

Add class `content scroll-container` to the Settings parent wrapper where you've now set overflowY

image

six7 (approved):

PR #632, opened 2022-04-22
Ft 49 reference shadow and typography tokens 565

swordEdge:


src/app/components/AliasBox.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { styled } from '@/stitches.config';
+
+const AliasStyleBox = styled('div', {
+  backgroundColor: '#f5f5f5',
```
six7, line 5: make sure to use tokens in here, we should not use raw hex values as this is not theme-able for e.g. a future dark theme.

same to other elements in this block. e.g. use '$bgDefault' here (tokens can be found in stitches.config.js)

src/app/components/BoxShadowInput.tsx
```diff
@@ -126,21 +131,21 @@ function SingleShadowInput({
     >
       <Box css={{ display: 'flex', justifyContent: 'space-between', alignItems: 'center' }} ref={ref}>
         {isMultiple && (
-        <Box css={{ display: 'flex', width: '$space$8' }}>
-          <IconButton tooltip="Click to drag" icon={<IconGrabber />} data-handler-id={handlerId} />
-        </Box>
+          <Box css={{ display: 'flex', width: '$space$8' }}>
```
six7, line 134: you should be able to just use `$8` here, which will then use a token from the sizes category

src/app/components/BoxShadowInput.tsx
```diff
@@ -197,36 +206,104 @@ export default function BoxShadowInput({
     }
   };
 
+  const handleMode = () => {
```
six7, line 209: this should also be a useCallback, right?

src/app/components/BoxShadowInput.tsx
```diff
@@ -197,36 +206,104 @@ export default function BoxShadowInput({
...
+    setMode((mode === 'input') ? 'alias' : 'input');
+    setAlias('');
+    setValue(newToken);
+  };
+
+  const handleAliasChange = (e) => {
```
six7, line 215: this should also be a useCallback, right?

src/app/components/ResolvedValueBox.tsx
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <Box css={{
+      display: 'flex', flexDirection: 'column', gap: '$2',
+    }}
+    >
+      {
+        (resolvedValue && isJson(resolvedValue)) ? (
+          (resolvedValue?.toString().charAt(0) === '[') ? (
+            <AliasBox>
+              <Box css={{ display: 'grid', margin: '12px', marginRight: '60px' }}>
```
six7, line 44: can we use tokens here instead of raw px values?

src/app/components/ResolvedValueBox.tsx
```diff
@@ -0,0 +1,85 @@
...
+                {
+                  findReferences(resolvedValue).map((value, index) => <p>{index + 1}</p>)
+                }
+              </Box>
+              <Box css={{ display: 'grid', margin: '12px' }}>
```
six7, line 49: same here

src/app/components/ResolvedValueBox.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    }
+                    return <p key={index}>{properties}</p>;
+                  })
+                }
+              </Box>
+            </AliasBox>
+
+          ) : (
+            <AliasBox>
+              <Box css={{ display: 'grid', margin: '12px', marginRight: '60px' }}>
```
six7, line 65: same here

src/app/components/ResolvedValueBox.tsx
```diff
@@ -0,0 +1,85 @@
...
+                {
+                  Object.keys(JSON.parse(resolvedValue)).map((key, index) => <p>{key}</p>)
+                }
+              </Box>
+              <Box css={{ display: 'grid', margin: '12px' }}>
```
six7, line 70: same here

six7 (changes requested):
Bit of functional feedback:
Once the token is created, the link seems to be lost. So when I afterwards go and edit the token, it does only show up the raw values, instead of it being in alias mode and using the alias token.

it should basically save like this:

`"value": "{name.of.alias}"`

right now it saves the whole object.

also, it should show alias mode when an alias is used, currently it shows the whole typography input showing raw values

https://user-images.githubusercontent.com/4548309/164891840-f4bc1ea0-22ad-4830-bf29-648a266fb408.mp4

src/app/components/BoxShadowInput.tsx
swordEdge: fixed

src/app/components/BoxShadowInput.tsx
swordEdge: fixed

src/app/components/ResolvedValueBox.tsx
swordEdge: fixed

src/app/components/ResolvedValueBox.tsx
swordEdge: fixed

src/app/components/ResolvedValueBox.tsx
swordEdge: fixed

src/app/components/ResolvedValueBox.tsx
swordEdge: fixed

swordEdge:
> Bit of functional feedback: Once the token is created, the link seems to be lost. So when I afterwards go and edit the token, it does only show up the raw values, instead of it being in alias mode and using the alias token.

> done

LiamMartens (approved):

PR #633, opened 2022-04-22
Fix numerical font weights

six7:
Fixes a bug that prevented users form using numerical font weights.

For example, a user that had `Museo Sans` that had a font weight of `500` could not use that combination. As Figma expects a string for `fontWeight`, but we returned `fontWeight` as we received it, in that case a number.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "figma-tokens",
     "version": "1.0.0",
-    "plugin_version": "94",
+    "plugin_version": "95",
     "description": "Figma Tokens",
     "license": "MIT",
     "scripts": {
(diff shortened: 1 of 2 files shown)


LiamMartens (approved):

leovogel:
Is this feature available in v99? I tried setting a fontWeight to a number value as that's how my engineering partner (I'm a designer) wants the JSON setup. But it doesn't update the styles in Figma. Using the text value of the font weight does update the style in Figma.

six7:
You'll have to use the same name as the weight dropdown in Figma itself shows. This fix was for fonts who are using these numbered weights, but it was no automatic conversion from eg 500 to Regular.

It's easy to map from 'Regular' to '500' in code, but there's no good automatic conversion the other way around, as each font can use whatever weight name it likes. Hopefully Figma supports this kind of mapping in the future.

PR #639, opened 2022-04-25
Token Transformer add `resolveReferences`option

### LINKED ISSUE #305: Keep aliases when transforming plugin JSON to Style Dictionary JSON
First of all love what you've created!

I'm using [Token Transformer as described in the docs](https://docs.tokens.studio/sync/github#7-how-to-use-tokens-stored-in-github-in-development)

It does state that:
> You can then use it to generate tokens to a JSON that is readable by Style Dictionary, without any aliases, without any math.

While this is great for the math part (e.g. `{baseUnit} * 2`) as SD doesn't have support for that, I would like it to keep the regular aliases (e.g. `{color.primary}`) in the output to allow for SD's [Output references](https://amzn.github.io/style-dictionary/#/version_3?id=output-references) introduced in v3.

This could either be added as CLI option or - even better - just changed to keep the aliases in the output, as SD already have support for that and can either be translated to actual values (SD default) or configured to be outputted as references.

**Describe alternatives you've considered**
🤷🏻 

**Additional context**

jakobe:
Closes #305

### DIFF
--- a/src/utils/__tests__/__snapshots__/convertTokensObjectToResolved.test.ts.snap
+++ b/src/utils/__tests__/__snapshots__/convertTokensObjectToResolved.test.ts.snap
@@ -180,6 +180,23 @@ Object {
       "value": "1 * {sizing.base}",
     },
   },
+  "text": Object {
+    "size": Object {
+      "base": Object {
+        "type": "fontSize",
+        "value": "16",
+      },
+      "default": Object {
+        "description": "Should NOT be resolved",
+        "type": "fontSize",
+        "value": "{text.size.base}{text.size.unit}",
+      },
+      "unit": Object {
+        "type": "fontSize",
+        "value": "px",
+      },
+    },
+  },
 }
 `;
 
@@ -252,6 +269,23 @@ Object {
       "value": 2,
     },
   },
+  "text": Object {
+    "size": Object {
+      "base": Object {
+        "type": "fontSize",
+        "value": 16,
+      },
+      "default": Object {
+        "description": "Should be resolved",
+        "type": "fontSize",
+        "value": "16px",
+      },
+      "unit": Object {
+        "type": "fontSize",
+        "value": "px",
+      },
+    },
+  },
 }
 `;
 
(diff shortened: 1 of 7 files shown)


mkitzmann (commented):
The `resolveReference` flag and its options should probably also be added to the README: https://github.com/six7/figma-tokens/blob/main/token-transformer/README.md

jakobe:
> Thank you @jakobe for this PR!
> 
> The `resolveReference` flag and its options should probably also be added to the README: https://github.com/six7/figma-tokens/blob/main/token-transformer/README.md

Good catch @mkitzmann 👌🏻 
I've updated the PR 👍🏻

six7:
Nice, this is awesome, thank you! 🎉

I'll get this merged and released this week 🙏

six7 (changes requested):
One thing that doesn't work for me yet is running it with `resolveReferences=false`.

It works fine for `resolveReferences=math`, but when i set it to false it still resolves those references.

I was able to pinpoint this to the type definition of `boolean | 'math'` which interprets `false` as a string, which is then true. If I set it to just `boolean` things work (but now math no longer works)

jakobe:
> One thing that doesn't work for me yet is running it with `resolveReferences=false`.
> 
> It works fine for `resolveReferences=math`, but when i set it to false it still resolves those references.
> 
> I was able to pinpoint this to the type definition of `boolean | 'math'` which interprets `false` as a string, which is then true. If I set it to just `boolean` things work (but now math no longer works)

Hi @six7 thx for catching this in the review! I wasn't paying attention to the CLI specifics, as the unit tests worked as expected, but I see now that yargs doesn't have support for union types (it converts the input to a string in all cases when type is `boolean | 'math'`).

I've added a fix that checks the input for `'true'|'false'` and converts back to a primitive boolean in that case.

six7:
Perfect, thanks! I'll give this a go later today. 🎉

six7 (approved):

six7:
Thank you! 🙏  I won't get around to releasing it today, but will do tomorrow.

jakobe:
> Thank you! 🙏 I won't get around to releasing it today, but will do tomorrow.

Awesome - thank you too 🙏🏻😊

mkitzmann:
Thank you for merging this PR 🙏 
@six7 It seems there is no new version released yet. Are there any issues regarding the release?

six7:
> Thank you for merging this PR 🙏 @six7 It seems there is no new version released yet. Are there any issues regarding the release?

Released it yesterday. Sorry for the delay!

PR #647, opened 2022-04-28
[License Key / Payment] Update & merge current version

roppazvan:
Demo: https://drive.google.com/file/d/1SFe62qT0mWNzwS502N4EkjDKVWQg3mgD/view?usp=sharing
* Added a new store section for user-related data
* Created an add license key component
* Add license key
* Validate license key on plugin open
* Error notifications
* Remove license key


LiamMartens (approved):

six7 (changes requested):
This works really well! Only 1 feedback:

When you enter a wrong license key, IMO it should validate on entering. Only if it’s valid should it save the license. Right now it saves it right away, even though its a wrong license key. I then get the option to remove license, but that doesn’t make sense as it shouldn’t have been saved in the first place. 

Let's make it so: It validates license key on pressing `Add licenseKey`, if valid, it saves it. If not, it doesn't save it and gives you the error.

roppazvan:
> 

Made the changes. Also had some merge conflicts. Fixed that too.

six7 (approved):

PR #655, opened 2022-04-30
Optimize tokens

six7:


src/app/components/TokenListing.tsx
```diff
@@ -106,30 +115,16 @@ const TokenListing: React.FC<Props> = ({
...
-              type="button"
-              onClick={() => {
-                showNewForm({});
-              }}
-            >
-              <Icon name="add" />
-            </button>
-          </Tooltip>
+          <IconButton
+            data-cy="button-add-new-token"
```
six7, line 122:
```suggestion
            dataCy="button-add-new-token"
```

src/app/components/TokenTree.tsx
```diff
@@ -59,17 +59,16 @@ const TokenTree: React.FC<Props> = ({
...
-                <Tooltip label="Add a new token" side="left">
-                  <button
+                <div className="opacity-0 group-hover:opacity-100 focus:opacity-100">
+                  <IconButton
+                    icon={<AddIcon />}
+                    tooltip="Add a new token"
+                    tooltipSide="left"
+                    onClick={handleShowNewForm}
                     disabled={editProhibited}
                     data-cy="button-add-new-token-in-group"
```
six7, line 69:
```suggestion
                    dataCy="button-add-new-token-in-group"
```

src/app/components/ImportedTokensDialog.tsx
```diff
@@ -75,7 +69,15 @@ export default function ImportedTokensDialog() {
   const [newTokens, setNewTokens] = React.useState(importedTokens.newTokens);
   const [updatedTokens, setUpdatedTokens] = React.useState(importedTokens.updatedTokens);
 
-  const handleCreateNewClick = () => {
+  const handleSetUpdatedTokens = React.useCallback((tokens: SingleToken[]) => {
```
LiamMartens, line 72: @six7 good to see useCallbacks! :smile:

src/app/components/ImportedTokensDialog.tsx
```diff
@@ -192,8 +194,8 @@ export default function ImportedTokensDialog() {
                 value={token.value}
                 description={token.description}
                 updateAction="Create"
-                removeToken={() => setNewTokens(newTokens.filter((t) => t.name !== token.name))}
-                updateToken={() => handleCreateSingleClick(token)}
+                removeToken={handleSetNewTokens(newTokens.filter((t) => t.name !== token.name))}
+                updateToken={handleCreateSingleClick(token)}
```
LiamMartens, line 198: Did you mean to write these like this? Now the handlers will be executed on render, which presumably is not the intention. The handlers itself should just use the `token` and `newTokens` variables and then you can just pass the functions into the props as is `removeToken={handleSetewTokens}`.

src/app/components/ImportedTokensDialog.tsx
```diff
@@ -224,8 +226,8 @@ export default function ImportedTokensDialog() {
                 description={token.description}
                 oldDescription={token.oldDescription}
                 updateAction="Update"
-                removeToken={() => setUpdatedTokens(updatedTokens.filter((t) => t.name !== token.name))}
-                updateToken={() => handleUpdateSingleClick(token)}
+                removeToken={handleSetUpdatedTokens(updatedTokens.filter((t) => t.name !== token.name))}
```
LiamMartens, line 229: Same thing as above

LiamMartens (changes requested):

src/app/components/ImportedTokensDialog.tsx
six7: Thanks, hope I got that right with this one. The handler itself now uses the token, however I still have to pass the current token in, like this:

```
  const handleIgnoreExistingToken = React.useCallback((token) => () => {
    setUpdatedTokens((updatedTokens.filter((t) => t.name !== token.name)));
  }, [setUpdatedTokens, updatedTokens]);

  ...
  removeToken={handleIgnoreExistingToken(token)}
```

LiamMartens: Then this actually needs to be separated into it's own component taking the `token` as a prop. Then you can create handlers with those tokens specifically.

LiamMartens (changes requested):

src/app/components/ImportedTokensDialog.tsx
six7: Got it, fixed in newest commit. 🙏

src/types/tokens/ImportToken.ts
```diff
@@ -0,0 +1,6 @@
+import { SingleToken } from './SingleToken';
+
+export type ImportToken = SingleToken & {
```
LiamMartens, line 3: Could you just change this with
```
export type ImportToken<Named = true> = SingleToken<Named, {
  oldValue?: string;
  oldDescription?: string;
}>
```

Keeps the typing more strict

LiamMartens (approved):
@six7 just one small comment - other than that good to go

PR #659, opened 2022-05-02
Ft 54 duplicate token group 515

MightyDevKnight:


src/app/components/TokenGroupHeading.tsx
```diff
@@ -1,36 +1,125 @@
...
 
+  const handleDuplicate = React.useCallback(() => {
+    const oldName = path.split('.').pop()?.toString() || '';
+    const newPath = path.slice(0, path.length - oldName.length);
+    duplicateGroup({ newPath, oldName });
+    setIsDuplicated(true);
+    setShowRenameTokenGroupField(true);
+  }, [path, duplicateGroup, setShowRenameTokenGroupField]);
+
+  const handleRenameTokenGroupSubmit = React.useCallback((e: React.FormEvent<HTMLFormElement>) => {
```
LiamMartens, line 40: @MightyDevKnight do you have eslint installed? It looks like eslint is complaining quite a bit on this so I suspect you don't

src/app/components/TokenGroupHeading.tsx
```diff
@@ -1,36 +1,125 @@
...
+          </ContextMenuItem>
+          <ContextMenuItem disabled={editProhibited} onSelect={handleRenameTokenGroup}>
+            Rename
+          </ContextMenuItem>
+          <ContextMenuItem disabled={editProhibited} onSelect={handleDuplicate}>
+            Duplicate
+          </ContextMenuItem>
+        </ContextMenuContent>
+      </ContextMenu>
+      <Modal isOpen={showRenameTokenGroupField} close={() => setShowRenameTokenGroupField(false)}>
```
LiamMartens, line 93: Can you just make this a useCallback as well

src/app/components/TokenGroupHeading.tsx
```diff
@@ -1,36 +1,125 @@
...
+        <Stack direction="column" justify="center" gap={4} css={{ textAlign: 'center' }}>
+          <Heading size="small">Rename {tokenGroupMarkedForChange}</Heading>
+          <form onSubmit={handleRenameTokenGroupSubmit}>
+            <Stack direction="column" gap={4}>
+              <Input
+                full
+                value={newTokenGroupName}
+                onChange={(e) => handleNewTokenGroupNameChange(e.target.value)}
```
LiamMartens, line 101: Same with this one

src/app/components/TokenGroupHeading.tsx
```diff
@@ -1,36 +1,125 @@
...
+                type="text"
+                name="tokengroupname"
+                required
+              />
+              <Stack direction="row" gap={4}>
+                <Button variant="secondary" size="large" onClick={() => setShowRenameTokenGroupField(false)}>
```
LiamMartens, line 107: And this one

src/app/components/TokenSetSelector.tsx
```diff
@@ -78,7 +78,9 @@ export default function TokenSetSelector() {
   const handleRenameTokenSet = React.useCallback((tokenSet: string) => {
     track('Renamed token set');
     handleNewTokenSetNameChange(tokenSet);
+    console.log('tokenSet', tokenSet);
```
LiamMartens, line 81: I assume these don't need to be here

src/app/store/models/tokenState.tsx
```diff
@@ -339,43 +332,93 @@ export const tokenState = createModel<RootModel>()({
...
-                value: Object.entries(token.value).reduce<Record<string, string | number>>((a, [k, v]) => {
-                  a[k] = replaceReferences(v.toString(), data.oldName, data.newName);
-                  return a;
-                }, {}),
-              } as SingleToken;
-            }
+    duplicateTokenGroup: (state, data: DuplicateTokenGroupPayload) => {
+      const { parent, path, oldName } = data;
+      const selectedGroup = state.tokens[parent].filter((token) => token.name.startsWith(`${path}${oldName}.`));
+      const _duplicatedGroup = selectedGroup.map((token) => {
```
LiamMartens, line 340: Eslint will complain about _ prefixed variables, if you could just fix the eslint errors that would be  great

src/app/store/models/tokenState.tsx
```diff
@@ -339,43 +332,93 @@ export const tokenState = createModel<RootModel>()({
...
-            if (typeof token.value === 'object') {
-              return {
-                ...token,
-                value: Object.entries(token.value).reduce<Record<string, string | number>>((a, [k, v]) => {
-                  a[k] = replaceReferences(v.toString(), data.oldName, data.newName);
-                  return a;
-                }, {}),
-              } as SingleToken;
-            }
+    duplicateTokenGroup: (state, data: DuplicateTokenGroupPayload) => {
```
LiamMartens, line 337: Something in this PR broke the tokenState type inferrering resulting in `dispatch.tokenState.updateDocument` not being seen as a valid method anylonger; could you just see if you can fix that again?

src/app/store/models/tokenState.tsx
```diff
@@ -339,43 +332,93 @@ export const tokenState = createModel<RootModel>()({
...
 
+      return {
+        ...state,
+        tokens: {
+          ...state.tokens,
+          [parent]: [...state.tokens[parent], ..._duplicatedGroup],
+        },
+      };
+    },
+    renameTokenGroup: (state, data: RenameTokenGroupPayloads) => {
```
LiamMartens, line 357: I would also like to start moving the reducers into their own files so we don't have a massive state file. I already did this for a few of them in `reducers/tokenState`; would be great if we can do that here.

LiamMartens (changes requested):

PR #672, opened 2022-05-04
Ft 54 duplicate token group 515

MightyDevKnight:


six7 (changes requested):
I think when you duplicate you should be able to just press enter to use the new name (as its a new group name?)

Also, there's still something strange going on here, when I duplicate a group and later delete that it sometime doesn't create/rename that. Could it be there's still something saved as in oldgroupname?

In addition there's a grey line right next to the group name, this seems like a mistake.

https://user-images.githubusercontent.com/4548309/166830582-83bd4075-7c53-42f4-bb44-68bee28c67b1.mp4

PR #673, opened 2022-05-04
Ft 46 autocomplete add to remap and all token inputs

swordEdge:


six7 (changes requested):
Bit of functional feedback:

- I think we need to change how the Autocomplete works. With this current implementation it's possible e.g. in a typography token to open up the popover multiple times, leading to a bad ux. We don't need to address this in this branch, but should do a followup soon after.
image

- For typography tokens and shadow tokens it currently show [object Object] as a value. It would probably be good to show the shorthand (the one we also show in the tooltip) here. Maybe we can abstract that so we can use the same shorthand in both this alias value and this autocomplete.

image

swordEdge:
> With this current implementation it's possible e.g. in a typography token to open up the popover multiple times, leading to a bad ux

I tried to open up the popover multiple times but there was no bad ux error similar to yours. I think it was already fixed.


> For typography tokens and shadow tokens it currently show [object Object] as a value. It would probably be good to show the shorthand

done

src/app/components/DownshiftInput.tsx
```diff
@@ -135,6 +136,27 @@ const DownshiftInput: React.FunctionComponent<DownShiftProps> = ({
     [resolvedTokens, filteredValue, type],
   );
 
+  const resolveValue = useCallback((token: SingleToken) => {
```
LiamMartens, line 139: There's no need to mutate `acc` when reducing since the returned value will be the next value anyways. This block also gives ESLint errors anyways - make sure you have ESLint enabled in your VS Code.

```
      if (Array.isArray(token.value)) {
        returnValue = token.value.reduce<string>((totalAcc, item) => {
          const singleReturnValue = Object.entries(item).reduce<string>((acc, [,value]) => (
            `${acc}${value.toString()}/`
          ), '');
          return `${totalAcc}${singleReturnValue},`;
        }, '');
      } else {
        returnValue = Object.entries(token.value).reduce<string>((acc, [,value]) => (
          `${acc}${value.toString()}/`
        ), '');
      }
```

src/app/components/InspectorTokenSingle.tsx
```diff
@@ -10,7 +10,16 @@ import { Dispatch } from '../store';
...
+import Button from './Button';
+import Heading from './Heading';
+import DownshiftInput from './DownshiftInput';
+import { StyledIconDisclosure, StyledInputSuffix } from './StyledInputSuffix';
+import Modal from './Modal';
+import Stack from './Stack';
+import ColorPicker from './ColorPicker';
+import { Properties } from '@/constants/Properties';
 
 export default function InspectorTokenSingle({
```
LiamMartens, line 24: A few ESLint errors on this file, but they are auto-fixable. Can you just format it using ESLint

src/app/components/InspectorTokenSingle.tsx
```diff
@@ -22,19 +31,66 @@ export default function InspectorTokenSingle({
...
+  }, [showAutoSuggest]);
+
+  const onConfirm = React.useCallback(() => {
+    handleRemap(token.category, token.value, newTokenName);
+    setShowDialog(false);
+  }, [token, handleRemap, newTokenName, mappedToken]);
+
+  const handleClick = React.useCallback(() => {
+    setShowDialog(true);
+  }, [showDialog]);
```
LiamMartens, line 88: It's not necessary to specify `showDialog` itself as a dependency here as it's not being used. Same with the other callbacks, you only need to specify the variables you actually use within the function body

src/app/components/SingleBoxShadowInput.tsx
```diff
@@ -0,0 +1,269 @@
+import React from 'react';
```
LiamMartens, line 1: Same here, please format it with ESLint

src/app/components/SingleBoxShadowInput.tsx
```diff
@@ -0,0 +1,269 @@
...
+  }, [shadowItem]);
+
+  const [showAutoSuggest, setShowAutoSuggest] = React.useState<Array<boolean>>(defalutShowAutoSuggest);
+  const [inputHelperOpen, setInputHelperOpen] = React.useState<boolean>(false);
+
+  const handleToggleInputHelper = React.useCallback(() => setInputHelperOpen(!inputHelperOpen), [inputHelperOpen]);
+
+  const handleAutoSuggest = React.useCallback((keyIndex: number) => (() => changeAutoSuggest(keyIndex)), [showAutoSuggest]);
+
+  const changeAutoSuggest = React.useCallback((index: number) => {
```
LiamMartens, line 70: Can you move the `changeAutoSuggest` function before the `handleAutoSuggest` one. The definition order is now incorrect.

src/app/components/SingleBoxShadowInput.tsx
```diff
@@ -0,0 +1,269 @@
...
+    const newShowAutoSuggest = [...showAutoSuggest];
+    newShowAutoSuggest[index] = !newShowAutoSuggest[index];
+    setShowAutoSuggest(newShowAutoSuggest);
+  }, [showAutoSuggest]);
+
+  const handleCloseAutoSuggest = React.useCallback((keyIndex: number) => (() => closeAutoSuggest(keyIndex)), [showAutoSuggest]);
+
+  const closeAutoSuggest = React.useCallback((index: number) => {
```
LiamMartens, line 78: Same here, order should be changed

src/app/components/SingleBoxShadowInput.tsx
```diff
@@ -0,0 +1,269 @@
...
+                name={key}
+                key={`boxshadow-input-${index}-${keyIndex}`}
+                value={String(shadowItem[key as keyof typeof propertyTypes])}
+                type={propertyTypes[key as keyof typeof propertyTypes]}
+                label={key}
+                showAutoSuggest={showAutoSuggest[keyIndex]}
+                resolvedTokens={resolvedTokens}
+                handleChange={onChange}
+                setShowAutoSuggest={handleCloseAutoSuggest(keyIndex)}
+                setInputValue={(newInputValue: string) => handleBoxshadowDownShiftInputChange(newInputValue, key)}
```
LiamMartens, line 236: Can we make a useCallback for this? It would mean seperating this into a new file.
Now it does give an ESlint error

src/app/components/TypographyInput.tsx
```diff
@@ -1,47 +1,94 @@
-import React, { useState } from 'react';
+import React from 'react';
```
LiamMartens, line 1: Same thing, please run the ESLint formatter

src/app/components/TypographyInput.tsx
```diff
@@ -1,47 +1,94 @@
...
+    return null;
   }, [internalEditToken, resolvedTokens]);
 
+  const handleMode = React.useCallback(() => {
+    const changeMode = (mode === 'input') ? 'alias' : 'input';
+    setMode(changeMode);
+    setAlias('');
+  }, [mode]);
+
+  const handleAutoSuggest = React.useCallback((keyIndex: number) => (() => changeAutoSuggest(keyIndex)), [showAutoSuggest]);
```
LiamMartens, line 76: Also same comment here, the order of these functions need to be fixed so they get declared correctly.

src/app/components/TypographyInput.tsx
```diff
@@ -66,40 +113,54 @@ export default function TypographyInput({
...
-              required
+              key={`typography-input-${index}`}
+              value={value}
+              type={properties[key as keyof typeof properties]}
+              label={key}
+              showAutoSuggest={showAutoSuggest[index]}
+              resolvedTokens={resolvedTokens}
+              handleChange={handleTypographyChange}
+              setShowAutoSuggest={handleCloseAutoSuggest(index)}
+              setInputValue={(newInputValue: string) => handleTypographyDownShiftInputChange(newInputValue, key)}
```
LiamMartens, line 127: And same comment here - please make a useCallback for this, or you could also useMemo the whole map and then Im happy to eslint ignore the callback

LiamMartens (changes requested):

src/app/components/InspectorTokenSingle.tsx
swordEdge: fixed

src/app/components/DownshiftInput.tsx
swordEdge: fixed

src/app/components/InspectorTokenSingle.tsx
swordEdge: fixed

src/app/components/InspectorTokenSingle.tsx
swordEdge: fixed

src/app/components/SingleBoxShadowInput.tsx
swordEdge: fixed

src/app/components/SingleBoxShadowInput.tsx
swordEdge: fixed

src/app/components/SingleBoxShadowInput.tsx
swordEdge: fixed

src/app/components/TypographyInput.tsx
swordEdge: fixed

src/app/components/TypographyInput.tsx
swordEdge: fixed

src/app/components/SingleBoxShadowInput.tsx
swordEdge: done

src/app/components/TypographyInput.tsx
swordEdge: done

src/app/components/InspectorTokenSingle.tsx
```diff
@@ -22,19 +31,67 @@ export default function InspectorTokenSingle({
...
+  const handleChange = React.useCallback<React.ChangeEventHandler<HTMLInputElement>>((e) => {
+    e.persist();
+    setNewTokenName(e.target.value);
+  }, []);
+
+  const handleColorValueChange = React.useCallback((color: string) => {
+    setNewTokenName(color);
+  }, []);
+
+  const handleDownShiftInputChange = React.useCallback((newInputValue: string) => {
```
six7, line 64: Can we make sure to just 1 component for DownshiftInput and not have the logic spread throughout the app? This should all be part of the downshiftinput Component ideally

src/app/components/InspectorTokenSingle.tsx
```diff
@@ -22,19 +31,67 @@ export default function InspectorTokenSingle({
...
+    if (newInputValue.charAt(0) === '$') setNewTokenName(newInputValue.slice(1, newInputValue.length));
+    if (newInputValue.charAt(0) === '{') setNewTokenName(newInputValue.slice(1, newInputValue.length - 1));
+    else setNewTokenName(newInputValue);
+  }, []);
+
+  const handleToggleInputHelper = React.useCallback(() => {
```
six7, line 70: I'd love to have the downshift component itself handle the logic of when the popover is open or not

src/app/components/InspectorTokenSingle.tsx
```diff
@@ -71,14 +127,73 @@ export default function InspectorTokenSingle({
...
+                      setInputValue={handleDownShiftInputChange}
+                      placeholder={
+                        Properties[token.category as keyof typeof Properties] === 'fill' ? '#000000, hsla(), rgba() or {alias}' : 'Value or {alias}'
+                      }
+                      prefix={
+                        Properties[token.category as keyof typeof Properties] === 'fill' && (
+                          <button
+                            type="button"
+                            className="block w-4 h-4 rounded-sm cursor-pointer shadow-border shadow-gray-300 focus:shadow-focus focus:shadow-primary-400"
+                            style={{ background: newTokenName, fontSize: 0 }}
```
six7, line 166: We cant use specific color values anymore since we have dark mode support. We need to use tokens eg $borderMuted as that value will adapt when user has dark mode active

src/app/components/SingleBoxShadowInput.tsx
```diff
@@ -0,0 +1,248 @@
...
+  index: number;
+  id: string;
+  type: string;
+}
+
+enum ItemTypes {
+  CARD = 'card',
+}
+
+const propertyTypes = {
```
six7, line 28: This is duplicative of the prop being used in the other component above, cant we combine that?

src/app/components/SingleBoxShadowInput.tsx
```diff
@@ -0,0 +1,248 @@
...
+}) {
+  const seed = useUIDSeed();
+  const defalutShowAutoSuggest = React.useMemo(() => {
+    if (shadowItem && typeof shadowItem === 'object') {
+      return Object.entries(shadowItem).map(() => false, {});
+    }
+    return [false];
+  }, [shadowItem]);
+
+  const [showAutoSuggest, setShowAutoSuggest] = React.useState<Array<boolean>>(defalutShowAutoSuggest);
```
six7, line 63: I'd really prefer to have all the autosuggeat logic living inside the DownShiftInput component and not spread around so many components. That component should be responsible for open/close state, current value etc.

Right now i noticed you have duplicated that logic to many parts when it would only need to be living in 1 component

src/app/components/TypographyInput.tsx
```diff
@@ -1,47 +1,91 @@
...
+    return null;
   }, [internalEditToken, resolvedTokens]);
 
+  const handleMode = React.useCallback(() => {
+    const changeMode = (mode === 'input') ? 'alias' : 'input';
+    setMode(changeMode);
+    setAlias('');
+  }, [mode]);
+
+  const changeAutoSuggest = React.useCallback((index: number) => {
```
six7, line 77: Same here, can we move all that into the 1 DownshiftInput.tsx ? It should really only live in 1 place

src/app/components/TypographyInput.tsx
```diff
@@ -1,47 +1,91 @@
...
+    const newShowAutoSuggest = [...showAutoSuggest];
+    newShowAutoSuggest[index] = !newShowAutoSuggest[index];
```
six7, line 79: Also we dont need multiple of these being open at the same time

six7 (changes requested):

src/app/components/InspectorTokenSingle.tsx
swordEdge: DownShiftInput component is used in several components including EditTokenForm. In other components, we expect DownShiftInput to return alias value because it will be resolved using getAliasValue e.g {size.0} => 0.
But in InspectorTokenSingle, we expect name to be returned because we use getTokenValue not getAliasValue. e.g: size.0 => 0.
So we have to deal with alias case in handleDownShiftInputChange function or mappedToken.

src/app/components/InspectorTokenSingle.tsx
swordEdge: This function handles the visibility of the ColorPicker panel which is out of the DownShiftInput. I think it's better to handle it here.

six7: but why do we need a colorpicker in inspect?? inspect should only allow you to choose a different token, not specify a value.

edit: i've checked out the plugin and it seems there was a miscommunication here.

we're not supposed to add a fully functional token input to the Inspector Remap. instead it should just allow you to pick another token only, not specify the value of the token. there's no need for a color picker or the user to enter a token value. theyre supposed to pick another token only.

image

six7: some other things i noticed:

- sometimes it doesnt give me any results. i applied a spacing token for gap (property itemSpacing), but it doesnt show me any results. if i change it to `all` it shows me results.

- i also noticed that the scroll is kind of broken in this one. it should just scroll the tokens, not the whole modal.


https://user-images.githubusercontent.com/4548309/168432820-aa9fec8c-2160-4175-b25e-d5f8fc2a68af.mp4

six7 (changes requested):
i also noticed the scroll is off for box shadows as well. can we please make it so that the scroll occurs in the popover, not the whole modal?

in this video there's also the thing where we're able to open up the popover multiple times. i'd like to solve this, but i feel this requires a refactor of the whole component to something where clicking outside closes the popover. feel free to do this in a followup branch.

https://user-images.githubusercontent.com/4548309/168434148-78356bc3-84cc-482c-8e2d-52aa158ce686.mp4

six7 (approved):

PR #676, opened 2022-05-05
Fix Token Transformer CLI when using sets

### LINKED ISSUE #675: Token Transformer CLI is broken when using sets
**Describe the bug**
Token Transformer has the `sets` option to specify which sets to include in the output.
When using this option in the 0.0.21 release the transformer throws an `TypeError: e.map is not a function` error.

**To Reproduce**
1. `npx token-transformer@0.0.21 input.json output.json global,dark` 
2. See error

**Expected behavior**
Not throwing ;)

**Additional info**
I've done a little bit of investigation, and it seems to me this error potentially always has been there since the CLI (`index.js`) doesn't convert the `string` input arg `sets` to an array, but `utils/transform.ts` => `utils/convertTokensObjectToResolved.ts` expect `sets`/`usedSets` to be a `string[]` array. The only reason it worked up until now is that the `mergeTokenGroups` method in `plugin/tokenHelpers.ts` only used `.length` and `.includes` which - coincidentally - also exists on `string` 😄 
The error now has surfaced as #604 also uses `.map` on the input arg which fails on the `string` type.

jakobe:
Closes #675

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "figma-tokens",
     "version": "1.0.0",
-    "plugin_version": "94",
+    "plugin_version": "95",
     "description": "Figma Tokens",
     "license": "MIT",
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "figma-tokens",
     "version": "1.0.0",
-    "plugin_version": "95",
+    "plugin_version": "96",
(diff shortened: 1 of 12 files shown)


jakobe:
@six7 Could I ask you to review this as soon as you have time, as the Token Transformer CLI is currently broken 😢 
Thanks 🙏🏻 😄

mahmoudmiz (approved):

PR #680, opened 2022-05-06
[License Key / Payment] Updated flow from KeyGen -> LaunchDarkly

roppazvan:
* Update flow so that the entitlements are sent to launch darkly for segmentation
* Added user email for better traceability
* Added flags to branch selector & token themes selector
* License key is now fetched from the server (not stored in client storage as before)

Demo: https://drive.google.com/file/d/1_YUWCjUoG3lomRBTqLMmyxTAdih7NPLb/view?usp=sharing


LiamMartens (changes requested):
Can we remove package-lock.json since we use yarn usually

six7:
@roppazvan there's a failing cypress test. the other failing test is fixed if you merge in the latest `next` branch.

six7 (approved):

PR #688, opened 2022-05-09
Minor optimizations

six7:
Includes some optimizations ahead of the upcoming release.


src/app/components/Input.tsx
```diff
@@ -26,6 +26,7 @@ type Props = React.InputHTMLAttributes<HTMLInputElement> & {
   isMasked?: boolean;
   onChange?: React.ChangeEventHandler<HTMLInputElement>;
   disabled?: boolean;
+  size?: 'small' | 'large';
```
LiamMartens, line 29: This causes a minor typing issue because input already has a `size` prop. You can update the `React.InputHTMLAttributes` line with `Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'>`. Can do later if you need to.

src/app/components/StorageProviderSelector.tsx
```diff
@@ -27,9 +31,17 @@ const StyledButton = styled('button', {
   },
 });
 
+type Props = {
```
LiamMartens, line 34: Good to see prop typings :smile:

LiamMartens (approved):

PR #697, opened 2022-05-11
fix typography and boxshadow tokens on alias level

six7:


mikekamminga:
Issues:

- Shadows: deep alias works within the same token set, but fails when referencing another token set, it just show the `{value.like.this}`. For typography it works fine.
- Not sure if we intended this, but when I remove shadow tokens, shadow does not get removed from the layer (video in slack)

PR #704, opened 2022-05-13
fix: Updated the JSONBin token schema validation

LiamMartens:


LiamMartens:
@six7 let me know if Im missing any properties from the tokens that we send upstream so I can add them to the singleTokenSchema

six7 (approved):

six7:
> @six7 let me know if Im missing any properties from the tokens that we send upstream so I can add them to the singleTokenSchema

i think it's fine, those should be the built-in ones. theoretically users could set any key/value combination they'd like in their tokens though?

PR #709, opened 2022-05-13
fix: Wait for license verification or no license status before identi…

LiamMartens:
…fying the user


LiamMartens:
Otherwise we try to pull tokens before receiving the correct flags from the license key.

six7 (approved):

PR #715, opened 2022-05-16
Add ability to use units in math expressions

nemzes:
This allows expressions with units to be interpreted, which in turn solves the issue of not being able to use tokens with units in aliases. For instance, this will now work:
```
t1 = 10px
t2 = {t1} * 2
```
All the hard work is done by the `post-calc-at-parser` package by @ota-meshi, so most of the credit should go to them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "octokit-commit-multiple-files": "^3.2.1",
     "polished": "^4.1.3",
     "postcss": "^8.2.13",
+    "postcss-calc-ast-parser": "^0.1.4",
     "react": "17.0.0",
     "react-colorful": "^5.5.1",
     "react-devtools": "^4.24.3",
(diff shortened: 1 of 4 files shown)


six7 (approved):

six7:
> Ugh, looks like I broke a lot of stuff. I think there were some failing tests before and then only looked at the test file I created 🤦
> 
> Will take a look.
> 
> One thing I changed is enforcing the return value of the token to always be a string (since it can contain a unit), while before it was being coerced into a number if it passed the parsing step (it would stay a string if it didn't). Would this be a problem? If not I'll amend the tests.

i think we should preserve that if a token value was a number it should stay a number, as the w3c draft allows tokens to be of type string and number among others. do you think this would be possible?

nemzes:
Yes, it should be. On it.

six7:
Amazing, thank you! Will be part of the next major release 🙏

PR #719, opened 2022-05-17
Ft 136 remove icons from components and merge to icons

MightyDevKnight:


-L
```diff
@@ -0,0 +1 @@
+Redirecting to https://classic.yarnpkg.com/install.sh
```
six7, line 1: Is this required?

src/app/components/Icon.tsx
```diff
@@ -1,25 +1,25 @@
 import React from 'react';
```
six7, line 1: Is this file still required? I think we can remove it and if there's imports left change those to svg?

src/app/components/ThemeSelector/ThemeSelector.tsx
```diff
@@ -15,7 +15,7 @@ import {
 import { Flex } from '../Flex';
 import Text from '../Text';
 import { styled } from '@/stitches.config';
-import { IconToggleableDisclosure } from '../icons/IconToggleableDisclosure';
+import { IconToggleableDisclosure } from '@/icons/IconToggleableDisclosure';
```
six7, line 18: Can we change this one to .svg as well?

six7 (changes requested):

-L
MightyDevKnight: removed

src/app/components/ThemeSelector/ThemeSelector.tsx
MightyDevKnight: This icon component has props. 
 {activeTokenSet}
<IconToggleableDisclosure open={tokenSetsVisible} />

src/app/components/Icon.tsx
MightyDevKnight: removed

src/app/components/ThemeSelector/ThemeSelector.tsx
six7: hm, ok. for now it's fine but i'll create a followup ticket to solve this better. icons should really only be .svg and if there's a transformation happening it should be a wrapping component using that icon

six7 (approved):

PR #743, opened 2022-05-24
Ft 173 bug starting plugin in new file leads to endless loading screen

swordEdge:


six7 (commented):
@swordEdge I'm not sure I understand why the snapshots were removed? Nevertheless I think we can close this PR as you fixed it in another branch (the ActionButton branch)?

LiamMartens (approved):

PR #762, opened 2022-05-26
Support GHE URL structures for PR popup

zslabs:
### Description

When using a custom base URL (GitHub Enterprise) as mentioned in https://github.com/six7/figma-tokens/issues/725 this new URL is not taken into account for the PR popup that appears when making changes. This PR adjusts the params to be object-based (I figured this way would break up the properties a bit cleaner) to support this and also updates the format of GitLab for consistency.

I started with `new URL()` but the string replace seemed like an easier way to start; as I'm unaware of all the different path segments a GHE URL can have.

There's also a few `// eslint-disable-next-line` additions as Husky was running up against these when I attempted to push and a https://eslint.org/docs/rules/no-case-declarations fix.

I didn't have a chance to fire this up in Figma, so another pair of 👀 would be appreciated. Let me know what you think!


six7:
This is great! Do you mind basing changes on the `next` branch? We're keeping main the current release so we can ship hotfixes. Next is the branch we'll ship as the next major release and we had quite a few refactorings.

zslabs:
Sure thing! Let me know if the most recent commit looks better. Thanks!

src/app/components/PushDialog.tsx
```diff
@@ -56,9 +59,8 @@ function ConfirmDialog() {
     case 'initial': {
       return (
         <Modal large isOpen close={onCancel}>
-          <form
-            onSubmit={() => onConfirm(commitMessage, branch)}
-          >
+          {/* eslint-disable-next-line react/jsx-no-bind */}
+          <form onSubmit={() => onConfirm(commitMessage, branch)}>
```
six7, line 63: Can you change this to a useCallback and then remove the eslint ignore?

src/app/components/PushDialog.tsx
```diff
@@ -70,6 +72,7 @@ function ConfirmDialog() {
                   full
                   label="Commit message"
                   value={commitMessage}
+                  // eslint-disable-next-line react/jsx-no-bind
```
six7, line 75: Same here

src/app/components/PushDialog.tsx
```diff
@@ -79,6 +82,7 @@ function ConfirmDialog() {
                   full
                   label="Branch"
                   value={branch}
+                  // eslint-disable-next-line react/jsx-no-bind
```
six7, line 85: Same

six7 (changes requested):
Just 3 comments about the eslint ignores. I know, it wasnt your code, but going forward we'd want to remove any eslint warnings by fixing them instead of ignoring them. 🙏

zslabs:
Sure! New commit just pushed up.

six7:
Currently still running into an error trying to build this (the `next` branch is type-safe and will fail on type errors), this line seems to offend as `id` doesn't exist for `LOCAL`: https://github.com/six7/figma-tokens/blob/62290a2828fc5f44add5bbde558f304706078c64/src/app/components/PushDialog.tsx#L22

PR #764, opened 2022-05-27
Ft 176 bug plugin doesnt apply git hub credentials

swordEdge:


src/types/tokens/TokensStore.ts
```diff
@@ -12,5 +12,7 @@ export type TokenStore = {
   checkForChanges?: boolean | null;
   activeTheme: string | null;
   themes: ThemeObjectsList;
-  storageType?: StorageType
+  storageType?: StorageType;
+  userId?:string | null;
```
six7, line 16: could we store these 2 some other place as their own type? they don't feel connected to TokenStore (things we store on the document)

src/plugin/asyncMessageHandlers/initiate.ts
```diff
@@ -41,7 +41,7 @@ export const initiate: AsyncMessageChannelHandlers[AsyncMessageTypes.INITIATE] =
     const oldTokens = await getTokenData();
     if (oldTokens) {
       notifyTokenValues({
-        ...oldTokens, activeTheme, usedTokenSet, storageType,
+        ...oldTokens, activeTheme, usedTokenSet, storageType, userId: currentUser?.id, licenseKey,
```
six7, line 44: can we separate these two value groups rather like this

```suggestion
        values: { ...oldTokens, activeTheme, usedTokenSet, storageType },
        userData: { userId: currentUser?.id, licenseKey }
```

and then change notifyTokenValues to accept these like that? and add useData as its own type? that way we're separating this we store on the document with user data we submit

six7 (changes requested):
For a multi file sync where the folder is nested like this: `packages/ds/figma/tokens`, this is what I get:

image

For a multi file sync that is using `tokens` as a filePath this is what I get (expected):
image

six7:
Nice!

PR #765, opened 2022-05-27
fix: don't overwrite linked style with raw value

### LINKED ISSUE #758: Nodes with library styles gets their selected style overwritten with hex value
**Describe the bug**
When a node in the document is attached to both a Figma Tokens plugin token (e.g. `color.core.gray.50`) and gets it's color from a library style (e.g. `color/core/gray/50`) the Figma color style is replaced with the hex value from the token.

This happens both when explicitly assigning a token that corresponds to a Figma color style from a library but also on overall updates from the plugin (e.g. when toggling token sets in the plugin).

The bug also surfaces when copying an element (e.g. a component) from the library master file where the color styles are defined and that element has a color token assigned to it. When pasting that element (e.g. as an instance) into another file that uses the color styles from the library the color styles are overwritten with hex values on updates from the plugin (e.g. when toggling token sets in the plugin).

_*Just for the record: Text and effect styles are also affected by this issue, I'm just using colors as an example here_

**To Reproduce**
1. Create a new file in Figma
2. Add a new token set (e.g. `Global`) in the Figma Tokens plugin
3. Add a color token (e.g. `color.core.gray.50`) in the plugin and create Figma color styles from the tokens (`Styles > Create styles`)
4. Publish the file as a library
5. Create a new file and enable the published library from step #&#x2060;4
6.  Create a rectangle and assign the color token created in step #&#x2060;3 (e.g. `color.core.gray.50`) through the plugin corresponding to the Figma library color style (`color/core/gray/50`)
7. See that the hex value is applied (e.g. `F6F6F6`) and not the Figma library color style (`color/core/gray/50`)
8. Assign the Figma library color style (`color/core/gray/50`) through Figma's `Design` panel.
9. Assign the color token created in step #&#x2060;3 (e.g. `color.core.gray.50`) through the plugin
10. See that the hex value is applied (e.g. `F6F6F6`) and not the Figma library color style (`color/core/gray/50`)

**Expected behavior**
I would expect the plugin to NOT overwrite my applied Figma color styles.  
Either because it can map the color token to the - non-local - color styles from the library OR  
when it cannot because it sees a color style is applied and doesn't overwrite it.

I guess the main reason for this bug is because [Figma only exposes the local paint styles](https://www.figma.com/plugin-docs/api/figma/#getlocalpaintstyles) and [does not (yet at least) expose a way to get library styles ](https://forum.figma.com/t/getlocalpaintstyles-but-for-libraries/1280)

@six7 As far as I can see the offending lines are in [src/plugin/setValuesOnNode.ts](https://github.com/six7/figma-tokens/blob/main/src/plugin/setValuesOnNode.ts#L82-L88).
If you agree on the expected behavior, maybe a simply check on existing `node.fillStyleId` before assigning would do the trick?
We could even extend it further with a check for whether the style exists (`figma.getStyleById(node.fillStyleId)`) and add that style to the `figmaStyleMaps` used to check for matching styles?

I'll be happy to do a fix and PR if you see fit 🤞🏻 😊

jakobe:
Fixes #758


six7 (commented):
One thing to do here IMO is extract all those functions out into their own files, possibly even the code introduced in the effects/paints/type functions itself - this will make it easier once the other changes land (so ideally it will then just be another condition:
a) matching local style
b) theme has a style id that we can use
c) backup (your pr)
d) raw hex value

in the future i'd imagine we'll also allow users to set a styleId on each of their tokens, which would come between `b` and `c` then.

jakobe:
Hi @six7 ,

I've cleaned up the code and moved all new functions to separate files as suggested - as well as removed some code duplication.

I've tried to follow the current structure, i.e. added to/updated `figmaTransforms`, `figmaUtils` and `utils` folders as appropriate. Let me know if you think anything needs to move somewhere else.

I also took the liberty to refactor the `isPaintEqual` method as it[ actually doesn't work at the moment when comparing Figma's rgba values to web rgba](https://forum.figma.com/t/figmas-rgb-color-values-seems-to-have-a-rounding-error/23540).

Finally I've fixed the failing tests (mocks missing `get|setSharedPluginData` methods).

The refactoring to own files are done in this commit https://github.com/six7/figma-tokens/pull/765/commits/62d25347468dff9c987fecfacd8ae5376c918f2b and this https://github.com/six7/figma-tokens/pull/765/commits/a4ed028ebb23267a9467a1cafd63db5bfc179cac

I consider this PR to be pretty done now, but I'll leave it as a draft PR until you ping me about [the theme style ids PR](https://github.com/six7/figma-tokens/pull/778).

Cheers,
Jakob 😊

six7:
Hey @jakobe - we finally merged the theme / style id connections changes into `next`. Do you mind merging `next` into your branch and giving it a go if it still works? Thank you 🎉

jakobe:
> Hey @jakobe - we finally merged the theme / style id connections changes into `next`. Do you mind merging `next` into your branch and giving it a go if it still works? Thank you 🎉

Hey @six7 great to hear 🎉 
Sorry for the late reply, just came back today after a long and needed vacation 😎 
I'll have a look today or tomorrow and get back to you 🤞🏻 
Cheers - J

jakobe:
> Hey @jakobe - we finally merged the theme / style id connections changes into `next`. Do you mind merging `next` into your branch and giving it a go if it still works? Thank you 🎉

Hey @six7 merged the latest `next` branch into my fork and solved all merge conflicts 👍🏻 
It seems to still work in the testing I've done myself - I've also asked my designer to do some additional testing on some of his works as well.

Would you or @LiamMartens have time to have a look at the PR and let me know if you see any improvements I should make? Thx, Jakes 😄

six7:
@jakobe could you fix the merge errors that seem to have occurred? We renamed a type which seems to cause that.

@LiamMartens could you look at this PR from a high level? It's solving the case where a user does not have use the NonLocal styles feature but wants to retain the style connection on a layer, right now we always change to hex even if there was no change in the token value, however it would be ideal if we'd just keep that style set on the layer if possible, which this PR does.

jakobe:
> @jakobe could you fix the merge errors that seem to have occurred? We renamed a type which seems to cause that.
> 
> @LiamMartens could you look at this PR from a high level? It's solving the case where a user does not have use the NonLocal styles feature but wants to retain the style connection on a layer, right now we always change to hex even if there was no change in the token value, however it would be ideal if we'd just keep that style set on the layer if possible, which this PR does.

Oh thx, didn't catch that one. Seems you fixed a typo on `next` branch after I merged it - it's fixed now 👍🏻 

@LiamMartens Also note, that this solves going the other way as well, i.e. overriding an applied non-local style with raw token values (i.e. selecting a dark theme in the plugin) and when going back to tokens that matches (e.g. default/light theme) this will re-connect/apply the non-local style when it matches the selected tokens in the plugin.

src/plugin/figmaUtils/styleMatchers/paintStyleMatchesColorToken.ts
```diff
@@ -0,0 +1,12 @@
+import { isPaintEqual } from '@/utils/isPaintEqual';
+import { convertToFigmaColor } from '../../figmaTransforms/colors';
+
+export function paintStyleMatchesColorToken(paintStyle: PaintStyle | undefined, colorToken: string) {
+  const stylePaint = paintStyle?.paints[0] ?? null;
```
LiamMartens, line 5: @jakobe @six7 do we want to add support for the LINEAR_GRADIENT fills or?

src/plugin/figmaUtils/styleMatchers/textStyleMatchesTypographyToken.ts
```diff
@@ -0,0 +1,79 @@
...
+  const tokenTextCase = transformValue(String(textCase), 'textCase');
+  if (tokenTextCase !== textStyle.textCase) {
+    return false;
+  }
+  // This will default to `NONE` if textDecoration token is not set:
+  const tokenTextDecoration = transformValue(String(textDecoration), 'textDecoration');
+  if (tokenTextDecoration !== textStyle.textDecoration) {
+    return false;
+  }
+  // TODO: Should description also match? 🤷
```
LiamMartens, line 74: @six7 I guess this is  a question to you

src/plugin/setValuesOnNode.ts
```diff
@@ -159,12 +184,29 @@ export default async function setValuesOnNode(
...
             data.fill,
             pathname,
             activeThemeObject?.$figmaStyleReferences ?? {},
             figmaStyleMaps.paintStyles,
           );
+
+          if (!matchingStyleId) {
+            // Local style not found - look for matching non-local style:
+            const styleIdBackupKey = 'fillStyleId_original';
+            const nonLocalStyle = getNonLocalStyle(node, styleIdBackupKey, 'fills');
```
LiamMartens, line 197: I'm a little confused as to what the whole style ID backup does and why it's necessary.

LiamMartens (changes requested):
@jakobe looks good overall - added a few minor questions.
Aside from that it would be great to add test coverage for the new utilities like the style matchers etc...

src/plugin/setValuesOnNode.ts
jakobe: > I'm a little confused as to what the whole style ID backup does and why it's necessary.

Hi @LiamMartens, the reason for this is, that when we set raw values on an element we loose the style (id) that it was linked to. For local styles this is not an issue, as we can get it back using the path of the token and `figma.getLocalPaintStyles|getLocalTextStyles|getLocalEffectStyles()`. 
There currently isn't any way to do this for non-local style, that's why we either backup the id before attaching raw values or re-attach the non-local style from the backup when the values matches the token values.

src/plugin/figmaUtils/styleMatchers/paintStyleMatchesColorToken.ts
jakobe: > @jakobe @six7 do we want to add support for the LINEAR_GRADIENT fills or?

Yeah, good point, I guess so, right @six7?
I can give that a whack in the coming days 👍🏻

jakobe:
> @jakobe looks good overall - added a few minor questions. Aside from that it would be great to add test coverage for the new utilities like the style matchers etc...

Good point, I'll have a look at adding tests 👍🏻

src/plugin/figmaUtils/styleMatchers/paintStyleMatchesColorToken.ts
six7: Yeah that would be good to have. 👍

src/plugin/figmaUtils/styleMatchers/textStyleMatchesTypographyToken.ts
six7: Let's not match in description, I'd assume this would cause user frustration as its not easy to spot

jakobe: Fixed 👍🏻

src/plugin/figmaUtils/styleMatchers/paintStyleMatchesColorToken.ts
jakobe: Added 👍🏻 
https://github.com/six7/figma-tokens/pull/765/commits/246c7bdf7e97346d1ec1f44de7f099a8e6f4ed13

jakobe:
> @jakobe looks good overall - added a few minor questions. Aside from that it would be great to add test coverage for the new utilities like the style matchers etc...

Hey @LiamMartens , added a bunch of tests to all the new style matchers util classes:
https://github.com/six7/figma-tokens/pull/765#commits-pushed-246c7bd

Also added the linear gradient color matcher, removed the description from text style matcher and added some [notes to your question regarding style id backup](https://github.com/six7/figma-tokens/pull/765#discussion_r950876791).

Will you give it a re-review? Thx 🙏🏻 😊

jakobe:
@LiamMartens & @six7 any news on this one? 🙏🏻 😊

src/plugin/setValuesOnNode.ts
six7: What we'd need to make sure is that this behavior doesn't get in the way of the non-local styles feature of Themes, which should always win. Is that still the case here?

jakobe: > What we'd need to make sure is that this behavior doesn't get in the way of the non-local styles feature of Themes, which should always win. Is that still the case here?

I understand your consideration here @six7 , and while I haven't tested that part as such (I don't have the PRO version yet 😊 ), code wise this new part of the code that looks for non-local styles _only_ gets triggered if the code on line 86-91 doesn't match a style (including checking the `activeThemeObject?.$figmaStyleReferences`) 🤞🏻

jakobe:
@six7 @LiamMartens [I did a little tweak yesterday](https://github.com/six7/figma-tokens/pull/765/commits/178fa62e1d7a4aa7ff8aa01c7745c883cc76baa3) to the code that makes the backup of the styleid before it overwrites the non-local style with raw values.

I'll try to explain:

It now checks if the path of the selected token matches the name of the non-local style, before it backs up the style id and overwrites it with raw values (e.g. when selecting a dark mode tab):
`{color.action-cta.default} => color/action-cta/default` and
`{color.action-cta.default} => action-cta/default` (when ignoring first part of token name for styles)

Otherwise it would keep "stale" styleids when assigning a different token, because the new token wouldn't match the linked non-local style.

**TL;DR:** We should only backup the linked non-local style when the value doesn't match but the token path matches the style name, because that signals it's the same token as the linked style but with a different value (e.g. dark mode).

Hope it makes sense 😊 🤞🏻

six7:
> @six7 I think this looks good from a code perspective - not sure what you want to do with the coverage diff situation. Did you check this out functionally?

I'll give this a functional go later this week 👍 

Regarding test coverage checks: Unfortunately our integration doesn't seem to be working for external collaborators :/ I'll check manually if coverage for added files increased or not

six7 (approved):

six7:
Merging this in, checked functionality 🙏 

We'll be releasing a beta for version 120 on Saturday which we're aiming to test for a week before releasing.

jakobe:
> Merging this in, checked functionality 🙏
> 
> We'll be releasing a beta for version 120 on Saturday which we're aiming to test for a week before releasing.

Great to hear, thanks a lot @six7 🙏🏻 

Sounds good with v120 and the timeline 👍🏻 😊

PR #772, opened 2022-05-29
fix branches selector

six7:
fixes old branches showing up in branch selector, basically just moved this from the release-102 branch to the next branch


swordEdge (commented):
It works well.

PR #782, opened 2022-05-31
fix bug with brokenlink

MightyDevKnight:


src/app/components/InspectorTokenGroup.tsx
```diff
@@ -20,7 +20,10 @@ export default function InspectorTokenGroup({ group, resolvedTokens }: { group:
       key={`${groupKey}`}
     >
       <Heading size="small">{groupKey}</Heading>
-      {groupValue.map((uniqueToken) => <InspectorTokenSingle key={`${uniqueToken.category}-${uniqueToken.value}`} token={uniqueToken} resolvedTokens={resolvedTokens} />)}
+      {groupValue.map((uniqueToken) => {
+        const isInResolvedTokens = resolvedTokens.find((token) => token.name === uniqueToken.value);
```
six7, line 24: can we move this to `InspectorTokenSingle` and handle it there instead of handling it outside?

Should probably also be in a useMemo or useCallback (cc @LiamMartens)

MightyDevKnight: Done

six7 (approved):

PR #789, opened 2022-06-01
Fix/ado

six7:


LiamMartens:
@six7 I fixed the issues with the ADO sync

PR #795, opened 2022-06-02
Ft 216 bugfix remap composition token doesnt work

swordEdge:


LiamMartens:
@swordEdge would you be able to explain why this didn't work and what the fix does?

swordEdge:
> would you be able to explain why this didn't work and what the fix does?

To remap composition token, first we need to remove all tokens in composition token which is applied to the node unlike regular tokens.
It is because in the case of regular tokens, new values will be applied to the node but composition token is made up with different kind of tokens so we need to remove all tokens in old composition token at first.
And then we can apply new tokens in new composition token.

e.g first: { borderRadius: 8px, opacity: 60% }, second: { opacity: 90% }
To remap first composition token to second composition token, we need to remove borderRadius and opacity values.
Without this process, borderRadius will be kept to be applied to the node even after remapping tokens.

LiamMartens (approved):

PR #803, opened 2022-06-02
Ft 222 bug composition tokens on hover showing box shadow as object object

swordEdge:


six7 (changes requested):
Doesn't display resolved values properly right now. Probably because some of those token types are of type `number` which isnt being handled right now

image

src/app/components/TokenTooltip/SingleCompositionValueDisplay.tsx
```diff
@@ -1,16 +1,52 @@
...
+});
+
+const StyledItemName = styled('div', {
+  flexShrink: 0,
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+});
+
+export const SingleCompositionValueDisplay: React.FC<Props> = ({ property, value }) => {
+  const resolveValue = React.useCallback((value: SingleToken['value'] | number) => {
```
LiamMartens, line 25: I would update this to use `useMemo` instead of `useCallback`. There's no need to re-calculate this value each time so we can memo it.

LiamMartens (changes requested):

PR #808, opened 2022-06-03
Ft 148 create missing tests for src utils alias

MightyDevKnight:


six7 (changes requested):
can you separate this PR from the tooltip one? this one seems to include both but some parts are commented out.

also, i noticed we currently have 2 files for `getAliasValue.test.ts` - can you merge the two into one file in this pr? the one in the `__tests__` folder contains one big `allTokens` object which would fit the scenarios well that you did here - iterating over each of them and making sure that the `input` (you called it `rawValue`) matches the `value`.

src/app/components/TokenTooltip/TokenTooltipContent.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+    rawValue: 'normal',
+  },
+  {
+    name: 'font-style.alias',
+    type: TokenTypes.OTHER,
+    value: 'normal',
+    rawValue: '{font-style.normal}',
+  },
+];
+// describe('TokenTooltip value', () => {
```
six7, line 181: can be removed i guess

src/app/components/TokenTooltip/TokenTooltipContent.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+  it('size alias test', () => {
+    const input: SingleToken = {
+      name: 'size.alias',
+      type: TokenTypes.SIZING,
+      value: '{size.6}',
+      rawValue: '2',
+    };
+    const { getByText } = render(<TokenTooltipContent token={input} />);
+    expect(getByText('{size.6}')).toBeInTheDocument();
+    expect(getAliasValue(input, tokens)).toEqual(2);
```
six7, line 229: can we change this line to not test `getAliasValue` but instead test the tooltip contents to contain `2` ?

i think we might need to "find" the node containing it (tooltip) first.

same for all following tests. this test is not about getAliasValue, its about showing the results. getAliasValue is already tested.

src/utils/alias/__tests__/checkifContainsAlias.test.ts
```diff
@@ -0,0 +1,18 @@
+import { checkIfContainsAlias } from '../checkIfContainsAlias';
+
+const tokenAliasValue = [
+  '{size.6}',
+  '$color.#450987',
+];
+const tokenWithOutAlias = '2';
+
+describe('contain alias test', () => {
+  it('ifContainAlias', () => {
```
six7, line 10:
```suggestion
  it('returns true when token value is an alias', () => {
```

src/utils/alias/__tests__/checkifContainsAlias.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    tokenAliasValue.forEach((tokenValue) => {
+      expect(checkIfContainsAlias(tokenValue)).toBe(true);
+    });
+  });
+  it('ifContain no Alias', () => {
```
six7, line 15:
```suggestion
  it('returns false when no alias was used', () => {
```

src/utils/alias/__tests__/getAliasValue.test.ts
```diff
@@ -1,109 +1,552 @@
...
+    type: TokenTypes.OTHER,
+    value: 'normal',
+  },
+];
+describe('getAliasValue', () => {
+  inputTokens.forEach((token, idx) => {
     it(`alias ${token.name}`, () => {
       // @TODO check this test typing
-      expect(getAliasValue({ value: token.input } as SingleToken, allTokens as unknown as SingleToken[])).toEqual(token.value);
+      expect(getAliasValue(token as SingleToken, inputTokens as unknown as SingleToken[])).toEqual(expectedTokens[idx].value);
```
six7, line 549: This will be a bit tough to maintain, as in the future if we wanted to add something we'd need to make sure the `idx` matches of expectedTokens to the inputTokens.

The previous way included both expected outcome but also the input in the same object, which makes it easier to add a new item. Can we change that to be 1 array where we just have a value for expected outcome and one for the value?

six7 (changes requested):

six7 (approved):

PR #827, opened 2022-06-07
Ft 263 bug multifile reopening the plugin in a multifile setup doesnt apply sync settings correctly

swordEdge:


src/app/components/Initiator.tsx
```diff
@@ -51,6 +54,35 @@ export function Initiator() {
     })
   ), []);
 
+  const getFeatureFlags = useCallback(async (userData: UserData) => {
```
six7, line 57: can we move this to its own file/function and write tests for that file then?

swordEdge: I moved it to its own file. And I created a ticket on Jira for test coverage.

LiamMartens (approved):

PR #834, opened 2022-06-08
Ft 235 downshift input autocomplete

MightyDevKnight:


src/stories/DownshiftInput.stories.mdx
```diff
@@ -0,0 +1,203 @@
...
+  {
+    value: '#f8fafc',
+    type: TokenTypes.COLOR,
+    name: 'color.slate.50',
+    rawValue: '#f8fafc'
+  },
+  {
+    value: '#f9fafb',
+    type: TokenTypes.COLOR,
+    name: 'color.gray.50',
```
six7, line 55: can you change these to be e.g. a red color with #ff0000 just so we can see the color swatches working? right now it's 3 very bright gray values

six7 (approved):

PR #843, opened 2022-06-08
Release/106

LiamMartens:
Fixes FT-267


six7:
I'll close this one as I'll only release the fix for the cache and leave the rest in the `next` branch.

PR #876, opened 2022-06-13
Ft 143 create tests for remote tokens

swordEdge:


LiamMartens (approved):

src/app/store/providers/github/github.tsx
```diff
@@ -88,7 +86,7 @@ export function useGitHub() {
           themes,
           usedTokenSet,
         });
-        pushDialog('success');
+        pushDialog('successfully pushed');
```
six7, line 89: i think we should keep this as 'success'. this isn't a text, it's a state (maybe we should convert it and all other states of pushdialog to an enum?)

see

src/app/components/PushDialog.tsx#L130

src/app/store/providers/ado/ado.tsx
```diff
@@ -43,8 +43,7 @@ export const useADO = () => {
       text: 'Pull from Ado?',
       description: 'Your repo already contains tokens, do you want to pull these now?',
     });
-    if (confirmResult === false) return false;
-    return confirmResult.result;
+    return confirmResult
```
six7, line 46: while testing this PR i noticed that we `pull` even though the user pressed `Cancel`. This is the same on the live plugin. Can we fix this here?

six7 (changes requested):

PR #886, opened 2022-06-15
Ft 280 url sync provider gets duplicated on edit

MightyDevKnight:


src/app/components/Initiator.tsx
```diff
@@ -186,7 +194,7 @@ export function Initiator() {
             break;
           }
           case MessageFromPluginTypes.API_PROVIDERS: {
-            dispatch.uiState.setAPIProviders(pluginMessage.providers);
+            dispatch.uiState.setAPIProviders(removeDuplicatedProviders(pluginMessage.providers));
```
six7, line 197: i dont think we need to remove duplicated providers. we should rather not duplicate them in the first place.

MightyDevKnight: You assigned this to @Hiroshi, right?

six7: Oh sorry, meant you. Was late for me 😅

src/app/components/StorageItemForm/URLForm.tsx
```diff
@@ -40,6 +47,22 @@ export default function URLForm({
     }
   }, [values, onSubmit]);
 
+  useEffect(() => {
+    if (!isNew) {
```
LiamMartens, line 51: This logic seems a bit weird? What is the intention here? You are removing the old provider if the one you are editing is not new? Should we not just fix the issue where the actual provider is not updated? This feels like a monkey patch.

LiamMartens (changes requested):

six7:
Can you merge `next` in?

PR #891, opened 2022-06-16
Ft 104 multi file sync handle set deletions

swordEdge:


src/storage/ADOTokenStorage.ts
```diff
@@ -325,7 +325,7 @@ export class ADOTokenStorage extends GitTokenStorage {
     return response;
   }
 
-  public async writeChangeset(changeset: Record<string, string>, message: string, branch: string, shouldCreateBranch: boolean = false): Promise<boolean> {
+  public async writeChangeset(changeset: Record<string, string>, message: string, branch: string, modifiedTokenSet: string[], shouldCreateBranch: boolean = false): Promise<boolean> {
```
swordEdge, line 328: modifiedTokenSet will be used if multifile feature is implemented on ado sync

swordEdge:
In multifile sync, We need to remove files for renaming and deleting tokenSet.
I introduced modifiedTokenSet to let the Git know which files has been changed.
Without it, we need to remove all files and create again, which takes longer time.

PR #900, opened 2022-06-16
feat: coverage for adotokenstorage

LiamMartens:


src/storage/ADOTokenStorage.ts
```diff
@@ -186,6 +189,8 @@ export class ADOTokenStorage extends GitTokenStorage {
 
   private async getItem(path: string = this.path): Promise<any> {
     try {
+      // @REAMDE setting includeContent to true
```
six7, line 192:
```suggestion
      // @README setting includeContent to true
```

src/storage/ADOTokenStorage.ts
```diff
@@ -280,13 +285,13 @@ export class ADOTokenStorage extends GitTokenStorage {
         return [
           {
             type: 'themes',
-            path: `${this.path}/$themes.json`,
+            path: this.path,
```
six7, line 288: So does this store $themes in the same file as the sets then? How did this even work before?

six7 (changes requested):
One minor typo and one question for my understanding

src/storage/ADOTokenStorage.ts
LiamMartens: It does - it has to because of how the multifile was set-up from the beginning. It is based on the filename. It was not possible to change the file structure anymore since it had alreayd been set in a flat file structure.

six7 (approved):

PR #904, opened 2022-06-17
Ft 220 gh text alias preview text size line height are swapped

swordEdge:


LiamMartens (approved):

src/app/components/ResolvedTypograhpyValueDisplay.tsx
```diff
@@ -28,12 +37,13 @@ export const ResolvedTypograhpyValueDisplay: React.FC<Props> = ({ value }) => {
     }}
     >
       <Box css={{ display: 'grid', marginRight: '$6' }}>
-        {properties.map((property) => (
-          <StyledPropertyItem key={seed(property)}>{property}</StyledPropertyItem>
+        {Object.values(properties).map((value) => (
```
six7, line 40: `value` is already defined in the upper scope

six7 (changes requested):
One thing about variable name

PR #906, opened 2022-06-17
fix issue with dulicating token-set-creates-a-new-blank-token-set

MightyDevKnight:


src/app/components/TokenSetSelector.tsx
```diff
@@ -32,6 +32,7 @@ const StyledButton = styled('button', {
     boxShadow: 'none',
     backgroundColor: '$bgSubtle',
   },
+  length: 0,
```
six7, line 35: what is this one for?

src/app/store/models/tokenState.tsx
```diff
@@ -137,7 +137,14 @@ export const tokenState = createModel<RootModel>()({
       }
 
       const newName = `${name}_Copy`;
-      return updateTokenSetsInState(state, null, [newName]);
```
LiamMartens, line 140: Could you instead update the `updateTokenSetsInState` logic. It is necessary because it also updates the set names in the themes etc.. as we need to make sure they remain in sync.

src/app/components/TokenSetSelector.tsx
MightyDevKnight: There was an error with `eslint` and it fixes automatically to add length.
Removed.

src/app/store/models/tokenState.tsx
```diff
@@ -137,7 +137,15 @@ export const tokenState = createModel<RootModel>()({
...
+      const newState = {
+        ...state,
+        tokens: {
+          ...state.tokens,
+          [newName]: [...state.tokens[name]],
+        },
+      };
+
+      return updateTokenSetsInState(newState, null);
```
LiamMartens, line 148: Could you just remove your added logic and replace it with `return updateTokenSetsInState(state, null, [newName, state.tokens[name]]);`. This will have the same behavior and then we don't need to duplicate the logic.

LiamMartens (changes requested):
Just 1 comment

src/app/store/models/tokenState.tsx
MightyDevKnight: Done.

LiamMartens (approved):

PR #911, opened 2022-06-18
Ft 251 tests add cypress test for creating themes based on token sets

MightyDevKnight:


cypress/integration/manage-themes.spec.js
```diff
@@ -0,0 +1,96 @@
...
+      .get('[data-cy="tokensettheme-item--dropdown-trigger--token-enabled-set"]').click()
+      .get('[data-cy="tokensettheme-item--dropdown-content--enabled"]').click()
+      .get('[data-cy="tokensettheme-item--dropdown-trigger--token-disabled-set"]').click()
+      .get('[data-cy="tokensettheme-item--dropdown-content--disabled"]').click()
+      .get('[data-cy="button-manage-themes-modal-save-theme"]').click();
+    cy.get('[data-cy="close-button"]').click();
+    createTokenSet({ name: 'token-extra' });
+
+    cy.get('[data-cy="themeselector-dropdown"]').click()
+      .get('.c-fqGJgX').click();
```
six7, line 91: can we set a data-cy selector on this? using a classname like this might break

cypress/integration/manage-themes.spec.js
```diff
@@ -0,0 +1,96 @@
...
+        }],
+      },
+    });
+    cy.receiveStorageTypeLocal();
+    
+    createTokenSet({ name: 'token-source' });
+    createTokenSet({ name: 'token-enabled' });
+    createTokenSet({ name: 'token-disabled' });
+
+    // receiveRemoteComponents();
```
six7, line 51: can we remove this if it's not needed?

LiamMartens (approved):

PR #915, opened 2022-06-20
Ft 104 multi file sync handle set deletions

swordEdge:


src/storage/schemas/multiFileSchema.ts
```diff
@@ -1,8 +1,8 @@
 import z from 'zod';
 import { TokenSetStatus } from '@/constants/TokenSetStatus';
-import { singleTokenSchema } from './singleTokenSchema';
+import { tokensMapSchema } from './tokensMapSchema';
 
-export const multiFileSchema = z.record(singleTokenSchema).or(z.array(z.object({
+export const multiFileSchema = z.record(tokensMapSchema).or(z.array(z.object({
```
LiamMartens, line 5:
```suggestion
export const multiFileSchema = tokensMapSchema.or(z.array(z.object({
```

LiamMartens (changes requested):

LiamMartens:
@swordEdge are we able to update the test coverage to pass?

swordEdge:
> @swordEdge are we able to update the test coverage to pass?

Modified test coverage for multi-file sync. I think it's still failing because of another unit test.

LiamMartens:
@swordEdge it's stil failing because `src/storage/GithubTokenStorage.ts` has a branch coverage diff of `-1`, maybe you can check which lines can be covered additionally?

PR #923, opened 2022-06-23
FT-160 save token sort order in metadata

LiamMartens:


six7 (changes requested):
For a metadata json that looks like this:

```
{
  "tokenSetOrder": [
    "global",
    "light",
    "dark",
    "theme"
  ]
}
```

it opens up in Figma like this:

image

six7:
I think we want to integrate the reorder behavior for the folder-based sync in the UI before we ship this to users. Otherwise we have the $metadata object without much use, as users can't reorder themselves.

PR #939, opened 2022-06-27
Feat/read pro features in free

six7:
This PR adds features of the pro version in the free version in a read-only state.

- Allows Git multifile read with free version
- Allows theme switching in the free version
- Allows seeing branch in bottom left in the free version

However,
- Writing to git multifile will be in Pro
- Managing themes will be in Pro
- Using branch switcher will be in Pro


LiamMartens (approved):
@six7 loooks good, maybe just try to get the coverage equal or increased?

six7:
@LiamMartens Could you take a look at changes since review? I had to change a few things as we didn't yet disable the Push button, leading to a bit of a confusing UX.

Also, I'm not sure if I should test for `can't write when multiFile is disabled` as this is part of GitTokenStorage for which we don't have tests yet. What do you think? Feels like we should test that once we tackle that file.

LiamMartens (approved):

LiamMartens:
Just got to resolve some conflicts

PR #957, opened 2022-07-01
Ft 335 support gitlab self hosted with no groups

roppazvan:
* Changed the 'projects' logic so that the plugin fetches the projects using the search GitLab API functionality.
* This should also fix: https://brand-code.atlassian.net/browse/FT-331
* Removed some unnecessary code
* Updated the existing tests for the new changes


src/app/store/providers/__tests__/clientFactory.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      id: fullPath,
+      secret,
+      baseUrl,
+    } as unknown as GitlabCredentials;
+    await clientFactory(context, false);
+
+    expect(GitlabTokenStorage).toHaveBeenCalledWith(secret, repositoryId, fullPath, baseUrl);
+  });
+
+  it('should call change path if ther is a filepath', async () => {
```
LiamMartens, line 35:
```suggestion
  it('should call change path if there is a filepath', async () => {
```

src/app/store/providers/gitlab/gitlab.tsx
```diff
@@ -53,7 +59,7 @@ export function useGitLab() {
   }, [confirm]);
 
   const pushTokensToGitLab = useCallback(async (context: GitlabCredentials) => {
-    const storage = await storageClientFactory(context);
+    const storage = await storageClientFactory(context, multiFileSync);
```
LiamMartens, line 62: Is there a specific reason for passing the `multiFileSync` into the `storageClientFactory` each time?

LiamMartens (changes requested):
Can you check why the coverage is so much lower? It looks like it was lowered on a bunch of unrelated tests (might be you just need to pull in `next`)

Also you can remove `.DS_Store` and had a few other comments

roppazvan:
> 

Pulled in next &  removed ds_store. Not sure about the coverage. I asked Jan too about it. Because I only removed some code. Also, I added tests for code I did not touch. just to make sure the coverage check passes.

LiamMartens (approved):

PR #958, opened 2022-07-01
multiselect implementation

### LINKED ISSUE #245: Multi-select on space tokens' contextual menus
Whenever I'm managing space tokens on nodes, I find myself often working with several sides at once. The current method of applying multiple sides the same token requires individual assignment (right click, select Left, right click again, select Right).

A multi-select method of choosing which properties you'd like to apply a token to. Since the menu already displays a checkbox to show if the token is currently applied, can that checkbox be interactive?

**Describe alternatives you've considered**
Another option might be a keyboard shortcut like Shift or Command that would allow for the selection of multiple properties at once, without closing the contextual menu.

**Additional context**
This probably only applies to spacing tokens since that's the one that has more than two options + an 'All', but maybe would have value in others too like Color. If a keyboard shortcut were to be used, it could be something that works for all tokens that have multiple properties without cluttering up the UI.

fokkerone:
### Add Multiselection for Contextmenu (Closes #245)
- 2 New TS Lint rules
- add DS_Store in .gitignore
- add Prettier as Formater, add some formating rules
- Solve LF/CRLF line ending problem for mdx stories
- Create 2 new exports in contextmenu and use these as alternative menu item in contextmenu

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,17 +1,17 @@
 module.exports = {
-  extends: ['airbnb', 'airbnb-typescript', 'plugin:react-hooks/recommended'],
+  extends: ["airbnb", "airbnb-typescript", "plugin:react-hooks/recommended"],
   parserOptions: {
-    project: './tsconfig.json'
+    project: "./tsconfig.json",
   },
-  ignorePatterns: ['**/*.js'],
+  ignorePatterns: ["**/*.js"],
   globals: {
-    figma: 'readable',
-    __html__: 'readable',
-    describe: 'readable',
-    it: 'readable',
-    expect: 'readable',
-    cy: 'readable',
-    jest: 'readable',
+    figma: "readable",
+    __html__: "readable",
+    describe: "readable",
+    it: "readable",
+    expect: "readable",
+    cy: "readable",
+    jest: "readable",
   },
   rules: {
     "import/prefer-default-export": 0,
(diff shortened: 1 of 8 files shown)


.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
six7, line 1: We already have a `.prettierrc.yml`, you can remove that

src/app/components/MoreButton/MoreButton.tsx
```diff
@@ -47,32 +51,48 @@ export const MoreButton: React.FC<Props> = ({
...
-            key={property.name}
+          <MoreButtonPropertyMulti
+            key={`${property.name}-${property?.label}`}
             value={value}
             property={property}
             onClick={onClick}
           />
         ))}
+
+        {/* <ContextMenuSeparator /> */}
```
six7, line 75: This comment isn't needed, is it?

src/app/components/MoreButton/MoreButtonPropertyMulti.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const mainNodeSelectionValues = useSelector(mainNodeSelectionValuesSelector);
+  const isActive = React.useMemo(
+    () => mainNodeSelectionValues[property.name] === value,
+    [value, property, mainNodeSelectionValues],
+  );
+  const handleClick = React.useCallback(() => {
+    onClick(property, isActive);
+  }, [property, isActive, onClick]);
+
+  return (
```
six7, line 34: This current implementation doesn't work by using up/down arrow keys. It seems to just skip those multi entries.

.eslintrc.js
```diff
@@ -1,17 +1,17 @@
 module.exports = {
-  extends: ['airbnb', 'airbnb-typescript', 'plugin:react-hooks/recommended'],
+  extends: ["airbnb", "airbnb-typescript", "plugin:react-hooks/recommended"],
```
six7, line 2: can you revert those changes? we dont use double quotes in our prettier config.

six7 (changes requested):
Overall nice! some minor things. Also would love to get @LiamMartens review on this as well from a typescript perspective.

.prettierrc
fokkerone: sorry forgot to delete after testing. will do!

src/app/components/MoreButton/MoreButton.tsx
fokkerone: Nope

src/app/components/MoreButton/MoreButtonPropertyMulti.tsx
fokkerone: ok, dont have this on my agenda. See what I can do to implement the default menu A11Y from radix on this one too

.eslintrc.js
fokkerone: you haven´t a uniform style https://github.com/six7/figma-tokens/blob/main/.eslintrc.js
releates to .prettierrc

LiamMartens: I saw you added a `prettierrc` which is fine - but these should still change to single quotes I believe

src/stories/TokenButton.stories.mdx
```diff
@@ -1,156 +1,215 @@
-import { Meta, Story, ArgsTable, Canvas, Preview } from '@storybook/addon-docs';
-import { Provider } from 'react-redux';
+import { Meta, Story, ArgsTable, Canvas, Preview } from "@storybook/addon-docs";
+import { Provider } from "react-redux";
 
-import { store } from '@/app/store';
-import { SingleToken } from '@/types/tokens';
-import { TokenTypes } from '@/constants/TokenTypes';
-import { BoxShadowTypes } from '@/constants/BoxShadowTypes';
+import { store } from "@/app/store";
```
LiamMartens, line 4: I guess these should also revert back to single quotes

LiamMartens (changes requested):
Just formatting

six7:
Closing this in favor of #969

PR #961, opened 2022-07-03
Improve token tooltips

### LINKED ISSUE #878: Change the way aliases show up in tooltips
441F3200-13EE-47C2-A7F3-B852DB89922B

We currently first show both raw and resolved value in tooltips, leading to duplicate listings.

We should change this to show the resolved/raw value inline, so instead of
```
paddingLeft: {spacing.xs}
paddingRight: {spacing.md}
paddingLeft: 16
paddingRight: 32
```

it becomes:

```
paddingLeft: {spacing.xs} 16
paddingRight: {spacing.md} 32
```

### LINKED ISSUE #938: Typography/Shadow reference tooltips don't show rawValue
Given I have the following JSON

```
{
  "typography": {
    "regular": {
      "value": {
        "fontFamily": "Inter",
        "fontWeight": "16",
        "lineHeight": "24",
        "fontSize": "24",
        "letterSpacing": "16"
      },
      "type": "typography"
    },
    "alias": {
      "value": "{typography.regular}",
      "type": "typography"
    }
  }
}
```

and I hover over typography.alias

I expect to see `{typography.regular}` instead of individual key value pairs as input.

However, I see just empty key value pairs.

Actual:
image

Expected:
image

Same is true for shadow tokens referencing other shadow tokens.

six7:
Changes the way we render tooltips, especially how the resolved value is displayed.

- Change tooltip styling
- Optimize TokenButton storybook so we can actually test tooltips and their resolved values

| Property | Before | After |
|--------|--------|--------|
| Simple | image | image |
| Shadow | image | image |
| Typography | image | image |
| Composition | image | image | 

Fixes #878
Fixes #938


src/app/components/TokenTooltip/TokenTooltipContentValue.tsx
```diff
@@ -7,83 +8,99 @@ import { TokensContext } from '@/context';
...
-        </div>
-      ) : null
+      <Stack direction="column" align="start" gap={2} wrap>
+        {Object.entries(token.value).map(([property, value], index) => (
+          <SingleCompositionValueDisplay
+            key={seed(index)}
+            property={property}
+            value={value}
+            // @TODO strengthen the type checking here
+            resolvedValue={get(resolvedValue, property) as CompositionTokenValue}
```
LiamMartens, line 54: @six7 these typings need to be reviewed anyways. It is because the resolving happens separately from the tokens themselves so it is difficult to match the typings together because they can be many different types of values. That's also why I have those `@TODO` comments. Hopefully we can resolve this if/when we review our internal data structure

LiamMartens (approved):

PR #975, opened 2022-07-06
fix set values on node

### LINKED ISSUE #972: V 112.rc1: Text changes to undefined when setting fill
**Describe the bug**
Text changes to undefined when setting fill. I can't reproduce in V 111.

**To Reproduce**
1. Add some text
2. Set color fill with token

**Expected behavior**
Text not changing when adding fill. :)

https://user-images.githubusercontent.com/388617/177497352-d8bb1c78-14ae-4a5e-b69e-eeae4e836366.mov

**Figma file (optional)**
https://www.figma.com/file/2eD2s8DPMUnsdvsor3mitA/Figma-Token-V-112.rc-1?node-id=0%3A1

**JSON (optional)**
N/A

six7:
Fixes #972

### DIFF
--- a/src/plugin/setValuesOnNode.ts
+++ b/src/plugin/setValuesOnNode.ts
@@ -14,8 +14,6 @@ export default async function setValuesOnNode(
   ignoreFirstPartForStyles = false,
 ) {
   try {
-    console.log('SET VALUES ON NODE', node, values, data, figmaStyleMaps);
-
     // BORDER RADIUS
     if (
       node.type !== 'CONNECTOR'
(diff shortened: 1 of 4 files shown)


src/plugin/setValuesOnNode.ts
```diff
@@ -203,7 +203,7 @@ export default async function setValuesOnNode(
       }
 
       // Real value for text layers
-      if (values.value !== null) {
```
swordEdge, line 206: It seems like that we only need `'value' in values`. We can remove `values.value !== null`.

PR #985, opened 2022-07-07
FT-160 feat: save token sort order in metadata

LiamMartens:
This now includes the UI for sorting


LiamMartens:
Note to self: need fix coverage

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "build-css": "postcss src/app/styles/tailwind.css -o src/app/styles/main.css",
     "watch-css": "postcss src/app/styles/tailwind.css -o src/app/styles/main.css -w",
-    "watch-js": "webpack --mode=development --watch",
+    "watch-js": "cross-env LAUNCHDARKLY_FLAGS=tokenThemes,gitBranchSelector,multiFileSync webpack --mode=development --watch",
```
six7, line 10: Can we remove this? I'd rather have flags be license key dependent so we can test things as they are in the real world

six7 (changes requested):
Getting `Error pushing to GitHub TypeError: Cannot read properties of undefined (reading 'forEach')` when trying to push changes to GitHub.

Also getting the same error for GitLab: `Error pushing to GitLab TypeError: Cannot read properties of undefined (reading 'forEach')`

And ADO: `Error pushing to ADO TypeError: Cannot read properties of undefined (reading 'forEach')`

six7 (approved):
Nice! 🎉

PR #991, opened 2022-07-08
432 token values should be trimmed on input and when passing to figma

### LINKED ISSUE #432: Token values should be trimmed on input and when passing to Figma
**Describe the bug**
#hex color sometimes won't work

**To Reproduce**
1. open plugin
2. create new color token
3. enter hex code
4. update style -> style with no property(in token panel it shows the color)

**Expected behavior**
when inputting hex code in the token panel and updating the color style, style should not be "no property". The property needs to align the color input in the token panel.

Screen Shot 2022-02-03 at 10 10 39 AM

**Figma file (optional)**

**JSON (optional)**

swordEdge:
Fixes #432

### DIFF
--- a/src/app/components/EditTokenForm.tsx
+++ b/src/app/components/EditTokenForm.tsx
@@ -88,7 +88,7 @@ function EditTokenForm({ resolvedTokens }: Props) {
     [internalEditToken],
   );
 
-  const handleBoxShadowChange = React.useCallback(
+  const handleBoxShadowValueChange = React.useCallback(
     (shadow: SingleBoxShadowToken['value']) => {
       setError(null);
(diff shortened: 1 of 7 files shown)


src/app/components/EditTokenForm.tsx
```diff
@@ -184,6 +184,23 @@ function EditTokenForm({ resolvedTokens }: Props) {
     [internalEditToken],
   );
 
+  const trimValue = React.useCallback((value: SingleToken['value']) => {
```
LiamMartens, line 187: Might be worth moving this to it's own utility function. That way we can more easily write a full coverage test for this as well.

LiamMartens (changes requested):

LiamMartens (approved):

PR #992, opened 2022-07-08
modify typography place holder

swordEdge:


LiamMartens (approved):

src/app/components/SingleTypographyDownShiftInput.tsx
```diff
@@ -34,7 +34,7 @@ export default function SingleTypographyDownShiftInput({
       handleChange={onChange}
       setInputValue={handleBoxshadowDownShiftInputChange}
       placeholder={
-        name === 'color' ? '#000000, hsla(), rgba() or {alias}' : 'Value or {alias}'
+        name === 'color' ? `${name} #000000, hsla(), rgba() or {alias}` : `${name} value or {alias}`
```
six7, line 37:
```suggestion
        `${name} value or {alias}`
```

as we are in `SingleTypographyDownShiftInput.tsx` this is never going to be `color`. We can remove that check for `color`.

six7 (changes requested):

LiamMartens (approved):

PR #1005, opened 2022-07-11
750 export typography alias incorrectly exported

swordEdge:


six7:
Nice! Resolve conflict then you can merge

six7 (approved):

PR #1007, opened 2022-07-11
872 adjust order of tokens in inspectannotate

swordEdge:
- [x] Functional review @six7


src/utils/convertToOrderObj.ts
```diff
@@ -0,0 +1,8 @@
+import { Properties } from "@/constants/Properties";
```
LiamMartens, line 1: Might be good to just write tests for these to get 100% coverage on those new utils

src/app/components/Initiator.tsx
```diff
@@ -21,6 +21,9 @@ import { notifyToUI } from '@/plugin/notifiers';
 import { StorageProviderType } from '@/constants/StorageProviderType';
```
LiamMartens, line 21: Looks fine but can we get the coverage to be equal or more

LiamMartens (changes requested):

six7:
Functionality works 👍

six7 (approved):

PR #1009, opened 2022-07-12
fix: FT-369 rename token group fix

LiamMartens:


LiamMartens:
Fixed in another branch

PR #1013, opened 2022-07-12
526 allow the currently active sync setting to be deleted

### LINKED ISSUE #526: Allow the currently active sync setting to be deleted
Currently I can only 'Add new credentials' or 'Edit' existing ones. I would like to be able to delete existing settings, especially since they (potentially) contain PAT for GitHub and other sensitive information. I'm not even able to fill in 'garbage info' afterwards because the plugin is then unable to connect to GitHub - a nice feature when you're adding a synch config the first time, less nice when you are trying anonymize data after initial add.

A delete button next to the 'Edit' option which permanently deletes the configuration would be ideal

swordEdge:
Fixes #526 

- [x] Functional review: @six7

### DIFF
--- a/src/app/components/Button/Button.tsx
+++ b/src/app/components/Button/Button.tsx
@@ -6,7 +6,7 @@ import { StyledButton } from './StyledButton';
 export interface ButtonProps {
   type?: 'button' | 'submit';
   form?: string
-  variant: 'secondary' | 'primary' | 'ghost' | 'warning';
+  variant: 'secondary' | 'primary' | 'ghost' ;
   onClick?: () => void;
   size?: 'large' | 'small';
   href?: string;
--- a/src/app/components/Button/Button.tsx
+++ b/src/app/components/Button/Button.tsx
@@ -6,7 +6,7 @@ import { StyledButton } from './StyledButton';
 export interface ButtonProps {
   type?: 'button' | 'submit';
   form?: string
-  variant: 'secondary' | 'primary' | 'ghost' ;
+  variant: 'secondary' | 'primary' | 'ghost';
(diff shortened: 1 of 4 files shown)


six7:
@swordEdge bit of functional feedback:

Remove the `Delete` button. Instead use what we already have: `Delete local credentials` beneath the repo. Sorry for causing extra work as this wasn't in the issue description. That way we don't introduce another state to that component.
image

src/app/components/Button/Button.tsx
```diff
@@ -6,7 +6,7 @@ import { StyledButton } from './StyledButton';
 export interface ButtonProps {
   type?: 'button' | 'submit';
   form?: string
-  variant: 'secondary' | 'primary' | 'ghost';
```
LiamMartens, line 9: Space can be removed

LiamMartens (approved):

six7:
@swordEdge could you add a test for that file so coverage passes?

src/app/components/StorageItem.tsx
```diff
@@ -23,14 +24,22 @@ const StorageItem = ({ item, onEdit }: Props) => {
   const branch = isGitProvider(item) ? item.branch : null;
 
   const { restoreStoredProvider, deleteProvider } = useRemoteTokens();
+  const { confirm } = useConfirm();
+
+  const askUserIfDelete = React.useCallback(async () => {
+    const shouldDelete = await confirm({
+      text: 'Do you really want to delete a sync setting?',
```
six7, line 31:
```suggestion
      text: 'Do you really want to delete this sync setting?',
```

PR #1014, opened 2022-07-12
854 UI crashing when entering reference pointing to typography token inside a different token type

swordEdge:


src/app/components/EditTokenForm.tsx
```diff
@@ -348,12 +349,7 @@ function EditTokenForm({ resolvedTokens }: Props) {
               <ColorPicker value={internalEditToken.value} onChange={handleColorValueChange} />
             )}
             {checkIfContainsAlias(internalEditToken.value) && (
-              <div className="flex p-2 mt-2 font-mono text-gray-700 bg-gray-100 border-gray-300 rounded text-xxs itms-center">
-                {internalEditToken.type === 'color' ? (
-                  <div className="w-4 h-4 mr-1 border border-gray-200 rounded" style={{ background: String(resolvedValue) }} />
-                ) : null}
-                {resolvedValue}
-              </div>
+              <ResolvedValueDisplay value={resolvedValue} isColorToken={internalEditToken.type === 'color'} />
```
LiamMartens, line 352: Can you use the `TokenTypes` enum; just for consistency

src/app/components/ResolvedValueDisplay.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { SingleTypographyToken } from '@/types/tokens';
+import { ResolvedShadowValueDisplay } from './ResolvedShadowValueDisplay';
+import { ResolvedTypograhpyValueDisplay } from './ResolvedTypograhpyValueDisplay';
+import { TokenBoxshadowValue, TokenTypograpyValue } from '@/types/values';
+
+type Props = {
+  value: string | number | TokenTypograpyValue | TokenBoxshadowValue | TokenBoxshadowValue[] | null;
```
LiamMartens, line 8: If this is supposed to take any value - you can also use `SingleToken['value']`

src/app/components/ResolvedValueDisplay.tsx
```diff
@@ -0,0 +1,38 @@
...
+  isColorToken: boolean;
+};
+
+export const ResolvedValueDisplay: React.FC<Props> = ({ value, isColorToken }) => {
+  if (Array.isArray(value)) {
+    return <ResolvedShadowValueDisplay shadows={value as TokenBoxshadowValue[]} />;
```
LiamMartens, line 14: Is the `as` necessary? I think due to the `isArray` type guard it should already resolve to be `TokenBoxshadowValue[]` since there are no other array types

src/app/components/ResolvedValueDisplay.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  if (value && typeof value === 'object') {
```
LiamMartens, line 17: Might be worth creating some generic type guard utilities in `utils/is`, similar to `isSingleTypographyToken`. Then you can also easily add coverage for those.

src/app/components/ResolvedValueDisplay.tsx
```diff
@@ -0,0 +1,38 @@
...
+    if ('fontFamily' in value) { // value is Typography value
+      return <ResolvedTypograhpyValueDisplay value={value as SingleTypographyToken['value']} />;
+    } if ('type' in value) { // value is BoxShadow value
+      return <ResolvedShadowValueDisplay shadows={[value as TokenBoxshadowValue]} />;
+    }
+    return (
+      <div className="flex p-2 mt-2 font-mono text-gray-700 bg-gray-100 border-gray-300 rounded text-xxs itms-center">
```
LiamMartens, line 24: Are we able to update these to stitches since we're moving them now anyways - might as well rework them (if not too much work)

LiamMartens (changes requested):

src/app/components/ResolvedValueDisplay.tsx
swordEdge: I discussed this with Jan and right now it is decided that we just resolve the bug so that there is no more crash in the plugin.
https://brandcode-team.slack.com/archives/C03C1F17012/p1657708093676259

LiamMartens (approved):

PR #1039, opened 2022-07-13
expand composition token

swordEdge:


src/app/components/modals/ExportModal.tsx
```diff
@@ -36,6 +37,10 @@ export default function ExportModal({ onClose }: Props) {
     setExpandShadow(!expandShadow);
   }, [expandShadow]);
 
+  const handleToggleExpandComposition = React.useCallback(() => {
+    setExpandCompoexpandComposition(!expandComposition);
```
LiamMartens, line 41: Just a typo but I assume this should be `setExpandComposition`

src/utils/convertToDefaultProperty.ts
```diff
@@ -0,0 +1,39 @@
+import { Properties } from '@/constants/Properties';
+
+export function convertToDefaultProperty(property: string): string {
```
LiamMartens, line 3: Could you just add a coverage test for this?

LiamMartens (changes requested):

LiamMartens (approved):

PR #1043, opened 2022-07-13
Fix/gitlab remove tokenset and duplicate

### LINKED ISSUE #1006: Renaming a token group name in a duplicated token set renames in the original set as well
**Describe the bug**
Renaming a token group name in a duplicated token set renames in the original set as well

**To Reproduce**
1. Duplicate a token set and rename it 
2. On the Duplicated token set, rename a token group
3. Check the original token set
4. The token group in the original token set is also renamed

**Expected behavior**
The token group in the original token set should not be affected by the renaming in the duplicated token set.

https://user-images.githubusercontent.com/66366384/178239522-b88d1119-6fbc-475c-a2f0-c7ce5657f5df.mov

**Figma file (optional)**

**JSON (optional)**

LiamMartens:
Fixes #1006 

- [x]  functional review: @esthercheran

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,3 +1,4 @@
+require('dotenv').config()
 /*
  * For a detailed explanation regarding each configuration property and type check, visit:
  * https://jestjs.io/docs/en/configuration.html
--- a/src/app/store/models/tokenState.test.ts
+++ b/src/app/store/models/tokenState.test.ts
@@ -89,6 +89,7 @@ describe('editToken', () => {
               updatedTokens: [],
             },
             activeTokenSet: 'global',
+            themes: [],
           },
         },
       },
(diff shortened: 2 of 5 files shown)


LiamMartens:
@esthercheran this contains the fix for the mutlifile push in Gitlab and the duplicate + rename issue. Let's test this before release of course; also wondering if we should provide a beta package to verify this fix because the issue is very specific.

six7:
@esthercheran can you functionally review this? I'm not aware of what to look for when reviewing this.

esthercheran:
@LiamMartens @six7 I think we should release a beta as the Gitlab push issue is very specific to the commit message rule. I have tested the other functionality, it is working fine.

LiamMartens:
This also fixes #1006

LiamMartens:
@six7 @esthercheran checked this with the user in the figma tokens slack

six7 (approved):

PR #1047, opened 2022-07-14
584 after duplicating a token take user straight to the token edit form

swordEdge:
- [x] Functional review: @six7


six7:
@swordEdge this isn't working yet. The EditTokenForm pops up, which is as expected, but hitting `Duplicate` doesn't do anything.

six7:
Functionality approved!

src/app/components/TokenTree.tsx
```diff
@@ -10,7 +10,7 @@ import AddIcon from '@/icons/add.svg';
 
 export type ShowFormOptions = {
   name: string;
-  isPristine?: boolean;
+  status?: string;
```
LiamMartens, line 13: Should this be typed as `EditTokenFormStatus` perhaps?

src/types/tokens/EditTokenObject.ts
```diff
@@ -21,7 +21,7 @@ import type { TokenTypeSchema } from './TokenTypeSchema';
 
 type EditTokenObjectProperties = {
   initialName: string;
-  isPristine: boolean;
+  status: string;
```
LiamMartens, line 24: Same here, this should probably be the enum you created.

LiamMartens (changes requested):

six7 (approved):

PR #1059, opened 2022-07-15
feat: implement styles tab

LiamMartens:


six7:
@LiamMartens is this ready for review?

LiamMartens:
@six7 yea I left a comment in Jira actually

six7 (changes requested):
Nice one! Gave it a functional test and works fine!

**Minor:** The `triangle` in the Accordion currently doesnt transform 90deg when it's collapsed. Should show ▶ when collapsed and ▼ when expanded.

**Minor:** Can you truncate the style names? Make sure to add a `title` attribute so the full style name is visible on hover though.

image

Reference:
image

**Minor:** Can you increase the click size of the triangle in the accordion? Right now it's `13px` which makes it hard to click. The icon size could stay the same but add more padding around it so we get to `24px`

**Metrics:**
I mentioned a few metrics that I'd like to track, in addition could we also track how users are using the Themes tab? Basically just add
- Created theme
- Edited theme
- Deleted theme
- Applied theme (using the dropdown or in the Themes tab)

And the mentioned
- Attach styles
- Unlink styles

LiamMartens:
@six7 I made the changes you requested :+1:

six7 (approved):
Nice! Merging it in and sending that version to users to gather feedback.

PR #1060, opened 2022-07-16
docs(token-transformer): add repository field to package.json

MH4GF:
This change provides npm package pages to display reference links to GitHub repositories.

# motivation

I learned about Token Transformer from [here](https://docs.tokens.studio/sync/github#7-how-to-use-tokens-stored-in-github-in-development) in the documentation and wanted to read the source code to learn more about its behavior.
However, I could not find it by googling and it took me a while to realize that it is located in a subdirectory of figma-tokens.

Therefore, I thought it would be easier for others to reference if a link was posted on the npm page.


six7:
Thanks!

PR #1063, opened 2022-07-18
1016 collapsing all the tokens at once using optionalt+click stopped working since the latest update

swordEdge:
- [x] functional review: @esthercheran

### DIFF
--- a/src/app/components/TokenListing.tsx
+++ b/src/app/components/TokenListing.tsx
@@ -74,6 +74,14 @@ const TokenListing: React.FC<Props> = ({
     }
   }, [dispatch, collapsedTokenTypeObj, tokenKey]);
 
+  React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/app/components/TokenListing.tsx
```diff
@@ -69,10 +69,15 @@ const TokenListing: React.FC<Props> = ({
     e.stopPropagation();
     if (e.altKey) {
       dispatch.uiState.toggleCollapsed();
+      const newCollapsedObj = Object.keys(collapsedTokenTypeObj).reduce<Record<string, boolean>>((acc, key) => {
```
six7, line 72: can you add test coverage for this whole scenario? basically collapsing and expanding of a token set with a click and another scenario with the alt key. expected would be that the token group is collapsed, and in the case of alt key that all token sets are collapsed.

six7 (changes requested):
Need to add test coverage

six7 (approved):

PR #1071, opened 2022-07-19
modify modal description

swordEdge:
- [ ] Functional review @esthercheran


esthercheran:
@swordEdge In the ticket the proposed message was:
                      Recover local changes?
You have local changes unsaved to the remote storage.
No                                                                            Yes

When the user clicks No, the plugin should pull from remote. On clicking Yes, the plugin should not pull from remote.

@six7 if you can confirm this.

six7:
Correct, we'd need to invert that logic. Let's do

**Recover local changes?**
`You have local changes unsaved to the remote storage.`
`[No, pull]` | `[Yes, recover]`

@swordEdge we need to invert the `Confirm` button logic - When the user confirms the dialog, the plugin should Recover local changes. When they cancel the dialog, the plugin should pull from remote.

six7 (approved):

six7:
there's a conflict, you can merge once that is resolved

PR #1073, opened 2022-07-19
922 token name validation doesnt respect prior levels of nesting

swordEdge:


six7:
@swordEdge this is identical to https://github.com/six7/figma-tokens/pull/1072 ?

PR #1075, opened 2022-07-19
Change import references to new notation

### LINKED ISSUE #1074: Change syntax of references to { when importing styles
Currently when we import typography styles we import references to atomic tokens with the old `$` notation. We should update that to use the preferred `{alias}` notation.

six7:
Fixes #1074

- [x] Functional review @six7


LiamMartens (changes requested):
Can you update the coverage? :grimacing:

six7:
> Can you update the coverage? 😬

Was just a base branch issue. I created the PR originally based on the `next` branch which caused coverage to seem decreased - i never added new logic though. Changing base and pushing again fixed it.

PR #1076, opened 2022-07-20
869 improve error messages when connecting to a remote repository

swordEdge:


six7 (changes requested):
I'd love it if we would also display `validation` errors, as in the contents failed to pass the validation checks we have in the various TokenStorage.ts files (e.g. in https://github.com/six7/figma-tokens/blob/32f8cdffe157812374059789000d8995b179848f/src/storage/FileTokenStorage.ts#L44) -> we should show the user then that their file contents don't pass the schema, e.g. `Contents don't pass schema validation`

src/storage/ADOTokenStorage.ts
```diff
@@ -309,6 +311,9 @@ export class ADOTokenStorage extends GitTokenStorage {
           })),
         ];
       }
+      return {
+        errorMessage: ErrorMessages.VALIDATION_ERROR,
+      };
     } catch (e) {
```
six7, line 317: should we return an errormessage as well when an error is caught?

src/storage/RemoteTokenStorage.ts
```diff
@@ -71,33 +76,43 @@ export abstract class RemoteTokenStorage<Metadata = unknown, SaveOptions = unkno
...
-        data.tokens = {
-          ...data.tokens,
-          ...parseTokenValues({ [file.name]: file.data }),
-        };
-      } else if (file.type === 'metadata') {
-        data.metadata = {
-          ...data.metadata,
-          ...file.data,
-        };
+    const content = await this.read();
```
six7, line 87: should we keep `files` here? i feel it's easier to read + the singular is `file` later on as well

src/storage/__tests__/GithubTokenStorage.test.ts
```diff
@@ -154,49 +155,66 @@ describe('GithubTokenStorage', () => {
     });
   });
 
-  it('can read from Git in single file format', async () => {
-    mockGetContent.mockImplementationOnce(() => (
+  // it('can read from Git in single file format', async () => {
```
six7, line 158: i think we'll still want that test

six7 (changes requested):
havent functionally tested this yet but went over the code

src/storage/ADOTokenStorage.ts
swordEdge: This is the case when the content failed validation schema

swordEdge: This is the case when the content failed validation schema.

LiamMartens (approved):

six7 (approved):

PR #1087, opened 2022-07-25
feat: Startup logic

LiamMartens:
@six7 I streamlined the startup logic as I had mentioned before. Now the plugin data will be pre-fetched in one go and sent to the front-end where it will be handled in a single step-by-step process reducing the amount of round-trips and making the plugin logic more clear to reduce the possibility for errors.

I did alter the loading behavior of the plugin slightly where the initial pulling of the tokens will now also happen during the loading screen, but the loading screen has also been updated with more clear information for the user (clearly stating the different steps; such as verifying the license, getting the LD flags etc, ...).

The process now also ensures the LD flags are ready before initializing any other logic in the plugin ensuring the right features are available to the user at the right time.

As a TODO to myself - I need to adjust the tests where required


LiamMartens:
https://user-images.githubusercontent.com/5265324/180692517-f7a59109-eedd-4ff7-a068-65ee9704abc8.mp4

six7:
This is awesome, I love the various loading states! I tried with a couple different scenarios but will do some more testing and then look at the code. Thank you for this! 🎉

LiamMartens:
@six7 updated all the tests etc..

src/app/components/AppContainer/__tests__/AppContainerIntegration.test.tsx
```diff
@@ -0,0 +1,363 @@
...
+    identify: () => Promise.resolve(),
+  }),
+  useFlags: () => ({}),
+}));
+
+describe('AppContainer (integration)', () => {
+  beforeEach(resetSuite);
+
+  beforeAll(() => {
+    // useLDClientSpy.mockImplementation(() => ({
```
six7, line 194: is this supposed to commented out?

src/app/components/AppContainer/__tests__/AppContainerIntegration.test.tsx
```diff
@@ -0,0 +1,363 @@
...
+    //   // identify: () => Promise.resolve(),
+    // } as any));
+    // useFlagsSpy.mockImplementation(() => ({
+    // }));
+  });
+
+  afterAll(() => {
+    resetSuite();
+    // useFlagsSpy.mockReset();
```
six7, line 203: same, do we need those comments?

src/app/components/AppContainer/startupProcessSteps/saveStorageInforamationFactory.ts
```diff
@@ -0,0 +1,16 @@
+import useStorage from '@/app/store/useStorage';
+import type { Dispatch } from '@/app/store';
+import type { StartupMessage } from '@/types/AsyncMessages';
+
+export function saveStorageInforamationFactory(
```
six7, line 5:
```suggestion
export function saveStorageInformationFactory(
```

should probably change any occurrence of this typo

src/app/components/LaunchDarkly/LDProvider.tsx
```diff
@@ -12,8 +11,7 @@ const ldClientSideId = process.env.LAUNCHDARKLY_SDK_CLIENT || '';
 
 export const LDProviderWrapper = ({ children }: LDProviderProps) => {
   const userId = useSelector(userIdSelector);
-  const licenseKey = useSelector(licenseKeySelector);
-  if (userId && licenseKey) {
+  if (userId) {
```
six7, line 14: just to confirm: was removing license key here intentional? i remember we limited this to just call launchdarkly when a license key was present to reduce the amount of API calls we need (as those have a plan limit)

src/utils/uiSettings.ts
```diff
@@ -62,9 +62,12 @@ export async function getUISettings(): Promise<SavedSettings> {
         inspectDeep,
       };
 
-      notifyUISettings(settings);
+      if (notify) {
+        notifyUISettings(settings);
+      }
     }
   } catch (err) {
+    console.error(err);
```
six7, line 70: just so i understand: when should we use `handleReactError` and when just a console err? It feels like we'd always want `handleReactError` as we can then track these?

six7 (changes requested):
haven't fully functionally tested yet but went through code and had just a few minor things to note. this is great, way more understandable!

six7:
One thing I noticed: If there's a credentials error during startup the plugin stays in `Fetching credentials` loading state forever. This happened with ADO sync.

image

src/utils/uiSettings.ts
LiamMartens: The `uiSettings.ts` file is used on the plugin side - not the react side; we can't really track much on the plugin side due to the sandboxing - so `handleReactError` should only be used in the UI side

src/app/components/LaunchDarkly/LDProvider.tsx
LiamMartens: Ahh I see - I'll re-add it and add a comment in code

LiamMartens:
@six7 the error occurs if the local api credentials are invalid or something?

LiamMartens:
@six7 issue with the loading state should be fixed

six7 (approved):

six7:
Can you resolve merge conflicts? After that let's merge this in.

PR #1110, opened 2022-08-03
add expandComposition

swordEdge:


six7 (changes requested):
Could you also add a few composition tokens in the https://github.com/six7/figma-tokens/tree/main/token-transformer/tokens files and make sure that we test that from a token-transformer level? 

The "tests" there are basically running token-transformer with a few inputs that should then produce certain outputs: https://github.com/six7/figma-tokens/blob/main/token-transformer/package.json

LiamMartens (approved):

six7 (changes requested):
Can you add a test for this in `token-transformer/transform.test.js`? You'll need to add a `output/composition.json` so the test know what is the expected result, you can look at the other tests to see how that was done.

token-transformer/package.json
```diff
@@ -10,6 +10,7 @@
   "main": "index.js",
   "scripts": {
     "build-typography-test": "node index.js tokens.json temp/typography.json base/options,base/sizing,core/typography base/options,base/sizing",
+    "build-composition-test": "node index.js tokens.json temp/composition.json base/options,base/sizing,core/composition base/options,base/sizing",
```
six7, line 13:
```suggestion
    "build-composition-test": "node index.js tokens.json temp/composition.json base/options,base/sizing,core/composition base/options,base/sizing --expandComposition=true",
```

six7: this needs to be enabled to be run

six7 (approved):

PR #1118, opened 2022-08-04
read-only-permissions-on-organization-github-account-are-not-respected

swordEdge:


src/storage/GithubTokenStorage.ts
```diff
@@ -121,83 +121,112 @@ export class GithubTokenStorage extends GitTokenStorage {
...
-            path: item.path,
-            sha: item.sha,
-            mode: getTreeMode(item.type),
-          })),
-          headers: octokitClientDefaultHeaders,
-        });
-
-        if (directoryTreeResponse.data.sha) {
-          const treeResponse = await this.octokitClient.rest.git.getTree({
+        let treeResponse;
```
LiamMartens, line 124: I see quite a lot has changed here - what was the actual fix?

LiamMartens (changes requested):

src/storage/GithubTokenStorage.ts
swordEdge: On organization github account, users are not allowed to create a tree with read-only permission.
And in the current logic, we create a tree object to get contents in the directory recursively.
That's why this issue happening.
So I changed the logic where we get a tree object not creating that one when creating a tree object failed.

LiamMartens: I see - so does the reading of the entire directory still work then with read-only permission? And if it does then can we cut out creating a tree altogether?

swordEdge: Yes. reading of the entire directory still workwith read-only permission.
Yes. we can cut out creating a tree altogether

LiamMartens: @six7 should we then cut out the `createTree` step here altogether? I assume it will make it faster since it takes out a API call.

six7: Lets remove it then 👍

swordEdge: @six7 @LiamMartens  Removed the createTree step.

LiamMartens (approved):

six7 (changes requested):
This currently fails when the filePath is set to a nested directory that didn't exist yet:

To reproduce:
- Go to sync settings of GitHub
- Change filePath to `some/random/path/that/doesntexist` (or anything else that is nested
- When the plugin asks you to push, you push

Notice that there's an error.

six7 (approved):

PR #1137, opened 2022-08-10
Fix/update adding and removing licenses

LiamMartens:


LiamMartens:
@six7 this should fix the issue with adding the license keys etc..

six7 (approved):

PR #1151, opened 2022-08-11
feat: Added CTRL+S to save updated json

### LINKED ISSUE #448: Option to Save & Update JSON with keyboard shortcut.
Currently I am working with creating Color values based on hue range. 
I have to edit the hue value and hit save and update. Mouse travel time is high.

It would be helpful if we can save json with a keyboard shortcut.

LiamMartens:
Fixes #448


src/app/components/TokensBottomBar.tsx
```diff
@@ -21,6 +22,25 @@ export default function TokensBottomBar({ handleUpdate, handleSaveJSON, hasJSONE
   const [exportModalVisible, showExportModal] = React.useState(false);
   const [presetModalVisible, showPresetModal] = React.useState(false);
 
+  useShortcut(['ControlLeft', 'KeyS'], handleSaveJSON);
```
six7, line 25: can we change this from `CTRL` to the `Meta` key (ctrl on windows, cmd on Mac)? Mac users will be used to the Cmd + S shortcut

six7 (changes requested):
just one comment about usage of Meta key instead of ctrl

six7 (approved):

PR #1172, opened 2022-08-16
fix: URL and JSONbin storage

### LINKED ISSUE #1164: URL Sync doesn't respect $themes or $metadata
When using URL sync with a .json stored like this:

```
{
  "global": {
    ...
  },
  "dark: {
    ...
  },
  ...
  "$themes": [
    {
      "id": "864976136bac5e6e5c63f8d3517bf5190003b738",
      "name": "Dark",
      "selectedTokenSets": {
        "global": "source",
        "light": "disabled",
        "dark": "enabled",
        "theme": "enabled"
      }
    },
    {
      "id": "60369cb76e892f243878189e417965b9e8eb6b52",
      "name": "Light",
      "selectedTokenSets": {
        "global": "source",
        "light": "enabled",
        "dark": "disabled",
        "theme": "enabled"
      }
    }
  ],
  "$metadata": {
    "tokenSetOrder": [
      "global",
      "light",
      "dark",
      "theme"
    ]
  }
}
```

We do not respect or even pickup $themes or $metadata at all, resulting in an empty themes array.

Same is true for the jsonbin-like format where we have a `values` property.

LiamMartens:
Updated URL storage provider to support themes and metadata
Updated JSONbin to support metadata

Fixes #1164


LiamMartens:
@six7 I did notice JSONBin currently always asks to recover local changes (also in the live plugin), We may want to review this behavior - but it's a bit of a weird one because it's the only provider type we automatically sync as opposed to the git providers. Perhaps there is merit in just aligning the logic altogether and also making this a manual action like the git providers?

six7 (changes requested):
## Two `version` and `updatedAt` placements
We now have 2 places where we track updatedAt and version, is this intended? I really like placing it in metadata fwiw!

image

image

Other than that this works as intended 🙏  Would love to know the thoughts behind the two version and updatedAt placements

six7:
> @six7 I did notice JSONBin currently always asks to recover local changes (also in the live plugin), We may want to review this behavior - but it's a bit of a weird one because it's the only provider type we automatically sync as opposed to the git providers. Perhaps there is merit in just aligning the logic altogether and also making this a manual action like the git providers?

Yeah agree, we should align as much as we can, making this a manual pull/push action would be great. I'll create an issue

six7 (approved):

PR #1178, opened 2022-08-16
1155 export doesnt include metadata or themes

swordEdge:


LiamMartens (approved):

src/app/store/remoteTokens.tsx
```diff
@@ -315,23 +316,23 @@ export default function useRemoteTokens() {
     track('fetchTokensFromFileOrDirectory');
     dispatch.uiState.startJob({ name: BackgroundJobs.UI_FETCHTOKENSFROMFILE });
 
-    let remoteData: RemoteResponseData<unknown> | null = null;
     if (files) {
-      remoteData = await readTokensFromFileOrDirectory(files);
+      const remoteData = await readTokensFromFileOrDirectory(files);
       if (remoteData?.status === 'success') {
+        const sortedTokens = applyTokenSetOrder(remoteData.tokens, remoteData.metadata?.tokenSetOrder ?? []);
```
six7, line 322:
```suggestion
        const sortedTokens = applyTokenSetOrder(remoteData.tokens, remoteData.metadata?.tokenSetOrder ?? Object.keys(remoteData.tokens);
```

If no metadata is present we should respect the order of the token sets in the JSON

six7 (changes requested):
Few things that dont work:

- When we are exporting a **single token set** we don't want $themes or $metadata to be included (Both with "Include parent keys" and without that)
- Importing a .json file that was exported with $metadata shows $metadata as a token set, which it shouldn't.
image

six7 (approved):

PR #1191, opened 2022-08-19
1099-improve-the-list-of-available-tokens-in-a-dropdown

swordEdge:
Currently when you are replacing a token via Inspect Mode, you get a list of all the available tokens. This list is only not sorted in a proper way and seems to sort on create date.
In the past:
image

After fixing: 
image


src/app/components/DownshiftInput/DownshiftInput.tsx
```diff
@@ -140,7 +140,15 @@ export const DownshiftInput: React.FunctionComponent<DownShiftProps> = ({
       .filter(
         (token: SingleToken) => !filteredValue || token.name.toLowerCase().includes(filteredValue.toLowerCase()),
       )
-      .filter((token: SingleToken) => token?.type === type && token.name !== initialName),
+      .filter((token: SingleToken) => token?.type === type && token.name !== initialName).sort((a, b) => {
+        if (a.name < b.name) {
```
LiamMartens, line 144: Can we just use `a.name.localeCompare(b.name)` instead?

LiamMartens (changes requested):

six7 (approved):

PR #1197, opened 2022-08-19
807 cannot set a number fontweight in typography

swordEdge:
We can't set fontWeight with a number between 100 to 900 which is the most common way to set it.

After fixing:
In Figma, there is a match rule between font family and font-weight, so every font weight doesn't work to every font family.
To test this, we can use `Montserrat` fontfamily which support all font weights from 100 to 900.

### DIFF
--- a/src/plugin/setTextValuesOnTarget.test.ts
+++ b/src/plugin/setTextValuesOnTarget.test.ts
@@ -67,4 +67,23 @@ describe('setTextValuesOnTarget', () => {
       textCase: 'TITLE',
     });
   });
+
+  it('it throws error, when there is no value in token', async () => {
+    await setTextValuesOnTarget(textNodeMock, {
+      description: 'Use with care',
+    });
+    expect(textNodeMock).toEqual({
+      ...textNodeMock,
+    });
+  });
+
+  it('it does nothing when the type of value is string', async () => {
+    await setTextValuesOnTarget(textNodeMock, {
+      description: 'Use with care',
+      value: 'string'
+    });
+    expect(textNodeMock).toEqual({
+      ...textNodeMock,
+    });
+  });
 });
--- a/src/plugin/setTextValuesOnTarget.test.ts
+++ b/src/plugin/setTextValuesOnTarget.test.ts
@@ -36,7 +36,7 @@ describe('setTextValuesOnTarget', () => {
     expect(textNodeMock).toEqual({ ...textNodeMock, fontName: { ...textNodeMock.fontName, style: 'Bold' } });
   });
 
-  it('convert number fontWeight and sets to the node', async () => {
+  it('converts a numerical fontWeight and sets to the node', async () => {
     loadFontAsyncSpy.mockImplementationOnce(() => (
       Promise.reject()
     ));
@@ -47,7 +47,7 @@ describe('setTextValuesOnTarget', () => {
     expect(textNodeMock).toEqual({ ...textNodeMock, fontName: { ...textNodeMock.fontName, style: 'Medium' } });
   });
 
-  it(`can't set number fontWeight to the node if there is no match fontWeight`, async () => {
+  it('can\'t set number fontWeight to the node if there is no matching fontWeight', async () => {
     loadFontAsyncSpy.mockImplementation(() => (
       Promise.reject()
     ));
--- a/src/plugin/setTextValuesOnTarget.test.ts
+++ b/src/plugin/setTextValuesOnTarget.test.ts
@@ -47,6 +47,14 @@ describe('setTextValuesOnTarget', () => {
     expect(textNodeMock).toEqual({ ...textNodeMock, fontName: { ...textNodeMock.fontName, style: 'Medium' } });
   });
 
+  it(`can't set number fontWeight to the node if there is no match fontWeight`, async () => {
+    loadFontAsyncSpy.mockImplementation(() => (
+      Promise.reject()
+    ));
+    await setTextValuesOnTarget(textNodeMock, { value: { fontWeight: '500' } });
+    expect(textNodeMock).toEqual({ ...textNodeMock, fontName: { ...textNodeMock.fontName } });
+  });
+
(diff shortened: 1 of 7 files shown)


src/plugin/setValuesOnNode.ts
```diff
@@ -29,7 +29,7 @@ export default async function setValuesOnNode(
   const stylePathPrefix = prefixStylesWithThemeName && activeThemeObject
     ? activeThemeObject.name
     : null;
-
+  console.log('values', values);
```
LiamMartens, line 32: logs

LiamMartens (changes requested):

src/plugin/setTextValuesOnTarget.ts
```diff
@@ -18,11 +19,12 @@ export default async function setTextValuesOnTarget(target: TextNode | TextStyle
       } = value;
       const family = fontFamily?.toString() || (target.fontName !== figma.mixed ? target.fontName.family : '');
       const style = fontWeight?.toString() || (target.fontName !== figma.mixed ? target.fontName.style : '');
-      await figma.loadFontAsync({ family, style });
+      const transformedValue = transformValue(style, 'fontWeights');
```
six7, line 22: Can we change this so we 
1) First try to set font weight as it is
2) If that fails try the mapping

Reason is that some (not a lot) fonts have weights actually named '500'

six7 (changes requested):

six7:
This works great so far!

One more thing that I feel would be really great if we could get in, as that way we'd probably be able to cover a lot more font weights

Some fonts define their weights not like `SemiBold` but `Semi Bold` (notice the space between Semi and Bold), or sometimes even `DemiBold`.

Could we pass in our mapping an array of possible weight names? Then we could extend that with some mappings that map in certain font weights. We'd then try to set each of them, and if there's a match, it will apply that?

In the case of `600` this would mean we'd return `['SemiBold', 'Semi Bold', 'DemiBold']`? We could use an array for each of these, as I'm sure we'd need to add more possible variations to the other weights later?

src/plugin/setTextValuesOnTarget.test.ts
```diff
@@ -35,6 +36,25 @@ describe('setTextValuesOnTarget', () => {
...
+      Promise.reject()
+    ));
+    loadFontAsyncSpy.mockImplementation(() => (
+      Promise.resolve()
+    ));
+    await setTextValuesOnTarget(textNodeMock, { value: { fontWeight: '500' } });
+    expect(textNodeMock).toEqual({ ...textNodeMock, fontName: { ...textNodeMock.fontName, style: 'Medium' } });
+  });
+
+  it(`can't set number fontWeight to the node if there is no match fontWeight`, async () => {
```
six7, line 50:
```suggestion
  it(`can't set number fontWeight to the node if there is no matching fontWeight`, async () => {
```

src/plugin/setTextValuesOnTarget.test.ts
```diff
@@ -35,6 +36,25 @@ describe('setTextValuesOnTarget', () => {
     expect(textNodeMock).toEqual({ ...textNodeMock, fontName: { ...textNodeMock.fontName, style: 'Bold' } });
   });
 
+  it('convert number fontWeight and sets to the node', async () => {
```
six7, line 39:
```suggestion
  it('converts a numerical fontWeight and sets to the node', async () => {
```

src/plugin/setTextValuesOnTarget.ts
```diff
@@ -18,14 +18,34 @@ export default async function setTextValuesOnTarget(target: TextNode | TextStyle
...
-      await figma.loadFontAsync({ family, style });
-      if (fontFamily || fontWeight) {
-        target.fontName = {
-          family,
-          style,
-        };
-      }
 
+      try {
+        await figma.loadFontAsync({ family, style });
```
six7, line 23: can we move the await inside the `if (fontFamily || fontWeight)`? Then it will only run if required

src/plugin/setTextValuesOnTarget.ts
```diff
@@ -18,14 +18,34 @@ export default async function setTextValuesOnTarget(target: TextNode | TextStyle
...
+            style,
+          };
+        }
+      } catch {
+        const candidateStyles = transformValue(style, 'fontWeights');
+        await Promise.all(
+          candidateStyles.map(async (candidateStyle) => (
+            await figma.loadFontAsync({ family, style: candidateStyle })
+              .then(() => {
+                if (fontFamily || fontWeight) {
```
six7, line 36: i think we don't even need this check? if we do, it should check for `family` and `candidateStyle` - as these are the properties we're setting here.

six7 (changes requested):
Nice! some minor comments

src/plugin/setTextValuesOnTarget.ts
```diff
@@ -18,14 +18,34 @@ export default async function setTextValuesOnTarget(target: TextNode | TextStyle
...
+            style,
+          };
+        }
+      } catch {
+        const candidateStyles = transformValue(style, 'fontWeights');
+        await Promise.all(
+          candidateStyles.map(async (candidateStyle) => (
+            figma.loadFontAsync({ family, style: candidateStyle })
+              .then(() => {
+                if (fontFamily || candidateStyle) {
```
six7, line 36: We dont need to check for fontFamily here as we're in an if condition that already checks for that on line 23.

So can just be a check for candidateStyle?

six7 (changes requested):
Small change + can we get test coverage to pass?

LiamMartens (approved):

six7:
Can we still get the test coverage to pass @swordEdge ? Then I'd merge this in 👍

six7 (approved):

jsamr:
@six7 Sorry to bother you! Has this fix been included in version 119?

six7:
We released this today with version 120 👍

PR #1211, opened 2022-08-22
1050-after-renaming-a-token-remember-the-previous-choice-in-the-remap-pop-up

swordEdge:
Would be nice if after renaming a token this popup remembers the previous choice. As usual, I almost always need to check “Document” but it always stays on “Page”.
179217967-9ea2c7b9-71ae-467b-8e15-1fca7a74d595

Right now: The plugin saves the previous choice as long as doesn't restart the plugin.


LiamMartens:
@six7 do we also need to remember this on the file or client ?

six7:
> @six7 do we also need to remember this on the file or client ?

Probably later on the client but let's see if the current solution satisfies that already.

src/app/store/models/uiState.tsx
```diff
@@ -122,6 +124,7 @@ export const uiState = createModel<RootModel>()({
     selectedLayers: 0,
     manageThemesModalOpen: false,
     scrollPositionSet: {},
+    currentUpdateMode: UpdateMode.DOCUMENT,
```
six7, line 127: can we change that back to `PAGE`? Document is something that takes a long time for some users so I'd rather keep the default to something that doesn't take minutes to complete

six7 (changes requested):
Let's also store this decision on the client (not file, but client). After testing it felt weird that it would always go back to another choice after restarting the plugin. Users will probably want to set this once, and then keep.

LiamMartens (approved):

PR #1214, opened 2022-08-23
1202-when-editing-a-token-of-a-custom-type-dont-overwrite-with-other

swordEdge:
Sometimes users want to use custom types of tokens, they can do so by specifying this in the JSON by just adding something custom to the "type" field. The plugin respects that, and will show that type as the category name.

However, when you edit that token in the UI, we overwrite that token type and reset it to "other".

After fixing: Token type is kept as it was.


six7 (commented):
Could we add a test scenario for this? So that we test if the user used a custom type and they edit it, we still keep this custom type?

LiamMartens (approved):

PR #1225, opened 2022-08-25
1224 plugin seems to reset window size when using deep inspect

swordEdge:
Fixed two issue.
https://github.com/six7/figma-tokens/issues/1223
https://github.com/six7/figma-tokens/issues/1224

The plugin seems to reset the window size when the user has

1. Changed window size
2. Then restarted the plugin
3. Then - without changing window size - changing Deep inspect
Notice how the plugin then resets the window size.

We seem to have caused a regression with Deep inspect:

To reproduce:

1. Activate deep inspect
2. Restart the plugin
3. Select a layer
4. Go to Inspect
5. Notice how even though it says Deep inspect is on, it only shows values of that current layer, not of children.
6. Turn Deep inspect off and on again
Notice that it now works

After fixing: Plugin keeps the window size and all children tokens are displayed after restarting the plugin.

### DIFF
--- a/src/app/components/AppContainer/startupProcessSteps/savePluginDataFactory.ts
+++ b/src/app/components/AppContainer/startupProcessSteps/savePluginDataFactory.ts
@@ -1,4 +1,4 @@
-import type { Dispatch } from '@/app/store';
+import { Dispatch } from '@/app/store';
 import type { StartupMessage } from '@/types/AsyncMessages';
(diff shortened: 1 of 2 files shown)


six7 (approved):
Nice! @LiamMartens maybe you can re-review this afterwards, I think some of the startup logic changes were causing this regression.

PR #1230, opened 2022-08-26
Add header for JSONBin to enable bin versioning

davidbroza:
By default, private bins in JSONBin have versioning turned off. This is not ideal for keeping backups of previous files published. This MR enables JSONBin versioning. 

To enable versioning, extra header needs to be added as described in: https://jsonbin.io/api-reference/bins/update 

Todo:
- [x] Versioning in JSONBin is available only for paid users, how can we address this within Figma Tokens plugin? (add toggle? / Check if user is free or paid via rest call ? )

### DIFF
--- a/src/storage/JSONBinTokenStorage.ts
+++ b/src/storage/JSONBinTokenStorage.ts
@@ -67,6 +67,7 @@ export class JSONBinTokenStorage extends RemoteTokenStorage<JsonBinMetadata> {
     this.defaultHeaders = new Headers();
     this.defaultHeaders.append('Content-Type', 'application/json');
     this.defaultHeaders.append('X-Master-Key', this.secret);
+    this.defaultHeaders.append('X-Bin-Versioning', 'true');
   }
 
   private convertJsonBinDataToFiles(data: JsonbinData): RemoteTokenStorageFile<JsonBinMetadata>[] {


six7:
> * [ ]  Versioning in JSONBin is available only for paid users, how can we address this within Figma Tokens plugin? (add toggle? / Check if user is free or paid via rest call ? )

Can this just be added and it still works for free users without erroring? We could just let users add this header manually in a custom Headers field maybe? We're aiming to consolidate the URL provider with the JSONBin provider to combine them, for URL we currently have a Headers field.

stramel:
Adding the header `x-bin-versioning` to a request to a FREE account will trigger an error. 

```
{"message":"Versioning is not available for the Free users. Upgrade to Pro plan http://api.jsonbin.io/pricing to avail this feature"}
```

I think allowing a header field would be a good option if we're consolidating JSONBin with URL.

six7:
> Adding the header `x-bin-versioning` to a request to a FREE account will trigger an error. 
> 
> 
> 
> ```
> 
> {"message":"Versioning is not available for the Free users. Upgrade to Pro plan http://api.jsonbin.io/pricing to avail this feature"}
> 
> ```
> 
> 
> 
> I think allowing a header field would be a good option if we're consolidating JSONBin with URL.

If we consolidate url and jsonbin to a generic url storage that allows both push/pull as outlined in https://github.com/six7/figma-tokens/pull/1140 this would solve this as we'd have a header there 👍

PR #1235, opened 2022-08-29
Adds support for $themes

forceofseth:


thomasmattheussen:
@forceofseth thanks for this!
I was wondering: shouldn't this also use $metadata.json to get the correct token set sort order?

six7:
Awesome, thank you! I'll hopefully get to functionally reviewing this in the next few days.

What @thomasmattheussen meant was that the sort order is relevant in the JSON itself as it has an implication on what token "wins" if multiple tokens define the same name.

I feel we could do this as a followup though, as with $themes you'd usually also activate/deactivate sets that would overwrite each other.

forceofseth:
@six7 I don't know why the tests with coverage are failing. I only can see an error in the GithubTokenStorage test. If i run tests locally on my machine the run without a problem.

thomasmattheussen:
> I feel we could do this as a followup though, as with $themes you'd usually also activate/deactivate sets that would overwrite each other.

@six7 is that how other designers think about it? Curious about how this differs from our setup.  Our designers think of dark mode as something that uses a light set as a source.

```
{
    "name": "Dark",
    "selectedTokenSets": {
      "Foundations": "source",
      "Light": "source",
      "Dark": "enabled"
    }
}

six7:
> @six7 I don't know why the tests with coverage are failing. I only can see an error in the GithubTokenStorage test. If i run tests locally on my machine the run without a problem.

We have a check that makes sure test coverage doesn't decrease, however that one seems to have an issue with contributions by outside contributors. Given you've added a test that checks results we can ignore this one I'd say

forceofseth:
@six7  can you give me an update regarding the review progress?

six7 (approved):
Sorry for the delay. I'll release this over the weekend!

Thank you so much for this! 🙏  🎉

six7:
Edit: Nevermind, I released this just now - no reason to wait 👍

thomasmattheussen:
@six7 is there a ticket that tracks this? :

> I feel we could do this as a followup though, as with $themes you'd usually also activate/deactivate sets that would overwrite each other.

This is severly blocking us atm.

six7:
> @six7 is there a ticket that tracks this? :
> 
> 
> 
> > I feel we could do this as a followup though, as with $themes you'd usually also activate/deactivate sets that would overwrite each other.
> 
> 
> 
> This is severly blocking us atm.
> 
> 
> 
> Could you maybe give a pointer as to where this should be fixed? Maybe we can look into it ourselves.

Sorry, I didnt get to this yet. I'll create one in a bit and include hints where this can be done (should be as easy as changing the order of set names according to the metadata file before passing it into the transform fn). Happy for you to pick this up if you'd like!

thomasmattheussen:
It'd be great if you could pick this up! 

I've already tried manually shifting sets around in $themes.json, but that didn't alter the outcome whatsoever. I get the feeling from your comment that this shouldn't be the case though?

thomasmattheussen:
@six7 I managed to create a fix for this: https://github.com/six7/figma-tokens/pull/1418

PR #1236, opened 2022-08-30
1135 allow bulk remap

swordEdge:
Users can remap several tokens at once.
Users can click bulk remap button in Inspect Tab and type part of the old name and new name.
Then all tokens are remaped with new name
image

### DIFF
--- a/src/app/components/modals/BulkRemapModal.tsx
+++ b/src/app/components/modals/BulkRemapModal.tsx
@@ -20,20 +20,22 @@ export default function BulkRemapModal({ isOpen, onClose }: Props) {
     onClose();
   }, [onClose]);
 
-  const onConfirm = React.useCallback(() => {
-    handleBulkRemap(newToken, matchingToken);
+  const onConfirm = React.useCallback(async () => {
+    await handleBulkRemap(newToken, matchingToken);
     onClose();
-  }, []);
+  }, [handleBulkRemap, onClose, newToken, matchingToken]);
 
-  const handleMatchingTokenChange = React.useCallback<React.ChangeEventHandler<HTMLInputElement>>((e) => {
-    e.persist();
-    setMatchingToken(e.target.value);
-  }, [matchingToken]);
+  const handleMatchingTokenChange = React.useCallback<React.ChangeEventHandler<HTMLInputElement>>(
+    (e) => {
+      e.persist();
+      setMatchingToken(e.target.value);
+    }, [matchingToken]);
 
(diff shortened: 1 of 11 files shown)


src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -0,0 +1,34 @@
...
+import { updatePluginData } from '../pluginData';
+import { sendSelectionChange } from '../sendSelectionChange';
+
+export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK_REMAP_TOKENS] = async (msg) => {
+  try {
+    const { oldName, newName } = msg;
+    const allWithData = await defaultNodeManager.findNodesWithData({});
+    const updatedNodes: NodeManagerNode[] = [];
+
+    allWithData.forEach((node) => {
```
six7, line 13: can we add a test for this file? so we really test that it's replacing those that need replacing and doesn't interfere with those that don't need replacing?

six7 (commented):
functionally this works as expected 👍

src/app/components/modals/BulkRemapModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    e.persist();
+    setNewName(e.target.value);
+  }, [newName]);
+
+  return (
+    <Modal large isOpen={isOpen} close={handleClose}>
+      <form
+        onSubmit={onConfirm}
+      >
+        <Stack direction="column" gap={4} css={{ minHeight: '215px', justifyContent: 'center' }}>
```
six7, line 43: Lef's remove minHeight, the content itself should give the height here

src/app/components/modals/BulkRemapModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    setOldName(e.target.value);
+  }, [oldName]);
+
+  const handleNewNameChange = React.useCallback<React.ChangeEventHandler<HTMLInputElement>>((e) => {
+    e.persist();
+    setNewName(e.target.value);
+  }, [newName]);
+
+  return (
+    <Modal large isOpen={isOpen} close={handleClose}>
```
six7, line 39: Can you move the Header to the title prop of the Modal? We're doing it like this for other Modals already. Also lets enable the showClose prop

six7 (changes requested):
Just some minor modal styling comments

six7 (approved):

PR #1242, opened 2022-09-01
1195 when deleting a token ask to delete any local styles

swordEdge:
When deleting a token, users can choose whether to delete related style either.
If the user selects an option, styles are removed from both the document style and the activeTheme
185567624-f99ee43c-2fd3-462e-baa2-a6aa5c7a733c


src/utils/is/isMatchingStyle.ts
```diff
@@ -0,0 +1,9 @@
+import { DeleteTokenPayload } from '@/types/payloads';
+
+export function isMatchingStyle(style: PaintStyle | TextStyle | EffectStyle, token: DeleteTokenPayload) {
```
LiamMartens, line 3: There is a util called `convertTokenNameToPath` which could be used here instead. Perhaps in combination with `matchStyleName`. See `setValuesOnNode.ts` line `80` for an example

src/app/store/useTokens.tsx
```diff
@@ -189,6 +190,16 @@ export default function useTokens() {
...
   }, [confirm, usedTokenSet, tokens, settings, dispatch.tokenState]);
 
+  const removeStylesFromTokens = useCallback(async (token: DeleteTokenPayload) => {
+    track('removeStyles', token);
+
+    await AsyncMessageChannel.ReactInstance.message({
+      type: AsyncMessageTypes.REMOVE_STYLES,
+      token,
+    });
+    // dispatch.tokenState.assignStyleIdsToCurrentTheme(createStylesResult.styleIds);
```
LiamMartens, line 200: We should probably implement the logic to also remove the token from the theme styles.

LiamMartens (changes requested):

src/app/store/models/reducers/tokenState/removeStyleIdsFromCurrentTheme.ts
```diff
@@ -0,0 +1,23 @@
+import type { TokenState } from '../../tokenState';
+
+export function removeStyleIdsFromCurrentTheme(state: TokenState, styleIds: Record<string, string>): TokenState {
+  // ignore if there is no active theme
+  if (!state.activeTheme) return state;
+  // ignore if the theme does not exist for some reason
+  const themeObjectIndex = state.themes.findIndex(({ id }) => state.activeTheme === id);
+  if (themeObjectIndex === -1) return state;
+  const tokenToRemove = Object.keys(styleIds)[0];
```
LiamMartens, line 9: I'm wondering if this should support the removal of multiple tokens? The interface kind of makes this a confusing behavior if it only picks the first one

LiamMartens (changes requested):

src/app/store/models/reducers/tokenState/removeStyleIdsFromCurrentTheme.ts
six7: Yeah let's support deleting multiple. It could be users have multiple tokens created that are actually named the same, so lets just iterate over all of them to delete them?

src/app/store/useManageTokens.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+jest.mock('./useTokens', () => ({
+  __esModule: true,
+  default: () => ({
+    removeStylesFromTokens: mockRemoveStylesFromTokens,
+  }),
+}));
+
+describe('useManageTokens', () => {
+  it('should be able to remove styles from the currentTheme when the user select option', async () => {
```
six7, line 24: Should be able to remove styles from any theme if the user deleted one

src/plugin/removeStylesFromPlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+  const stylePathSlice = settings?.ignoreFirstPartForStyles ? 1 : 0;
+  const stylePathPrefix = settings?.prefixStylesWithThemeName && activeThemeObject
+    ? activeThemeObject.name
+    : null;
+
+  const pathname = convertTokenNameToPath(token.path, stylePathPrefix, stylePathSlice);
+
+  allStyles.forEach((style) => {
+    if (isMatchingStyle(token.path, pathname, activeThemeObject?.$figmaStyleReferences ?? {}, figmaStyleMaps)) {
+      style.remove();
```
six7, line 34: In here we'd also need to pass back that styleId so we can delete those connected figmaStyleReferenced

six7 (changes requested):

six7 (changes requested):
This does not work yet.. it deletes all styles instead of deleting just the style that should be deleted. happens both when themes are present and when they're not present

https://user-images.githubusercontent.com/4548309/189213577-19342687-d69e-483f-b562-1f3f998e60a9.mp4



could you make sure to include a short video in the PRs once they're ready for review? then we could give feedback like that quicker 🙏

swordEdge:
> This does not work yet.. it deletes all styles instead of deleting just the style that should be deleted. happens both when themes are present and when they're not present
> 
>  CleanShot.2022-09-08.at.21.50.03.mp4 
> could you make sure to include a short video in the PRs once they're ready for review? then we could give feedback like that quicker 🙏


@six7  Checkout this video. 
https://www.loom.com/share/6fb659e27cd8446c925fbd38eb7bc940

LiamMartens:
@swordEdge what is the current status here?

swordEdge:
@swordEdge Hi.

LiamMartens (approved):

PR #1243, opened 2022-09-02
1194 when renaming a token ask to rename any local styles in addition to the remap tokens functionality we have

swordEdge:
When we rename a token, we can update the styles in the document and activeTheme either.

Check this video.
https://www.loom.com/share/3be8399dee5a4277a080d4a618214042


LiamMartens:
Also coverage :grimacing:

src/app/store/models/reducers/tokenState/renameStyleIdsToCurrentTheme.ts
```diff
@@ -0,0 +1,26 @@
+import type { TokenState } from '../../tokenState';
+
+export function renameStyleIdsToCurrentTheme(state: TokenState, oldName: string, newName: string): TokenState {
+  // ignore if there is no active theme
+  if (!state.activeTheme) return state;
```
six7, line 5: For this PR the same feedback as for the other: it could be the user doesnt have that theme open but deletes a token - which deletes a style - that might be connected to a figmaStyleReference of an inactive theme

six7 (changes requested):

src/app/store/models/reducers/tokenState/renameStyleIdsToCurrentTheme.ts
swordEdge: @six7 I'll complete this pr after 1195 has been merged.

src/app/store/models/reducers/__tests__/renameStyleIdsToTheme.test.ts
```diff
@@ -0,0 +1,75 @@
...
+import { RootModel } from '@/types/RootModel';
+import { models } from '../../index';
+
+type Store = RematchStore<RootModel, Record<string, never>>;
+
+describe('renameStyleIdsToCurrentTheme', () => {
+  let store: Store;
+
+  it('should rename styleId to current theme', async () => {
+    store = init<RootModel>({
```
LiamMartens, line 11: btw you can use `createMockStore` instead

LiamMartens (approved):

LiamMartens:
@swordEdge could you solve the conflicts?

PR #1271, opened 2022-09-09
1270 when renaming a token group the reference to the token is lost

swordEdge:
- automatically change references in other tokens pointing to this group (with this type): like if another token was using {colors.red.500} and i rename the group `colors` to `palette` then we change references to `palette.red.500` etc
- and then ask the user to remap their tokens (the modal that appears when you rename a token)]

@six7 Update the video
https://www.loom.com/share/f735c80667244452a259f8d2c6c4d1b1


src/app/components/TokenGroup/TokenGroupHeading.tsx
```diff
@@ -50,11 +56,31 @@ export function TokenGroupHeading({
...
   }, []);
 
-  const handleRenameTokenGroupSubmit = React.useCallback((e: React.FormEvent<HTMLFormElement>) => {
+  const handleRenameTokenGroupSubmit = React.useCallback(async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
 
     setShowNewGroupNameField(false);
     renameGroup(`${path}${copyName}`, `${newTokenGroupName}`, type);
+
+    const shouldRemap = await confirm({
```
six7, line 65: Can we move this to a hook? That would keep this component cleaner.

six7 (changes requested):
Thanks for adding the video!

Can you make sure this also works `n` levels deep? Right now the implementation (I think) only works on the first level?

For example, this doesn't work:

`some.level.deep` -> `abc.level.deep` --> this results in references not being changed (the bulk remap works! but the references inside the plugin don't).

six7:
Also, let's add tests for the added functionality (inside the hook it will be easier to test it in isolation)

six7 (approved):
🎉

PR #1280, opened 2022-09-13
1275-removing-token-sets-sometimes-fails

swordEdge:
Sometimes when using Git storage removing a token set fails. It seems to be connected to the name of the set, as it's not happening all the time.

https://www.loom.com/share/a383c5d6b04e42b1aede02bd60f91846

### DIFF
--- a/src/storage/GithubTokenStorage.ts
+++ b/src/storage/GithubTokenStorage.ts
@@ -10,6 +10,7 @@ import {
 } from './GitTokenStorage';
 import { SystemFilenames } from '@/constants/SystemFilenames';
 import { ErrorMessages } from '@/constants/ErrorMessages';
+import { joinPath } from '@/utils/string';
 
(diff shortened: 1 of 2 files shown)


LiamMartens (approved):

six7 (changes requested):
This seems to be removing files that I did not remove.

In this commit I only deleted the file `Colors/Archive.json` - but `Base/Colors/Archive.json` was deleted as well.

Can we add tests for this please?

swordEdge:
> In this commit I only deleted the file `Colors/Archive.json` - but `Base/Colors/Archive.json` was deleted as well.
@six7 In my computer it works properly. Could you check this video and again in your computer.
https://www.loom.com/share/a383c5d6b04e42b1aede02bd60f91846

six7:
Great, works now! Merging in

PR #1281, opened 2022-09-13
1278-120-rc1-renaming-token-groups-causes-them-to-collapse

swordEdge:
Renaming a token group seems to have a strange side-effect: It collapses the token group while renaming.
Expected: Don't collapse groups while renaming
https://www.loom.com/share/eeb6db0d63884bd6b5e1a0be6ad1457e


src/app/components/TokenGroup/TokenGroupHeading.tsx
```diff
@@ -79,6 +79,7 @@ export function TokenGroupHeading({
   }, [duplicateGroup, path, type]);
 
   const handleToggleCollapsed = useCallback(() => {
+    console.log('collapse');
```
LiamMartens, line 82: log statement

LiamMartens (changes requested):

six7 (approved):

PR #1319, opened 2022-10-04
1316-tokens-not-applying

swordEdge:
https://user-images.githubusercontent.com/25951419/193707486-7c8bdfca-8b52-4107-8c97-4e989b32943d.mp4

### DIFF
--- a/src/utils/trySetStyleId.ts
+++ b/src/utils/trySetStyleId.ts
@@ -12,42 +12,34 @@ export async function trySetStyleId(node: BaseNode, type: StyleType, styleId: st
         .catch(() => resolve(styleId));
     });
   }
+  try {
+    if (type === 'fill' && 'fillStyleId' in node) {
+      node.fillStyleId = actualStyleId;
+      return (
+        node.fillStyleId === actualStyleId || (styleKeyMatch && styleKeyMatch[0] === node.fillStyleId)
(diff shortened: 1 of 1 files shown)


src/utils/trySetStyleId.ts
```diff
@@ -12,42 +12,34 @@ export async function trySetStyleId(node: BaseNode, type: StyleType, styleId: st
...
-    node.textStyleId = actualStyleId;
-    return (
-      node.textStyleId === actualStyleId
-      || (styleKeyMatch && styleKeyMatch[0] === node.textStyleId)
-    );
+    if (type === 'effect' && 'effectStyleId' in node) {
+      node.effectStyleId = actualStyleId;
+      return node.effectStyleId === actualStyleId || (styleKeyMatch && styleKeyMatch[0] === node.effectStyleId);
+    }
+  } catch (e) {
```
LiamMartens, line 41: Just something I'm wondering but what kind of error was being thrown here?

LiamMartens (approved):

src/utils/trySetStyleId.ts
six7: The error that was thrown is `message: "in set_fillStyleId: Cannot set style successfully: Cannot set deleted style"`. I guess our logic could be improved by not trying to set a deleted style in a way. This PR however fixes it for now as it then just sets the hex.

PR #1325, opened 2022-10-06
support font familty fall back

swordEdge:
https://user-images.githubusercontent.com/25951419/194247298-38830679-404e-4808-ba58-5f9d97063069.mp4

### DIFF
--- a/src/plugin/setTextValuesOnTarget.ts
+++ b/src/plugin/setTextValuesOnTarget.ts
@@ -2,7 +2,10 @@
 import { SingleTypographyToken } from '@/types/tokens';
 import { transformValue } from './helpers';
 
-export default async function setTextValuesOnTarget(target: TextNode | TextStyle, token: Pick<SingleTypographyToken, 'value' | 'description'>) {
+export default async function setTextValuesOnTarget(
+  target: TextNode | TextStyle,
+  token: Pick<SingleTypographyToken, 'value' | 'description'>,
+) {
   try {
     const { value, description } = token;
     if (typeof value !== 'string') {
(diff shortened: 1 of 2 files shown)


src/plugin/setTextValuesOnTarget.ts
```diff
@@ -17,7 +17,9 @@ export default async function setTextValuesOnTarget(target: TextNode | TextStyle
         textCase,
         textDecoration,
       } = value;
-      const family = fontFamily?.toString() || (target.fontName !== figma.mixed ? target.fontName.family : '');
+      const splitFontFamily = fontFamily?.split(',')[0];
```
six7, line 20: The idea is that if the first font cant be applied we keep trying the others as well, one by one

six7 (changes requested):

src/plugin/setTextValuesOnTarget.ts
```diff
@@ -28,24 +31,46 @@ export default async function setTextValuesOnTarget(target: TextNode | TextStyle
...
+              if (candidateFonts[i]) {
+                target.fontName = {
+                  family: candidateFonts[i].family,
+                  style: candidateFonts[i].style,
+                };
+                isApplied = true;
+              }
+            })
+            .catch((e) => {
+              console.log('Error setting fontWeight on target', e);
```
six7, line 69:
```suggestion
```

I removed this on purpose on `main` because users said it was annoying as it spammed the console.

six7 (changes requested):

six7 (approved):

PR #1327, opened 2022-10-07
1261 introduce new start screen with helpful links

MightyDevKnight:
ScreenShot_20221007152411
ScreenShot_20221007152653

### DIFF
--- a/src/app/components/Callout.tsx
+++ b/src/app/components/Callout.tsx
@@ -25,7 +25,7 @@ export default function Callout({
   heading, description, action, id,
 }: Props) {
   return (
-    <Box css={{ backgroundColor: '$bgSubtle', padding: '$4', borderRadius: '$default' }}>
+    <Box css={{ backgroundColor: '$startScreenAlertBg', padding: '$4', borderRadius: '$default' }}>
       <Stack direction="row" gap={2}>
         <Box css={{ color: '$interaction' }}>
           <IconBell />
--- a/src/app/components/Callout.tsx
+++ b/src/app/components/Callout.tsx
@@ -1,15 +1,10 @@
 import React from 'react';
-import { styled } from '@/stitches.config';
+import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
 import Heading from './Heading';
-import IconBell from '@/icons/bell.svg';
 import Stack from './Stack';
 import Text from './Text';
 import Box from './Box';
-
-const StyledButton = styled('button', {
-  fontSize: '$xsmall',
-  color: '$interaction',
-});
+import Button from './Button';
 
 type Props = {
   id: string
@@ -17,7 +12,7 @@ type Props = {
   description: React.ReactNode
   action: {
     text: React.ReactNode
-    onClick: React.ComponentProps<typeof StyledButton>['onClick']
+    onClick: React.ComponentProps<typeof Button>['onClick']
   }
 };
 
@@ -27,15 +22,15 @@ export default function Callout({
   return (
     <Box css={{ backgroundColor: '$startScreenAlertBg', padding: '$4', borderRadius: '$default' }}>
       <Stack direction="row" gap={2}>
-        <Box css={{ color: '$interaction' }}>
-          <IconBell />
+        <Box css={{ color: '$dangerBg' }}>
+          <ExclamationTriangleIcon />
         </Box>
         <Stack align="start" direction="column" gap={2}>
           <Heading>{heading}</Heading>
           <Text size="xsmall">{description}</Text>
-          <StyledButton data-cy={id} type="button" onClick={action.onClick}>
+          <Button data-cy={id} id="button-configure" size="small" variant="primary" onClick={action.onClick}>
             {action.text}
-          </StyledButton>
+          </Button>
         </Stack>
       </Stack>
     </Box>
--- a/src/app/components/Callout.tsx
+++ b/src/app/components/Callout.tsx
@@ -21,8 +21,8 @@ export default function Callout({
 }: Props) {
   return (
     <Box css={{ backgroundColor: '$startScreenAlertBg', padding: '$4', borderRadius: '$default' }}>
-      <Stack direction="row" gap={2}>
-        <Box css={{ color: '$dangerBg' }}>
+      <Stack direction="row" gap={4}>
+        <Box css={{ color: '$dangerBg', marginTop: '$3' }}>
           <ExclamationTriangleIcon />
         </Box>
         <Stack align="start" direction="column" gap={2}>
--- a/src/app/components/StartScreen.tsx
+++ b/src/app/components/StartScreen.tsx
@@ -57,7 +57,7 @@ function StartScreen() {
 
   return (
     <div className="h-auto p-4 my-auto content scroll-container">
-      <Stack direction="column" gap={4} css={{ display: 'flex', padding: '2rem', background: '$startScreenBg' }}>
+      <Stack direction="column" gap={4} css={{ padding: '2rem', background: '$startScreenBg' }}>
         <Stack direction="row" gap={2}>
           <StyledFigmaMark />
           <StyledFigmaLetter />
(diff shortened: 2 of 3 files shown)


LiamMartens (approved):

cypress/integration/startup.spec.js
```diff
@@ -64,6 +64,6 @@ describe('Loads application', () => {
   it('shows welcome page when no token values are given', () => {
     cy.visit('/');
     cy.startup(mockStartupParams);
-    cy.get('p').contains('Welcome to Figma Tokens');
+    cy.get('a').contains("Found an issue We're on GitHub!");
```
six7, line 67:
```suggestion
    cy.get('a').contains("Found an issue? We're on GitHub!");
```

src/app/components/Callout.tsx
```diff
@@ -1,41 +1,36 @@
...
   heading, description, action, id,
 }: Props) {
   return (
-    <Box css={{ backgroundColor: '$bgSubtle', padding: '$4', borderRadius: '$default' }}>
-      <Stack direction="row" gap={2}>
-        <Box css={{ color: '$interaction' }}>
-          <IconBell />
+    <Box css={{ backgroundColor: '$startScreenAlertBg', padding: '$4', borderRadius: '$default' }}>
+      <Stack direction="row" gap={4}>
+        <Box css={{ color: '$dangerBg', marginTop: '$3' }}>
```
six7, line 25: You're applying a token to the foreground color of an object, applying a Background token for that is not good.

```suggestion
        <Box css={{ color: '$dangerFg', marginTop: '$3' }}>
```

src/app/components/Callout.tsx
```diff
@@ -1,41 +1,36 @@
...
+          <ExclamationTriangleIcon />
         </Box>
         <Stack align="start" direction="column" gap={2}>
           <Heading>{heading}</Heading>
           <Text size="xsmall">{description}</Text>
-          <StyledButton data-cy={id} type="button" onClick={action.onClick}>
+          <Button data-cy={id} id="button-configure" size="small" variant="primary" onClick={action.onClick}>
```
six7, line 31: This component can be used outside Start screen as well, let's not bind it to it. The ID should be dynamic.

src/app/components/StartScreen.tsx
```diff
@@ -1,15 +1,40 @@
...
+  width: '90px',
+  height: '55px',
+});
+
+const StyledFigmaMark = styled(FigmaMark, {
+  width: '55px',
+  height: '55px',
+});
+
+const HelpfulLinks = styled('a', {
```
six7, line 29: I would call this singular HelpfulLink or so.

src/app/components/StartScreen.tsx
```diff
@@ -1,15 +1,40 @@
...
+  color: '$textMuted',
+  fontSize: '$xsmall',
+  '&:hover, &:focus': {
+    color: '$onInteraction',
```
six7, line 33: onInteraction should not be used here, you can use $text instead.

onInteraction is the 'white' text on 'blue' (interaction) background. If you apply it on a regular background it would mean the text in Figma Light mode is `white` on `white`, so users cant see it.

src/app/components/StartScreen.tsx
```diff
@@ -32,40 +57,64 @@ function StartScreen() {
...
+        {storageType?.provider !== StorageProviderType.LOCAL ? (
           <Callout
             id="callout-action-setupsync"
             heading="Remote storage detected"
-            description={`This document is setup with a remote token source on ${storageType.provider}. Ask your team for the credentials, then enter them in the Sync dialog.`}
+            description="This document was setup with a remote storage. Ask your team for the credentials, then enter them in the Sync dialog."
             action={{
               onClick: onSetSyncClick,
-              text: 'Set up sync',
+              text: 'Enter credential',
```
six7, line 102:
```suggestion
              text: 'Enter credentials',
```

src/app/components/StartScreen.tsx
```diff
@@ -32,40 +57,64 @@ function StartScreen() {
...
+          <Stack direction="row" gap={2}>
+            <Button id="button-configure" size="small" variant="primary" onClick={onSetDefaultTokens}>
+              Get started with a new file
+            </Button>
+          </Stack>
         )}
+        <Stack direction="row" gap={2}>
+          <GitHubLogoIcon />
+          <a href="https://github.com/six7/figma-tokens" style={{ color: '$textMuted', fontSize: '$xsmall' }} target="_blank" rel="noreferrer" className="underline">
+            Found an issue We&#39;re on GitHub!
```
six7, line 115:
```suggestion
            Found an issue? We&#39;re on GitHub!
```

src/stitches.config.ts
```diff
@@ -59,6 +59,8 @@ export const stitchesInstance = createStitches({
       loadingScreenFg: '#FFFFFF',
       loadingScreenFgMuted: '#CECECE',
       loadingScreenBg: '#141414',
+      startScreenBg: '#363636',
```
six7, line 62: as mentioned above these should be called differently not tied to the startScreen

src/app/components/StartScreen.tsx
```diff
@@ -32,40 +57,64 @@ function StartScreen() {
...
   return (
     <div className="h-auto p-4 my-auto content scroll-container">
-      <Stack direction="column" gap={4}>
-        <a href="https://jansix.at/resources/figma-tokens?ref=figma-tokens-plugin" target="_blank" rel="noreferrer">
-          {/* eslint-disable-next-line */}
-          <img alt="Figma Tokens Splashscreen" src={require('../assets/tokens-intro.jpg')} className="rounded" />
-        </a>
+      <Stack direction="column" gap={4} css={{ padding: '2rem', background: '$startScreenBg' }}>
+        <Stack direction="row" gap={2}>
+          <StyledFigmaMark />
```
six7, line 62: Can you call this 'FigmaTokensLogoMark' and 'FigmaTokensLogo'? This is not the Figma Mark but our logo.

src/app/components/StartScreen.tsx
```diff
@@ -1,15 +1,40 @@
...
+  height: '55px',
+});
+
+const HelpfulLinks = styled('a', {
+  color: '$textMuted',
+  fontSize: '$xsmall',
+  '&:hover, &:focus': {
+    color: '$onInteraction',
+  },
+  rel: 'noreferrer',
```
six7, line 35: This is a html property, not a css one.
```suggestion
```

src/app/components/StartScreen.tsx
```diff
@@ -32,40 +57,64 @@ function StartScreen() {
 
   return (
     <div className="h-auto p-4 my-auto content scroll-container">
-      <Stack direction="column" gap={4}>
-        <a href="https://jansix.at/resources/figma-tokens?ref=figma-tokens-plugin" target="_blank" rel="noreferrer">
-          {/* eslint-disable-next-line */}
-          <img alt="Figma Tokens Splashscreen" src={require('../assets/tokens-intro.jpg')} className="rounded" />
-        </a>
+      <Stack direction="column" gap={4} css={{ padding: '2rem', background: '$startScreenBg' }}>
```
six7, line 60:
```suggestion
      <Stack direction="column" gap={4} css={{ padding: '2rem', background: '$bgSubtle', justifyContent: 'center' }}>
```

This is what it looked like for me in Light mode. 
image

src/app/components/StartScreen.tsx
six7: Here's what it looks like right now in light mode: 
image

src/app/components/StartScreen.tsx
```diff
@@ -32,40 +57,64 @@ function StartScreen() {
 
   return (
     <div className="h-auto p-4 my-auto content scroll-container">
```
six7, line 59:
```suggestion
    <div className="h-auto p-4 content scroll-container">
```

six7: Can you also change this container to have

```
height: 100%
display: flex
```

src/app/components/StartScreen.tsx
```diff
@@ -32,40 +57,64 @@ function StartScreen() {
```
six7, line 57: Can you also fix that on line 49 we currently have
`dispatch.uiState.setActiveTab(Tabs.TOKENS)`
this should be
`dispatch.uiState.setActiveTab(Tabs.SETTINGS)`

Right now when you click Enter credentials you stay on start screen but it should take you to settings. which the above suggestion does

six7 (changes requested):

six7:
@MightyDevKnight can you link the issue?

six7 (approved):
Nice 🎉

PR #1328, opened 2022-10-07
1293 changes to license key display

MightyDevKnight:
ScreenShot_20221007172703


six7:
@MightyDevKnight can you link the issue?

also, let's not show the `Update license key` when a license key is already entered. Let's make the user remove their existing key first, and then they can add a new one.

PR #1330, opened 2022-10-08
Changes bulk remap to selection and includes

six7:
This PR changes bulk remap to no longer scan the currently selected updateMode but instead always takes selection (because that's what should happen in Inspect)

Also changes implementation so the `match` doesn't just have to startWith the word, but can be anywhere.

### DIFF
--- a/src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
+++ b/src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
@@ -1,4 +1,5 @@
 import { AsyncMessageChannelHandlers } from '@/AsyncMessageChannel';
+import { UpdateMode } from '@/constants/UpdateMode';
 import { AsyncMessageTypes } from '@/types/AsyncMessages';
 import { defaultNodeManager, NodeManagerNode } from '../NodeManager';
 import { updatePluginData } from '../pluginData';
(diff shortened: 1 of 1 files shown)


src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -7,13 +8,13 @@ import { sendSelectionChange } from '../sendSelectionChange';
...
     const { oldName, newName } = msg;
-    const allWithData = await defaultNodeManager.findNodesWithData({});
+    const allWithData = await defaultNodeManager.findNodesWithData({ updateMode: UpdateMode.SELECTION });
     const updatedNodes: NodeManagerNode[] = [];
 
     allWithData.forEach((node) => {
       const { tokens } = node;
       const updatedTokens = Object.entries(tokens).reduce<Record<string, string>>((acc, [key, val]) => {
-        if (val.startsWith(oldName)) {
+        if (val.includes(oldName)) {
```
LiamMartens, line 17: Will this "over" match?

LiamMartens (approved):
@six7 just a question

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
six7: It will find **any** occurrence of the oldName. If it is found, it will replace the first appearance of it.

This is how Figma's bulk rename is designed and what Bulk remap is modeled after (in a simple form).

Previously it only matched on startsWith, but some users found the function confusing as it didn't trigger for them (even though they expected it to)

PR #1339, opened 2022-10-12
1200 introduce onboarding modal flow

MightyDevKnight:
ScreenShot_20221012211313
ScreenShot_20221012211318
ScreenShot_20221012211325
ScreenShot_20221012211334


src/app/components/OnboardingFlow.tsx
```diff
@@ -0,0 +1,123 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import Heading from './Heading';
+import Button from './Button';
+import Modal from './Modal';
+import { getLastopened } from '@/selectors';
+import Stack from './Stack';
+
+export default function OnboardingFlow() {
+  const FALLBACK_LAST_OPENED = 1616241985291;
```
six7, line 10: Instead of using a date here I would like this to work like this:
- If there is no "lastOpened" flag set on the client, show the onboarding flow.
- After opening the plugin once, we set lastOpened
- Any subsequent launch would no longer show the onboardingModal because there is a `lastOpened` set.

That way we do not need this fallback_last_opened and it will be easy for us to reset this later - by just removing `lastOpened`.

src/app/components/OnboardingFlow.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const onboardingflow = [
+    {
+      image: {
+        alt: '',
+        filename: 'https://user-images.githubusercontent.com/4548309/190957841-bc189fb7-0351-4206-9e38-a4b85506d300.png',
```
six7, line 15: Can you add these images to our local assets?

src/app/components/OnboardingFlow.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }, []);
+
+  return (
+    <Modal showClose isOpen={lastOpened === FALLBACK_LAST_OPENED && onboardingFlowOpen} close={handleClose}>
+      <Stack direction="column" gap={4}>
+        <Heading size="medium">Get Started</Heading>
+        <div>
+          {onboardingflow.map((item, index) => (
+            <Stack
+              // eslint-disable-next-line no-underscore-dangle
```
six7, line 82: is this needed? i dont see no underscore

src/utils/storyblok.ts
```diff
@@ -1,5 +1,7 @@
 import { StoryblokStory } from '@/types';
 
+const FALLBACK_LAST_OPENED = 1616241985291;
```
six7, line 3: same as above

six7 (changes requested):
I'd like to move away from the FALLBACK_LAST_OPENED both in the code you changed here as well as in src/utils/getLastOpened.ts -> if the user has never opened the plugin, the property should not be set - and we can show the onboarding modal

six7:
@MightyDevKnight can we get tests to pass & coverage?

src/app/components/AppContainer/__tests__/AppContainerIntegration.test.tsx
```diff
@@ -199,6 +199,7 @@ describe('AppContainer (integration)', () => {
       withOrWithoutLicense({
         ...mockStartupParams,
         localTokenData: null,
+        lastOpened: Date.now(),
```
LiamMartens, line 202: I guess this should not be a date anymore? Just for consistency

LiamMartens (changes requested):
@six7 @MightyDevKnight just 1 minor consistency change - but separately, it's probably worth converting those into WebPs since they are anywhere between 75 and 50% smaller overall

src/app/components/AppContainer/__tests__/AppContainerIntegration.test.tsx
six7: i think it should still be a date? like we store Date.now() when you open the plugin. but if you never opened it, it will be at 0.

six7 (approved):

PR #1341, opened 2022-10-12
1266 add border token type

swordEdge:
https://www.loom.com/share/824690fd6fd542a9923195670e6ddf8a


six7 (changes requested):
- Can we move the `Border` type from the order to the other `Border width` and `Border radius` tokens so they appear in the UI after each other?

other than that this is great! haven't looked at code though!

six7:
+ coverage please 🙏

LiamMartens (approved):

PR #1342, opened 2022-10-13
1201 introduce dimension type

swordEdge:
https://www.loom.com/share/5475c9670f01492e9d97aafaea106834

### DIFF
--- a/src/app/components/EditTokenForm.tsx
+++ b/src/app/components/EditTokenForm.tsx
@@ -8,8 +8,10 @@ import Input from './Input';
 import ColorPicker from './ColorPicker';
 import useConfirm from '../hooks/useConfirm';
 import useTokens from '../store/useTokens';
-import { EditTokenObject, SingleBoxShadowToken, SingleToken } from '@/types/tokens';
-import { checkIfContainsAlias, getAliasValue } from '@/utils/alias';
+import {
+  EditTokenObject, SingleBoxShadowToken, SingleDimensionToken, SingleToken,
+} from '@/types/tokens';
+import { checkIfAlias, checkIfContainsAlias, getAliasValue } from '@/utils/alias';
 import { ResolveTokenValuesResult } from '@/plugin/tokenHelpers';
 import { activeTokenSetSelector, updateModeSelector, editTokenSelector } from '@/selectors';
 import { TokenTypes } from '@/constants/TokenTypes';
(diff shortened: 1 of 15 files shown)


six7 (changes requested):
This works great for dimension. However I noticed an issue with other token types:


## Spacing, sizing, radius, borderWidth tokens seem to also have validation

I did notice an issue with old tokens that were of type `spacing`, `sizing`, `borderRadius`, `borderWidth`.

They also seem to have gotten that validation that we introduced (Must include either px or rem).

To be clear that should not apply to them, they can stay `14`, not `14px`.

CleanShot 2022-10-14 at 09 11 19

## Validation only performs on enter
It seems validation of the token value only performs when users try to submit the value.
We can already do that on `blur`, so when the user leaves the focus of the `value` input we already give that feedback

## Add borderRadius to the contextMenu when right-clicking dimension tokens

One thing that we forgot during shaping this issue:

- Can you also add `borderRadius` in the dropdown where you can choose options?

CleanShot 2022-10-14 at 09 10 21

Border radius
- All
- Top Left
- Top Right
- Bottom Right
- Bottom Left

### Dimension type should be suggested for `spacing`, `sizing`, `borderRadius`, `borderWidth` tokens

We should include tokens of type `dimension` in the dropdown that appears when you're on either of the mentioned tokens, so you can use it there.

src/plugin/setValuesOnNode.ts
```diff
@@ -28,6 +28,7 @@ export default async function setValuesOnNode(
   ignoreFirstPartForStyles = false,
   prefixStylesWithThemeName = false,
 ) {
+  console.log('value', values);
```
LiamMartens, line 31: log

LiamMartens (changes requested):

src/app/components/EditTokenForm.tsx
```diff
@@ -44,11 +46,16 @@ function EditTokenForm({ resolvedTokens }: Props) {
   const [internalEditToken, setInternalEditToken] = React.useState<typeof editToken>(editToken);
   const { confirm } = useConfirm();
 
+  const isValidDimensionToken = React.useMemo(() => internalEditToken.type === TokenTypes.DIMENSION && (internalEditToken.value?.endsWith('px') || internalEditToken.value?.endsWith('rem') || checkIfAlias(internalEditToken as SingleDimensionToken, resolvedTokens)), [internalEditToken, resolvedTokens, checkIfAlias]);
```
LiamMartens, line 49: @six7 what about other values like `em`, `%`, etc... should this check just be a regex allowing something like `/[0-9]+[^\d]+/`

LiamMartens (approved):
@six7 just 1 q

src/app/components/EditTokenForm.tsx
six7: The format only allows px or rem for now so I think it's ok for now! We might revisit this when the format changes though.

> The value must be a string containing a number (either integer or floating-point) followed by either a "px" or "rem" unit (future spec iterations may add support for additional units).

https://tr.designtokens.org/format/#dimension

LiamMartens:
@swordEdge just some conflicts - all good aside from that

LiamMartens (approved):

six7 (approved):

PR #1355, opened 2022-10-17
1262 introduce in product explainers

MightyDevKnight:
ScreenShot_20221017153832
ScreenShot_20221017153817
ScreenShot_20221017153825


src/app/components/InspectorMultiView.tsx
```diff
@@ -78,6 +82,10 @@ export default function InspectorMultiView({ resolvedTokens }: { resolvedTokens:
     );
   }, [dispatch.inspectState, inspectState.selectedTokens.length, uiState.selectionValues]);
 
+  const closeOnboarding = React.useCallback(async () => {
+    dispatch.uiState.setOnboardingExplainerInspect('false');
```
LiamMartens, line 86: Why is this a string value of `false` and not an actual boolean?

src/app/components/InspectorMultiView.tsx
```diff
@@ -123,7 +131,34 @@ export default function InspectorMultiView({ resolvedTokens }: { resolvedTokens:
...
+                <IconButton onClick={closeOnboarding} icon={<Cross1Icon />} />
+              </Stack>
+              <p className="text-xs">
+                This is where applied tokens of your selection show up, you can use Deep Inspect to scan the selected layers and all of its children.
+              </p>
+              <a
+                target="_blank"
+                rel="noreferrer"
+                href="https://docs.figmatokens.com/multi-inspect?ref=onboarding_explainer_inspect"
+                className="inline-flex text-xs text-primary-500"
```
LiamMartens, line 155: Can we use stitches here? Otherwise we need to replace this with stitches later again so there's no point

src/app/components/InspectorMultiView.tsx
```diff
@@ -123,7 +131,34 @@ export default function InspectorMultiView({ resolvedTokens }: { resolvedTokens:
...
+            }}
+            >
+              <Stack direction="row" gap={2} justify="between">
+                <Stack direction="row" justify="between" gap={2} align="center">
+                  <InfoCircledIcon className="text-primary-500" />
+                  <Heading size="medium">Inspect</Heading>
+                </Stack>
+                <IconButton onClick={closeOnboarding} icon={<Cross1Icon />} />
+              </Stack>
+              <p className="text-xs">
```
LiamMartens, line 148: Also replace with stitches styling

src/app/components/Settings/Settings.tsx
```diff
@@ -36,11 +39,42 @@ function Settings() {
...
+                <IconButton dataCy="closeButton" onClick={closeOnboarding} icon={<Cross1Icon />} />
+              </Stack>
+              <p className="text-xs">
+                Connect your tokens to an external source of truth that you can push and pull to. This allows you to use tokens across files.
+              </p>
+              <a
+                target="_blank"
+                rel="noreferrer"
+                href="https://docs.figmatokens.com/sync/sync?ref=onboarding_explainer_syncproviders"
+                className="inline-flex text-xs text-primary-500"
```
LiamMartens, line 71: Same comment, let's use stitches since we are removing tailwind

src/app/components/TokenSetSelector.tsx
```diff
@@ -211,6 +218,34 @@ export default function TokenSetSelector({ saveScrollPositionSet } : { saveScrol
         New set
         <IconAdd />
       </StyledButton>
+      {uiState.onboardingExplainerSets === 'true' && (
+        <Box css={{
+          display: 'flex', flexDirection: 'column', gap: '$2', padding: '$4', borderBottom: '1px solid $borderMuted', borderTop: '1px solid $borderMuted',
+        }}
+        >
+          <Stack direction="row" gap={2} justify="between">
```
LiamMartens, line 226: I feel like this block has been repeated a few times now, can we split this into it's own reusable component and use that instead. Otherwise we have a bunch of duplicate code to maintain and refactor later.

src/figmaStorage/OnboardingExplainerInspectProperty.ts
```diff
@@ -0,0 +1,6 @@
+import { FigmaStorageProperty, FigmaStorageType } from './FigmaStorageProperty';
+
+export const OnboardingExplainerInspectProperty = new FigmaStorageProperty<string>(
+  FigmaStorageType.CLIENT_STORAGE,
+  'onboardingExplainerInspect',
```
LiamMartens, line 5: You can add a parser to this to transform the string into boolean and vice versa. This would be better for the DX in React - since doing `=== 'false'` is not ideal.

LiamMartens (changes requested):

src/app/components/InspectorMultiView.tsx
MightyDevKnight: I tried to use a boolean, but when I save the false value, the onboardingflag saved as null.

LiamMartens (approved):

src/app/components/OnboardingExplainer.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export default function OnboardingExplainer({ data, closeOnboarding }: Props) {
+  return (
+    <Box css={{
+      display: 'flex', flexDirection: 'column', gap: '$2', padding: '$4', border: '1px solid $borderMuted', borderTop: '1px solid $borderMuted',
+    }}
+    >
+      <Stack direction="row" gap={2} justify="between">
+        <Stack direction="row" justify="between" gap={2} align="center">
+          <InfoCircledIcon className="text-primary-500" />
```
six7, line 25: Can you use Stitches for this? We want to move away from tailwind

src/app/components/OnboardingExplainer.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <p className="text-xs">
+          {text}
+          <br />
+        </p>
+      ))}
+      <a
+        target="_blank"
+        rel="noreferrer"
+        href={data.url}
+        className="inline-flex text-xs text-primary-500"
```
six7, line 40: Same here

src/app/components/OnboardingExplainer.tsx
```diff
@@ -0,0 +1,46 @@
...
+    >
+      <Stack direction="row" gap={2} justify="between">
+        <Stack direction="row" justify="between" gap={2} align="center">
+          <InfoCircledIcon className="text-primary-500" />
+          <Heading size="medium">{data.title}</Heading>
+        </Stack>
+        <IconButton dataCy="closeButton" onClick={closeOnboarding} icon={<Cross1Icon />} />
+      </Stack>
+      {data.text.split('\n').map((text) => (
+        <p className="text-xs">
```
six7, line 31: Same here

src/utils/getOnboardingExplainerInspect.ts
```diff
@@ -0,0 +1,23 @@
+import { OnboardingExplainerInspectProperty, LastOpenedProperty } from '@/figmaStorage';
+
+export default async function getOnboardingExplainerInspect() {
```
six7, line 3: Can these individual files not be combined into one? It seems to me we now have 3x the logic even though they're all doing the same thing just with a different key? Feels like it could be optimized to reduce code?

src/utils/plugin/startup.ts
```diff
@@ -14,6 +17,9 @@ export async function startup() {
     activeTheme,
     userId, // the user ID is used for license key binding
     lastOpened,
+    onboardingExplainerSets,
```
six7, line 20: For example here it would be ideal if we didnt have 3 calls but just one to fetch all these?

six7 (changes requested):

src/app/components/OnboardingExplainer.tsx
```diff
@@ -1,10 +1,26 @@
...
 import Heading from './Heading';
 import Stack from './Stack';
 import IconButton from './IconButton';
 
+const StyledInfoIconButton = styled(InfoCircledIcon, {
+  color: '$interactionSubtle',
+});
+
+const StyledTextPlan = styled('p', {
+  fontWeight: '$medium',
```
six7, line 14:
```suggestion
  fontWeight: '$normal',
```

src/app/components/OnboardingExplainer.tsx
```diff
@@ -22,25 +38,20 @@ export default function OnboardingExplainer({ data, closeOnboarding }: Props) {
...
+          <StyledInfoIconButton />
           <Heading size="medium">{data.title}</Heading>
         </Stack>
         <IconButton dataCy="closeButton" onClick={closeOnboarding} icon={<Cross1Icon />} />
       </Stack>
       {data.text.split('\n').map((text) => (
-        <p className="text-xs">
+        <StyledTextPlan>
           {text}
           <br />
```
six7, line 49:
```suggestion
```

not needed

src/app/components/OnboardingExplainer.tsx
```diff
@@ -22,25 +38,20 @@ export default function OnboardingExplainer({ data, closeOnboarding }: Props) {
     >
       <Stack direction="row" gap={2} justify="between">
         <Stack direction="row" justify="between" gap={2} align="center">
-          <InfoCircledIcon className="text-primary-500" />
+          <StyledInfoIconButton />
           <Heading size="medium">{data.title}</Heading>
```
six7, line 42:
```suggestion
          <Heading size="small">{data.title}</Heading>
```

six7 (changes requested):
When the onboarding is placed in the left section the border should not be on all sides, instead it should be only borderTop.

CleanShot 2022-11-01 at 10 14 32@2x

-->

CleanShot 2022-11-01 at 10 15 31@2x

six7 (approved):

PR #1365, opened 2022-10-20
fix sync styles

swordEdge:
https://www.loom.com/share/2820c621b48849c6b89404716a6269f9

### DIFF
--- a/src/app/components/AppContainer/__tests__/AppContainerIntegration.test.tsx
+++ b/src/app/components/AppContainer/__tests__/AppContainerIntegration.test.tsx
@@ -229,7 +229,7 @@ describe('AppContainer (integration)', () => {
               <AppContainer {...params} />
             </Provider>,
           );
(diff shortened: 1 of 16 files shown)


LiamMartens (approved):

LiamMartens:
@swordEdge just tests

six7 (changes requested):
@swordEdge this approach doesn't work. We cannot just remove and then recreate the style. This will break any connection the user had with that style.

We need to rename the style properly

https://user-images.githubusercontent.com/4548309/197030286-743a6949-5693-47e9-bb6a-cb13e79b143c.mp4

swordEdge:
> @swordEdge this approach doesn't work. We cannot just remove and then recreate the style. This will break any connection the user had with that style.
> 
> We need to rename the style properly

@six7  It seems like that there is no way just rename the style. ???

six7:
Works! Can you fix tests and coverage? Then let's get this merged in 🚀

PR #1369, opened 2022-10-21
fix ctrl/enter only works on modal

swordEdge:


LiamMartens (approved):

six7:
@swordEdge can you additionally add the Cmd+Enter functionality in the EditTokenForm as well? that was also part of the issue

swordEdge:
> @swordEdge can you additionally add the Cmd+Enter functionality in the EditTokenForm as well? that was also part of the issue

@six7 You mean that if the user Cmd+Enter on EditTokenFormMdal, token should be saved. Right?

PR #1373, opened 2022-10-24
356-github-sync-plugin-losing-access-tokens-sometimes

swordEdge:


LiamMartens (approved):

src/app/components/StorageItem.tsx
```diff
@@ -61,13 +61,17 @@ const StorageItem = ({ item, onEdit }: Props) => {
...
         </Box>
-        <button
-          type="button"
-          className="inline-flex text-left text-red-600 underline text-xxs"
-          onClick={handleDelete}
-        >
-          Delete local credentials
-        </button>
+        {
+          !isActive() && (
```
six7, line 65: can you revert this now that we now this didn't cause it? so the Delete button is visible when it's active even.

six7:
@swordEdge can you base this on main so we can release?

Issue #1182, opened 2022-08-17
token-transformer merges tokens before deciding what set to exclude


six7:
If you have the following structure:
```
themes
  Foundations
  Light
  Dark
```

and the "Dark" theme is setup like this:
```
Foundations: SOURCE
Light: SOURCE
Dark: ENABLED
```

And your Light theme contains all tokens that the dark theme uses:

```
Foundations:
{
  "color": {
    "neutral": {
      "10": {
        "value": "#E6EFFF",
        "type": "color"
      },
      "50": {
        "value": "#2B70E8",
        "type": "color"
      }
    }
  }
}

Light:
{
  "default": {
    "value": "{color.neutral.50}",
    "type": "color"
  },
  "gray": {
    "value": "{default}",
    "type": "color"
  },
  "background": {
    "value": "{color.neutral.10}",
    "type": "color"
  }
}

Dark:
{
  "gray": {
    "value": "{default}",
    "type": "color"
  },
  "background": {
    "value": "{color.neutral.10}",
    "type": "color"
  }
}
```

And you want to transform this with token-transformer like this:

`npx token-transformer themes output.json Foundations,Light,Dark Foundations,Light`

Your output file will be empty. Reason for that is that we probably merge tokens before we decide what to exclude. Ideally, something set as SOURCE doesn't overwrite something set as ENABLED, and we don't remove tokens that are part of any set that is ENABLED.

krizkonte:
I think that I having a relationed issue.

Imagine that I got 3 sets:
* core (source)
* desktop(enabled)
* mobile(enabled)

If I try to select a single set as:
`token-transformer tokens.json core.json **core** `
This works is fine, and i got only de core tokens only.

But if i try too use 2 sets like:
`token-transformer tokens.json core.json **core,desktop**`
This already return a blank json, even if I do not use the excludes command.

alecgeatches:
Just to tag on with a similar problem that may be related from the issue description:

> Ideally, something set as SOURCE doesn't overwrite something set as ENABLED

We have a font family in our core set of tokens `globals` and are overriding it in an enabled set. From `globals`:

![core-font-family](https://user-images.githubusercontent.com/17870752/192201385-327e589a-f96d-4af9-a537-5fc988f3f017.png)

In an enabled token set `expressive-type` we're overriding the font set with `Aktiv Grotesk VF` and use a reference to it:

[...]

PR #1379, opened 2022-10-26
1182-token-transformer-merges-tokens-before-deciding-what-set-to-exclude

swordEdge:


LiamMartens (changes requested):
@swordEdge looks fine, just need to update the test

six7:
@swordEdge updating expected test outputs is the wrong way. Why would we want those tests to all be failedToResolve??

swordEdge:
> @swordEdge updating expected test outputs is the wrong way. Why would we want those tests to all be failedToResolve??

@six7 Is it wrong? I just excluded all token sets  represented on excludes token sets before merge tokens.
This is why it is not resolved.

six7:
Yes. I would not change existing tests. I would add an additional test that does what is described in the issue. And then i'd try to make this new test not fail, plus existing ones not fail without changing them.

six7:
@swordEdge let's pause this one for now - all other issues are more important than this one.

thomasmattheussen:
I've tested this change in our setup and this seems to do what we expect!

thomasmattheussen:
> I've tested this change in our setup and this seems to do what we expect!

It seems I spoke too soon. I'm still getting a bunch of `failedToResolve` after transforming in our setup.

thomasmattheussen:
The original issue https://github.com/six7/figma-tokens/issues/1182 seems fixed by https://github.com/six7/figma-tokens/pull/1418

six7:
> The original issue #1182 seems fixed by #1418

Thank you, can confirm it is fixed - added tests for it as well now.

PR #1392, opened 2022-10-28
Feat/add new provider

andrewx82:
This is with reference to https://github.com/six7/figma-tokens/discussions/1112 . It exposes the same interface as the existing JSONBin storage provider, however it abstracts away the specifics on the JSONBin endpoint such as its hardcoded url , fixed headers ,etc and allows those to be set in the storage provider settings in the case that someone would like to create their own endpoint that is read/write. I understand the remote URL storage provider is ready only at the moment.


andrewx82:
This is an update of the old draft copy of the PR I had open https://github.com/six7/figma-tokens/pull/1140

.I've resolved the issues with the event like interface as well as the added tests and checked the e2e flow

andrewx82:
@LiamMartens apologies for the long delay

six7:
> The failure of the test coverage seems to be at a pipeline level I believe ? 

Yes, we currently have a problem with external contributions and the test coverage.

I'll review over the coming days, i don't know when @LiamMartens has capacity though

LiamMartens:
@six7 I have started reviewing this and will finish tomorrow!

src/storage/GenericVersionedStorage.ts
```diff
@@ -0,0 +1,188 @@
...
+    let body: string | undefined = JSON.stringify({
+      version: pjs.plugin_version,
+      updatedAt,
+      values: {
+        options: {},
+      },
+    }, null, 2);
+    let method: string;
+    switch (flow) {
+      case GenericVersionedStorageFlow.READ_ONLY:
```
LiamMartens, line 58: We should probably just throw an error if we are trying to create in read only mode

src/storage/GenericVersionedStorage.ts
```diff
@@ -0,0 +1,188 @@
...
+        const GenericVersionedData = validationResult.data as unknown as GenericVersionedData;
+        return this.convertGenericVersionedDataToFiles(GenericVersionedData);
+      }
+    }
+
+    return [];
+  }
+
+  public async write(files: RemoteTokenStorageFile<GenericVersionedMeta>[]): Promise<boolean> {
+    const dataObject: GenericVersionedData = {
```
LiamMartens, line 151: Same thing here, it might make sense to throw an error if this is set-up as read only and we try to write

LiamMartens (changes requested):
Some general code feedback;

Furthermore also a question to @six7, it might be worth performing a migration for this. Ie upon reading the `ApiProvidersProperty`. This way we can consolidate the providers:
- Move the JSONBin and URL types to this GenericVersioned type with the appropriate settings
- Remove JSONBin and URL from our codebase in favor of this generic impl.

src/app/components/StorageItemForm/GenericVersioned.tsx
```diff
@@ -0,0 +1,199 @@
...
+          <Button variant="secondary" size="large" onClick={onCancel}>
+            Cancel
+          </Button>
+
+          <Button variant="primary" type="submit" disabled={!values.id}>
+            Save
+          </Button>
+        </Stack>
+        {hasErrored && (
+          <div className="bg-red-200 text-red-700 rounded p-4 text-xs font-bold" data-cy="provider-modal-error">
```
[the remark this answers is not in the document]
six7: We don't so far, it's one of the remaining things to do in https://github.com/six7/figma-tokens/issues/1246 - for this PR I'd say this usage is OK as we'll replace it everywhere those error boxes are used

six7:
> Some general code feedback;
> 
> Furthermore also a question to @six7, it might be worth performing a migration for this. Ie upon reading the `ApiProvidersProperty`. This way we can consolidate the providers:
> 
> * Move the JSONBin and URL types to this GenericVersioned type with the appropriate settings
> * Remove JSONBin and URL from our codebase in favor of this generic impl.

I'm wondering how we could perform a migration if items are stored on user's devices? I.e. a user could open the plugin 5 months from now. Is there a good way how we could create these migrations to be run on `next open` - even if that isn't now but rather later?

LiamMartens:
@six7 my suggestion would be to create some kind of migration system (potentially based on last plugin version) in the outgoing function of the `ApiProvidersProperty` for example 

Currently this looks like this:
```
export const ApiProvidersProperty = new FigmaStorageProperty<StorageTypeCredentials[]>(
  FigmaStorageType.CLIENT_STORAGE,
  'apiProviders',
  (incoming) => JSON.stringify(incoming),
  (outgoing) => {
    type PossibleIncomingType = StorageTypeCredentials[] | Record<string, StorageTypeCredentials>;
    const result = tryParseJson<PossibleIncomingType>(outgoing) ?? [];
    if (Array.isArray(result)) return result;
    return Object.values(result).filter((value) => (
      typeof value === 'object'
    ));
  },
);
```

But we could add migrations here so it can update the JSONBin and URL providers to be the generic one instead upon reading (and potentially immediately write it back as well)

src/app/components/StorageItemForm/GenericVersioned.tsx
```diff
@@ -0,0 +1,199 @@
...
+                onChange={onHeaderChange}
+                type="text"
+                name="value"
+                data-index={i}
+              />
+              <button
+                type="button"
+                onClick={handleClose}
+                data-index={i}
+                className="p-4 hover:bg-gray-100 rounded focus:outline-none"
```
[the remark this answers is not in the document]
andrewx82: Sure thing, I think at the time I did the initial PR that was still the de facto way.

src/app/components/StorageItemForm/GenericVersioned.tsx
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    headers[index] = newHeader;
+
+    const newHeaders = headers.filter((x) => x.name.length);
+    headerChange(newHeaders);
+  }, [headerChange, headers]);
+
+  const handleClose = useCallback((e) => {
+    const index = Number(e.target.getAttribute('data-index'));
```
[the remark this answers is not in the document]
andrewx82: I think you're referring to this line correct?

 ` const newHeaders = removedHeaders.filter((x) => x.name.length);`
 
 that works in tandem with the header initialization where I add in an extra item with an empty name and value .
 
I did that so it would fit with the current system of always having a row at the for a user to fill out, but also so that it prevents a user from potentially having multiple empty header rows they would need to close.

Also so that we wouldn't need to do another cleanup pass later on before the provider was created to filter out any potentially empty rows as well

src/app/components/StorageItemForm.tsx
```diff
@@ -3,28 +3,46 @@ import GitForm from './StorageItemForm/GitForm';
...
+  }
 };
 
 export default function StorageItemForm({
   isNew = false, onChange, onSubmit, onCancel, values, hasErrored, errorMessage,
 }: Props) {
   const { bitBucketSync } = useFlags();
-
+  if (isNew) {
+    values = getInitialValues(values);
```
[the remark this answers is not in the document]
andrewx82: I agree, I hate this, it moves the flow outside the actual form handlers themselves

src/storage/GenericVersionedStorage.ts
andrewx82: No that would be a valid case. The "creation" would check against the remote to pull the latest tokens at the time of setup.

src/app/components/StorageItemForm/GenericVersioned.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+
+export default function GenericVersionedForm({
+  onChange, onSubmit, onCancel, values, hasErrored, errorMessage,
+}: Props) {
+  const handleSubmit = React.useCallback((event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    const validationResult = zodSchema.safeParse(values);
+    if (validationResult.success) {
```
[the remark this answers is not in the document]
andrewx82: I used the same code as the other forms are using when it comes to validation, but you're right, I didn't see them display the errors to the user by using the validationResult

andrewx82:
@LiamMartens Do you think the new changes are satisfactory?

six7:
> @six7 my suggestion would be to create some kind of migration system (potentially based on last plugin version) in the outgoing function of the `ApiProvidersProperty` for example
> 
> Currently this looks like this:
> 
> ```
> export const ApiProvidersProperty = new FigmaStorageProperty<StorageTypeCredentials[]>(
>   FigmaStorageType.CLIENT_STORAGE,
>   'apiProviders',
>   (incoming) => JSON.stringify(incoming),
>   (outgoing) => {
>     type PossibleIncomingType = StorageTypeCredentials[] | Record<string, StorageTypeCredentials>;
>     const result = tryParseJson<PossibleIncomingType>(outgoing) ?? [];
>     if (Array.isArray(result)) return result;
>     return Object.values(result).filter((value) => (
>       typeof value === 'object'
>     ));
>   },
> );
> ```
> 
> But we could add migrations here so it can update the JSONBin and URL providers to be the generic one instead upon reading (and potentially immediately write it back as well)

I like that! On road, migrate and change the stored data. We'd probably need to keep this in the plugin for a while - but eventually be able to remove it.

andrewx82:
Hi all, are the updated changes satisfactory?

src/stories/GenericVersionedStorageForm.stories.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+<Meta 
+  title='GenericVersionedForm' 
+  component={GenericVersioned}
+  args={{
+    onChange:()=>{},
+    values: {
+      name: '', 
+      additionalHeaders: [],
+      flow: GenericVersionedStorageFlow.READ_WRITEt,
```
six7, line 13: is the t at the end on purpose?

src/app/store/providers/generic/versionedStorage.ts
```diff
@@ -0,0 +1,212 @@
...
+} from '@/types/StorageType';
+import { RemoteResponseData } from '@/types/RemoteResponseData';
+import { saveLastSyncedState } from '@/utils/saveLastSyncedState';
+
+export async function updateGenericVersionedTokens({
+  tokens, themes, context, updatedAt, oldUpdatedAt = null,
+}: UpdateRemoteFunctionPayload) {
+  const { id, additionalHeaders, flow } = context as GenericVersionedStorageType;
+  try {
+    if (!id) throw new Error('Missing Generic Versioned Storage ID ');
```
six7, line 27:
```suggestion
    if (!id) throw new Error('Missing Generic Versioned Storage ID');
```

src/app/store/providers/generic/versionedStorage.ts
```diff
@@ -0,0 +1,212 @@
...
+          case GenericVersionedStorageFlow.READ_WRITE:
+          case GenericVersionedStorageFlow.READ_WRITE_CREATE:
+            storage.save(payload);
+            break;
+          default:
+        }
+      } else {
+        // Tell the user to choose between:
+        // A) Pull Remote values and replace local changes
+        // B) Overwrite Remote changes
```
six7, line 63: We can remove that part from here (we never got to that part for JSONBin but something we should do in a future release - for all providers)

src/app/components/StorageItemForm/GenericVersioned.tsx
```diff
@@ -0,0 +1,207 @@
...
+              <DropdownMenuRadioItem value={GenericVersionedStorageFlow.READ_WRITE_CREATE}>
+                <Text>
+                  {GenericVersionedStorageFlow.READ_WRITE_CREATE}
+                </Text>
+              </DropdownMenuRadioItem>
+            </DropdownMenuRadioGroup>
+          </DropdownMenuContent>
+        </DropdownMenu>
+        <Stack direction="column" gap={4}>
+          <div className="font-bold">Additional Headers</div>
```
six7, line 159: Can we change this to a `<Heading>` so we're not using Tailwind here?

six7 (changes requested):
Just some smaller items I noticed. I'll DM you in Slack for more!

src/stories/GenericVersionedStorageForm.stories.mdx
andrewx82: Great catch

andrewx82:
I've created a toy storage endpoint that uses the CREATE flow https://github.com/andrewx82/figma-tokens-generic-storage-example

andrewx82:
I've altered the merge target to be the `next` branch to get access to the new updates to ensure there are no regressions, as well as added in the icons support.

Issue #1353, opened 2022-10-17
Gradient from Figma Tokens has different position of anchor points


samuelg0rd0n:
**Describe the bug**
When I import style (color gradient) from Figma to Figma Tokens and the gradient is, for example, from top left corner to the bottom right corner, and then apply it from Figma Tokens back to a shape in Figma, the gradient doesn't start in top left corner and doesn't end in bottom right corner. Instead, it's much smaller. It starts several pixels away from those corners. The difference is quite big and of course the applied gradient then doesn't look as intended. Colors on both ends take up more area in the shape.

**To Reproduce**
1. Create a gradient fill in Figma
2. Import styles in Figma Tokens
3. Apply this gradient color token back to a shape

**Expected behavior**
I understand that it's probably not possible for Figma Tokens to remember the positions of those anchor points so I'd expect the gradient to be across the whole shape. From one corner to another or from one side to another (depends on the direction and the shape).

**Screenshots or Screencasts**
<img width="259" alt="Screenshot 2022-10-17 at 16 06 39" src="https://user-images.githubusercontent.com/4357381/196232256-c3149803-1861-46fc-9f39-417f0d34e39d.png">
<img width="249" alt="Screenshot 2022-10-17 at 16 07 04" src="https://user-images.githubusercontent.com/4357381/196232264-e28c9888-169b-4a01-9c19-3ed11e67c27e.png">

**Figma file (optional)**
https://www.figma.com/file/qDrg5Om4DYmDLgQO523prm/Figma-Tokens-playground?node-id=0%3A1

six7:
Thanks for the detailed report @samuelg0rd0n 🙏

Pinging @LiamMartens on this one - any idea if this is something that we could fix?

LiamMartens:
@six7 probably - I will have to investigate

six7:
Released in 129

PR #1397, opened 2022-11-01
fix: Gradient scale after rotation

LiamMartens:


LiamMartens:
@six7 this should fix #1353

six7 (approved):

six7:
Nice 🎉

PR #1402, opened 2022-11-02
Fix typo in node.test.ts

sairion:
While I was browsing source, I found this. 
don't think it is intended, but please feel free to close it or merge :)


six7:
🎉  thanks for spotting that one!

PR #1405, opened 2022-11-03
1308 optimize themesjson file

swordEdge:
image

image


src/app/store/providers/bitbucket/bitbucket.tsx
```diff
@@ -224,16 +228,17 @@ export function useBitbucket() {
...
             !isEqual(content.tokens, tokens)
-            || !isEqual(content.themes, themes)
+            || !isEqual(optimizeThemes(content.themes), optimizedThemes)
             || !isEqual(content.metadata?.tokenSetOrder ?? Object.keys(tokens), Object.keys(tokens))
           ) {
             const userDecision = await askUserIfPull();
             if (userDecision) {
               const sortedValues = applyTokenSetOrder(content.tokens, content.metadata?.tokenSetOrder);
-              saveLastSyncedState(dispatch, sortedValues, content.themes, content.metadata);
+              const recoveredThemes = recoverOptimizedThemes(content.themes, content.tokens);
```
LiamMartens, line 237: @swordEdge why is this necessary? I would say it is probably okay to notify the user of the changed values because of the optimization and to let them simply write it back to the remote repository.

FYI @six7

LiamMartens (changes requested):
@swordEdge I would consolidate this logic inside theme related logic so it's more centralized;
specifically in 2 places:
- `setTokenData.ts` - we can add a `status !== TokenSetStatus.DISABLED` into the selectedTokenSets filter to remove the DISABLED status token sets from the theme when reading (this will prompt the user to write back to the repo since it won't match anymore - but from my POV this is okay since we are changing the data @six7 )
- `saveTheme` - here we can filter out the disabled token sets when updating/adding the theme. This way they won't make into the data in the first place.

This way there is a much better separation of concerns between the remote content/file logic and the themes logic.

src/app/store/providers/bitbucket/bitbucket.tsx
swordEdge: > @swordEdge why is this necessary? I would say it is probably okay to notify the user of the changed values because of the optimization and to let them simply write it back to the remote repository.
> 
> FYI @six7

@LiamMartens You're right. Right now users already have the non-optimized themes, so if we remove this part, plugin asks to write again to remote repository.

swordEdge: @six7 Waiting for your response.

six7: Yes, lets ask and push back 👍

LiamMartens (approved):

PR #1409, opened 2022-11-03
Improve error setting font weight/family

six7:
Fixes https://github.com/six7/figma-tokens/issues/1396

Shows a more prominent error when font weight/family combination can't be applied.

Also adds error tracking so we're aware of which combinations don't work.

CleanShot 2022-11-03 at 22 21 50@2x

### DIFF
--- a/src/plugin/setTextValuesOnTarget.ts
+++ b/src/plugin/setTextValuesOnTarget.ts
@@ -1,8 +1,7 @@
 /* eslint-disable no-param-reassign */
-import * as Sentry from '@sentry/react';
 import { SingleTypographyToken } from '@/types/tokens';
 import { transformValue } from './helpers';
-import { notifyUI } from './notifiers';
+import { notifyException, notifyUI } from './notifiers';
 
 export default async function setTextValuesOnTarget(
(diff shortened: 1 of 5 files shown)


src/plugin/setTextValuesOnTarget.ts
```diff
@@ -66,7 +68,8 @@ export default async function setTextValuesOnTarget(
               }
             })
             .catch(() => {
-              // TODO: Track this in mixpanel so we can add missing weights
+              notifyUI(`Error setting font family/weight combination for ${family}/${style}`, { error: true });
+              Sentry.captureException({ error: 'Font not found', family, style });
```
LiamMartens, line 72: Do we know if sentry actually works inside the plugin sandbox? (Otherwise we might need to forward a msg to the UI code to send the message to Sentry)

LiamMartens (approved):
@six7 just one question

src/plugin/setTextValuesOnTarget.ts
six7: You're right - that didn't even work! Added that forward msg now!

PR #1412, opened 2022-11-04
1404 plugin crashes when referencing a token with ++ in the naming of the tokentoken group

swordEdge:
https://www.loom.com/share/49720892a81641c0a59d7eaffb32785d


src/app/components/DownshiftInput/DownshiftInput.tsx
```diff
@@ -126,7 +126,34 @@ export const DownshiftInput: React.FunctionComponent<DownShiftProps> = ({
   const referenceTokenTypes = useReferenceTokenType(type as TokenTypes);
   const getHighlightedText = useCallback((text: string, highlight: string) => {
     // Split on highlight term and include term into parts, ignore case
-    const parts = text.split(new RegExp(`(${highlight})`, 'gi'));
+    let parts: string[] = [];
```
LiamMartens, line 129: @swordEdge this seems convoluted, we can just update the regex by escaping the special characters like this:
` const parts = text.split(new RegExp(`(${highlight.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')})`, 'gi'));`

LiamMartens (changes requested):

LiamMartens (approved):

PR #1423, opened 2022-11-09
Fix types for expanded (typography) tokens

thomasmattheussen:
Fixes https://github.com/six7/figma-tokens/issues/1417

The ideal fix IMO would be to make all typography types singular. That is a more elaborate (and breaking) change though. So I thought this could already be a quick fix so it's at least consistent :)


six7 (approved):
TY! Will try to release this sometime next week.

PR #1426, opened 2022-11-10
546 applying and updating tokens works for one user but another user in the same file doesnt see the changes

swordEdge:


six7 (changes requested):
This is the wrong approach.

I would suggest to try to use the plugin with two different Figma Users in the same file to see where the problem lies.

Let both have the plugin open, then let User A create and apply a few tokens on one, then let User B apply tokens as well on the same rectangle.

Expected is that User B doesn't override the tokens but is just overriding the token they applied + is informed of the tokens that were applied by User A

six7 (changes requested):
this is unfortunately again the wrong approach. this invalidates the cache every time selection changes. that would mean that users have a very hard time using the plugin.

- Cache should stay and be performant
- Listen to the `documentchanged` event on Figma (that will fire whenever a user made a change). Make it so that if this fires it invalidates the cache for the changed node.
- That way users would always have an up to date cache / Inspect - even if other users make that change.

src/plugin/sendDocumentChange.ts
```diff
@@ -0,0 +1,9 @@
+import { defaultNodeManager } from './NodeManager';
+import { findAll } from '@/utils/findAll';
+
+export async function sendDocumentChange(event: DocumentChangeEvent) {
+  const allNodes = findAll([figma.root], false);
```
six7, line 5: Can you make it work without this step? This is bad performance if we scan the doc on every document change. Find the changed id's in the cache rather

six7 (changes requested):
Also adding tests would be good.

six7 (changes requested):
This is already great, thank you for that direction!

Could you make one more change? Right now I have to click on another tab and then back to Inspect to see the changes. I would like these to be reflected in real-time. So when new values come in, we show that instantly. In the case of that video below: I should not need to click on another tab and back to see the black token applied.

https://user-images.githubusercontent.com/4548309/204135075-b9f85982-ead7-4481-8002-7b1193d101bb.mp4

This will be a great improvement for sure! Let's also add some tests to this.

I would also like to hear from @mikekamminga and @esthercheran on how this impacts performance in large files when 2 users work in it.

rbosker:
@swordEdge The functionality is working, only the UI is not updating at the same time. Now you first have to visit another tab in the plugin, to make the changes in the UI visible. 

Maybe you can add a watcher when the state is updated to re-render the UI.

six7:
@swordEdge did you see the feedback?

swordEdge:
> @swordEdge did you see the feedback?

> @six7 Sorry. I just forgot to push. haha.

six7:
very cool! thank you!

PR #1447, opened 2022-11-22
1425 gradients cant be used directly in compositions

swordEdge:
https://www.loom.com/share/c5aaf49cd71f491a92d22efe2e5ff40f

### DIFF
--- a/src/plugin/setValuesOnNode.ts
+++ b/src/plugin/setValuesOnNode.ts
@@ -421,7 +421,7 @@ export default async function setValuesOnNode(
       }
 
       if (values.asset && typeof values.asset === 'string') {
-        if ('fills' in node && data.asset) {
+        if ('fills' in node) {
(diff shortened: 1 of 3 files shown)


src/plugin/setValuesOnNode.ts
```diff
@@ -421,7 +421,7 @@ export default async function setValuesOnNode(
       }
 
       if (values.asset && typeof values.asset === 'string') {
-        if ('fills' in node && data.asset) {
```
six7, line 424: Is this by accident?

six7 (changes requested):

src/plugin/setValuesOnNode.ts
swordEdge: @six7 data.assets condition is not neccessary. You'll notice that if you see the source code. That's why I removed this part.

six7 (approved):

PR #1453, opened 2022-11-24
1264 refactor how we show syncsettings

swordEdge:
image
image
image


six7 (changes requested):
i think the Vercel deployment fails because we have a `storybook` story that renders the SyncSettings. can you make sure this doesnt fail?

Also, some things:

There's an error in the console about keys:
CleanShot 2022-11-27 at 12 47 37@2x

The validation doesn't seem to work when the remote doesn't contain tokens (i.e. setting it to path `tokens` but that path doesn't exist). It does work when the credentials are incorrect, though 👍 

CleanShot 2022-11-27 at 12 49 24@2x

Otherwise this is great - I'll do the style changes and addition of the logos after we merge this in 👍  Can you fix the stuff I mentioned above?

swordEdge:
> The validation doesn't seem to work when the remote doesn't contain tokens (i.e. setting it to path `tokens` but that path doesn't exist). It does work when the credentials are incorrect, though 👍

> @six7 I'm not sure what you mean? In your example, if the path doesn't exist, then the plugin tries to push. no?

six7 (approved):

PR #1464, opened 2022-12-02
1462 issue related to inherittypelevel

swordEdge:


six7:
This feels a little overly complex. The tokens that we send to the git clients shouldnt be aware of such logic, we need to do that one step before. We also have a property called 'internal__Parent' which we do not expose, cant we just do it like we do there? Like in a function where we collect tokens to send to the git clients, not inside the actual sync solutions. Right now theres too much logic in there.

PR #1472, opened 2022-12-08
Sticky Inspect action bar

dermyhughes:
Issue:
When you are deep inspecting a lot of elements that the container overflows to scrolling, the action bar containing Select All, Bulk Remap, and Remove Selected can get hidden and requires scrolling up.

Fix:
This change moves that action bar outside of the scroll-container so the action bar is always visible no matter how far you scroll down, removing the need to scroll back to the top and improving UX.

ezgif-4-dd4f5af8f0

If no tokens are available in the inspect container, the bar will not be shown.

image


six7:
@dermyhughes Thank you so much, this is awesome and such a quick fix. I'll merge this in!

PR #1479, opened 2022-12-13
401 custom rem configuration support

swordEdge:


src/app/components/Settings/Settings.tsx
```diff
@@ -2,19 +2,26 @@
...
+import {
+  ignoreFirstPartForStylesSelector, prefixStylesWithThemeNameSelector, uiStateSelector, tokensSelector, usedTokenSetSelector, activeTokenSetSelector, aliasBaseFontSizeSelector,
+} from '@/selectors';
 import Stack from '../Stack';
 import Box from '../Box';
 import AddLicenseKey from '../AddLicenseKey/AddLicenseKey';
 import { Divider } from '../Divider';
-import { track } from '@/utils/analytics';
 import OnboardingExplainer from '../OnboardingExplainer';
+import DownshiftInput from '../DownshiftInput';
```
six7, line 23: can you create a new component for `<RemConfiguration>` where all this new code lives? that cleans up the files and doesnt clutter Settings.tsx too much.

src/plugin/figmaTransforms/generic.test.ts
```diff
@@ -2,15 +2,15 @@ import { convertTypographyNumberToFigma, fakeZeroForFigma } from './generic';
 
 describe('convertTypographyNumberToFigma', () => {
   it('converts an input number-like string and returns a transformed value', () => {
```
six7, line 4: can you add tests here to test with also not the default value? e.g. that it still works when used with value sthat are not 16? like a string, an object, a `2px` a `4rem` etc.

six7 (changes requested):
very nice! i think the only thing missing is a few tests to make sure this behavior works even with a custom rem font size, doesnt break. otherwise nice!!

PR #1480, opened 2022-12-14
1411 when changing a group name let users change the full path

swordEdge:
https://www.loom.com/share/8851332c3c194d7590e2363ba01a616e


six7 (approved):
🎉

PR #1484, opened 2022-12-15
1476 investigate json editor improvements

swordEdge:
image


six7:
Can you create a writeup of what our options are, advantages of each option and bundle size increases for each option?

PR #1498, opened 2022-12-28
1166 introduce color modifiers

swordEdge:


src/app/components/EditTokenForm.tsx
```diff
@@ -229,6 +225,25 @@ function EditTokenForm({ resolvedTokens }: Props) {
     }
   }, [internalEditToken]);
 
+  const handleColorModifyChange = React.useCallback((newModify: ColorModifier) => {
```
six7, line 228: should we move this logic also into ColorTokenForm? then it would be nicely separated?

src/app/store/useManageTokens.tsx
```diff
@@ -21,6 +22,7 @@ type EditSingleTokenData = {
   description?: string;
   oldName?: string;
   shouldUpdateDocument?: boolean;
+  $extensions?: { 'com.figmatokens': { modify: ColorModifier } }
```
six7, line 25: can we replace all occurences of `com.figmatokens` with `studio.tokens` ? We've changed our name recently, so let's reflect that as well in the extension 👍

src/utils/alias/__tests__/getAliasValue.test.ts
```diff
@@ -53,6 +53,37 @@ describe('getAliasValue', () => {
       value: '#d0efeb80',
       type: TokenTypes.COLOR,
     },
+    {
+      name: 'colors.ligth-extension',
```
six7, line 57: `light` not `ligth`

six7 (commented):
havent been able to fix the convertModifiedColorToHex error you mentioned either :/

SorsOps:
Took at look at the issue when having Figma execute the built bundle, seems to be colorjs.io has ES6 code present in all their dist files. Easiest thing is to just have a loader handle it 

The following changes need to be done to resolve it 

In babel.config.js

```js
...
  plugins: ['@babel/proposal-class-properties', '@babel/proposal-object-rest-spread', "@babel/transform-typescript","@babel/plugin-proposal-private-methods"],
  ...
```

In webpack.config.js (and webpack-transform.config.js to be safe) , add the following under rules

```js
      {
        test: /\.c?js$/,
        use: [
          {
            loader: 'babel-loader',
          },
        ],
        exclude:  /node_modules\/(?!(colorjs.io)\/)/,
      },
```

This means that we exclude transpilation of all other node module but the colorjs.io module. Technically it could be an include ,but it handles an raw js we use in the future

and lastly 

```sh
yarn add -D @babel/plugin-proposal-private-methods
```

Then our webpack should handle transpiling it down to ES5

src/utils/convertModifiedColorToHex.ts
```diff
@@ -1,29 +1,23 @@
-import chroma from 'chroma-js';
-import Color from 'colorjs.io';
+import Color from 'colorjs.io/dist/color.legacy';
```
SorsOps, line 1: The babel loader should fix our issues for us
```suggestion
import Color from 'colorjs.io';
```

src/app/components/ColorTokenForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+      />
+      {inputHelperOpen && (
+        <ColorPicker value={internalEditToken.value} onChange={handleColorValueChange} />
+      )}
+      <Box css={{ display: 'flex', justifyContent: 'space-between', alignItems: 'center' }}>
+        <Heading size="xsmall">Modify</Heading>
+        {
+          !modifyVisible ? (
+            <IconButton
+              tooltip="Add a new modify"
```
six7, line 220:
```suggestion
              tooltip="Add new modifier"
```

six7 (changes requested):
some more things i encountered:
- we're currently saving the `value` field as a `string` in JSON. can we save it as a `number` instead?
- can you add something to the tooltip of color tokens that use a modifier? just beneath the value: `lighten(0.5) / LCH` so `{operation}({value}) / {space})` or for mix: `mix(#ffffff, 0.5) / LCH`.

src/app/components/ColorTokenForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+        label={internalEditToken.schema?.property}
+        resolvedTokens={resolvedTokens}
+        initialName={internalEditToken.initialName}
+        handleChange={handleColorChange}
+        setInputValue={handleColorDownShiftInputChange}
+        placeholder="#000000, hsla(), rgba() or {alias}"
+        prefix={(
+          <button
+            type="button"
+            className="block w-4 h-4 rounded-sm cursor-pointer shadow-border shadow-gray-300 focus:shadow-focus focus:shadow-primary-400"
```
six7, line 203: could you convert this to Stitches css? we plan to remove tailwind.

src/app/components/ColorTokenForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+                    value={internalEditToken?.$extensions?.['studio.tokens']?.modify?.color}
+                    type={TokenTypes.COLOR}
+                    resolvedTokens={resolvedTokens}
+                    handleChange={handleMixColorInputChange}
+                    setInputValue={handleMixColorChange}
+                    placeholder="#000000, hsla(), rgba() or {alias}"
+                    prefix={(
+                      <button
+                        type="button"
+                        className="block w-4 h-4 rounded-sm cursor-pointer shadow-border shadow-gray-300 focus:shadow-focus focus:shadow-primary-400"
```
six7, line 296: could you convert this to Stitches css? we plan to remove tailwind.

src/app/components/ColorTokenForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+            }
+            <Box css={{ display: 'flex', position: 'relative', width: '100%' }} className="input">
+              <StyledPrefix isText>{getIconComponent}</StyledPrefix>
+              <StyledInput onChange={handleModifyValueChange} value={internalEditToken?.$extensions?.['studio.tokens']?.modify?.value} required />
+            </Box>
+          </>
+        )
+      }
+      {(checkIfContainsAlias(internalEditToken.value) || internalEditToken?.$extensions?.['studio.tokens']?.modify) && (
+      <div className="flex p-2 mt-2 font-mono text-gray-700 bg-gray-100 border-gray-300 rounded text-xxs itms-center">
```
six7, line 319: could you convert this to Stitches css? we plan to remove tailwind.

src/app/components/ColorTokenForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+        {internalEditToken.type === 'color' ? (
+          <div className="w-4 h-4 mr-1 border border-gray-200 rounded" style={{ background: String(displayColor) }} />
```
six7, line 321: could you convert this to Stitches css? we plan to remove tailwind.

src/utils/is/isSingleColorToken.ts
```diff
@@ -0,0 +1,7 @@
+import { TokenTypes } from '@/constants/TokenTypes';
+import { SingleColorToken, SingleToken } from '@/types/tokens';
+
+export function isSingleColorToken(token: SingleToken | any): token is SingleColorToken {
```
six7, line 4: Let's call this isColorToken instead. Color tokens can only be single.

src/app/components/ColorPickerTrigger.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Box from './Box';
+
+export default function ColorPickerTrigger({
+  onClick, background,
+}: { onClick?: () => void, background: string | undefined }) {
+  return (
```
six7, line 7: Never put an onClick on a div, this is bad for a11y. Instead, render a button: `styled('button', ..`

six7:
@swordEdge One last thing: Can you make sure users can enter both `0.5` as well as `0,5` for the `value`? For me right now I have to enter `0,05` - I believe this depends on localization? Can we parse the input on change and change that to 0.5 always (the decimal point, not the decimal comma). Figma does the same if you type `0,5` in a width in Figma it converts to `0.5`.

six7:
I found one more edge case:

If a user already had something else in their `$extensions` property (which will be likely given this is going to be how in the spec users are able to add extra metadata) we currently override the whole object

To reproduce:
- add a $extensions field with some other identifier to a token
- then add a color modifier to that token via the UI and save
- notice how we removed the other $extensions value.

I'd expect that same $extensions property to still be there. We do not want to delete that for the user, we just want to append.

https://user-images.githubusercontent.com/4548309/211678220-afe9d2a8-4ff2-4378-81c3-da0b8557e56f.mp4

six7 (approved):
🎉

PR #1520, opened 2023-01-05
1517-importing-paragraphindent-not-working

swordEdge:
https://www.loom.com/share/824d4c68dec7484b94a503b3235f6fdd


six7:
Could you base this on the `release-133` branch? Then we can include it in a fix release.

PR #1521, opened 2023-01-05
1517-importing-paragraphindent-not-working-base-release-113

swordEdge:


src/plugin/pullStyles.ts
```diff
@@ -137,6 +137,7 @@ export default function pullStyles(styleTypes: PullStyleOptions): void {
       .map((size, idx) => ({
         name: `paragraphIndent.${idx}`,
         value: size.toString(),
```
six7, line 139: We'll need to add `px` at the end here so they are correct dimension tokens.

six7 (changes requested):

six7 (approved):

PR #1534, opened 2023-01-10
Copy token path

fokkerone:


six7 (changes requested):
This doesn't work when used with the keyboard (i.e. right click a token then use arrow keys to navigate down).

Can't we just use `navigator.clipboard.writeText(value)`?

fokkerone:
I don´t think that there is an Navigator.Clipboard in Figma-Electron App. 
To make it work, we might use.

`
const {clipboard} = require('electron');
clipboard.writeText('Example String');
let clipboardStr = clipboard.readText();
`
in Electron and make use of Navigator.Clipboard in Figma-Web
thats why I make use of these universal Clipboard solutions.

I try to make it work more a11Y

six7 (approved):
awesome, thank you for that contribution 🎉  i'll merge this in `next` so it'll go in the next major release.

fokkerone:
@jan : We need to test it in figma-web. due to the fact dev plugins only runs in figma app while developing. 
plz write me after you release.

PR #1536, opened 2023-01-11
1527 integrate none as a keyword in token values

swordEdge:
https://www.loom.com/share/f2e1de3388fd4041b0c233a31e4bc70e

### DIFF
--- a/src/app/components/InspectorTokenSingle.tsx
+++ b/src/app/components/InspectorTokenSingle.tsx
@@ -99,7 +99,7 @@ export default function InspectorTokenSingle({
         {
           isBrokenLink && token.value !== 'none' && <IconBrokenLink />
         }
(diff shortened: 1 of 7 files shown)


six7 (changes requested):
One thing I noticed is that when you set an asset token on a layer, e.g. `https://images.unsplash.com/photo-1673557143243-39f2995a3bca?ixlib=rb-4.0.3&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=2340&q=80` and then you set that token's value to `none` the plugin says `Failed to load image`. In that case it's correct we do not set an image, but we also shouldn't even try to do so (so the error is not shown).

I also think we should show the `none` icon on Inspect for these tokens.

PR #1537, opened 2023-01-12
GitLab sync: Include inherited members when checking for write permissions

robin-ambachtsheer:
We're currently blocked from using the plugin because after connecting to our private GitLab server the plugin refuses to let us write any changes (even though it was able to create a commit with the new json file).

After some debugging we found that the GitLab API was returning a 404 error when trying to fetch the member data for the project. This is because our GitLab repository inherits the members from another group and it doesn't have any direct members itself. To get all the members including the inherited ones you need to fetch another endpoint.

By including the `includeInherited` option in the calls to `GroupMembers` and `ProjectMembers` from the `@gitbeaker/browser` package this issue will be resolved for us as it will hit the correct endpoint. According to GitLab's own documentation this should include all members at all times https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-and-invited-members, so it should not have any impact on existing use cases.

### DIFF
--- a/src/storage/GitlabTokenStorage.ts
+++ b/src/storage/GitlabTokenStorage.ts
@@ -90,11 +90,15 @@ export class GitlabTokenStorage extends GitTokenStorage {
     try {
       if (!currentUser || currentUser.state !== 'active') return false;
 
-      const groupPermission = await this.gitlabClient.GroupMembers.show(this.groupId, currentUser.id);
+      const groupPermission = await this.gitlabClient.GroupMembers.show(this.groupId, currentUser.id, {
+        includeInherited: true,
+      });
       return groupPermission.access_level >= GitLabAccessLevel.Developer;
     } catch (e) {
       try {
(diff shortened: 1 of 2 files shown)


robin-ambachtsheer:
@six7 would you mind giving this a look? It's blocking our usage and we'd love to move forward with integrating your fantastic plugin into our process.

six7:
Actually could you update tests so they're not failing? I guess we'll just need to update those with the now added additional parameter.

six7 (changes requested):
Actually could you update tests so they're not failing? I guess we'll just need to update those with the now added additional parameter.

robin-ambachtsheer:
Of course, sorry I thought I had run the tests but apparently not. Added the additional parameter to the tests.

six7:
Awesome, thank you! I'll merge this in tomorrow morning and release this 🙏

PR #1538, opened 2023-01-12
1475 showing styles in inspect

swordEdge:
https://www.loom.com/share/a96dfe88a07d43a5baf83e1ff56b9121


src/app/components/InspectorResolvedToken.tsx
```diff
@@ -2,15 +2,21 @@ import React from 'react';
...
+type Props = {
+  name: string;
+  value: SingleToken['value']
+  type: string;
+};
+
+export default function InspectorResolvedToken({ style }: { style: Props }) {
   // TODO: Introduce shared component for token tooltips
-  if (!token) {
+  if (!style) {
```
six7, line 19: why the rename to style?

six7 (changes requested):
Can we remove the `Remap` functionality from the style entries in Inspect? Right now it's non-functional for tokens.

Otherwise this is really great! 👍 I pushed the icon changes.

six7 (approved):
nice, just the merge conflict now

PR #1548, opened 2023-01-16
1527 integrate none as a keyword in token values base on main

swordEdge:


six7 (changes requested):
This PR seems to have removed the ability to

- Create a rectangle, set background to a token called `background`
- Create a component out of that rectangle
- Duplicate it
- On the rectangle layer, select it and select that token in Inspect to `Set as none`.

It shows up as `none` now, but it only seems like it. We actually never save that to the pluginData anymore from the code (you remove that function). Once you reopen the plugin and apply again, it will revert back, it shouldn't.


https://user-images.githubusercontent.com/4548309/212753288-f8dcc663-7015-4aa6-a0ff-f1a5d60bd768.mp4

src/plugin/pluginData.ts
```diff
@@ -92,25 +92,15 @@ export async function removePluginData({ nodes, key, shouldRemoveValues = true }
...
-        removeValuesFromNode(node, prop);
-      });
-    }
+    node.setPluginData(key, 'none');
+    tokensSharedDataHandler.set(node, key, 'none');
+    await defaultNodeManager.updateNode(node, (tokens) => (
+      omit(tokens, key)
+    ));
+    removeValuesFromNode(node, key);
+    node.setPluginData('values', 'none');
```
six7, line 103: This isn't required.

The `values` property was a property we had in plugin versions 1-72. It is deprecated, and is only used for the case where older plugin users have this on their layers and we need to migrate it.

You can remove it from the `none` property.

six7 (commented):
just one minor thing otherwise i think this is good to go!

six7 (approved):

PR #1552, opened 2023-01-17
1299 duplicate token group across sets

swordEdge:


src/app/components/TokenGroup/TokenGroupHeading.tsx
```diff
@@ -29,62 +27,71 @@ export type Props = {
...
 
   const handleNewTokenGroupNameChange = React.useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
     setNewTokenGroupName(e.target.value);
   }, []);
 
-  const handleSetNewTokenGroupNameFileClose = React.useCallback(() => {
-    setShowNewGroupNameField(false);
+  const handleRenameTokenGroupModalClose = React.useCallback(() => {
+    setShowRenameTokenGroupModal(false);
+    setNewTokenGroupName(path);
```
six7, line 72: this feels a bit wrong. why change the path on modal close? cant we just set the token group name when we open the rename modal?

src/app/components/TokenGroup/TokenGroupHeading.tsx
```diff
@@ -29,62 +27,71 @@ export type Props = {
...
+  }, [newTokenGroupName, path, renameGroup, type, remapTokensInGroup]);
 
-    setShowNewGroupNameField(false);
-    renameGroup(`${path}${copyName}`, `${newTokenGroupName}`, type);
-
-    remapTokensInGroup({ oldGroupName: `${path}${copyName}.`, newGroupName: `${newTokenGroupName}.` });
-    setIsTokenGroupDuplicated(false);
-    setCopyName('');
-  }, [copyName, newTokenGroupName, path, renameGroup, type, remapTokensInGroup]);
+  const handleDuplicateTokenGroupSubmit = React.useCallback(async (e: React.FormEvent<HTMLFormElement>) => {
```
six7, line 57: i think for the size of this file it'd make sense to put all the duplicate modal logic into its own file

src/app/components/MultiSelectDownshiftInput.tsx
```diff
@@ -0,0 +1,215 @@
+import React from 'react';
```
six7, line 1: I think this is too complicated. Can we instead just use a `Dropdown` and have users select items from here, like outlined in the issue? the token input is a bit too much for this use case

six7 (changes requested):
other than the mentioned stuff it works great.

six7 (approved):

PR #1579, opened 2023-01-31
1570-color-modifiers-not-able-to-delete-a-modifier-after-saving-a-modified-token

swordEdge:
https://www.loom.com/share/a1baf8b5a77044d9bfd0847c1f7df342


src/app/components/EditTokenForm.tsx
```diff
@@ -206,11 +206,17 @@ function EditTokenForm({ resolvedTokens }: Props) {
   const removeColorModify = React.useCallback(() => {
     const newValue = internalEditToken;
     delete newValue?.$extensions;
+    console.log('newe', {
```
six7, line 209: log

PR #1580, opened 2023-02-01
1567-issue-when-adding-shadows-in-reference-mode

swordEdge:
@six7 You can see it again.
https://www.loom.com/share/e0fbe664f55d48998961248f56ea8187

### DIFF
--- a/src/app/components/BoxShadowInput.tsx
+++ b/src/app/components/BoxShadowInput.tsx
@@ -62,7 +62,7 @@ export default function BoxShadowInput({
   const addShadow = React.useCallback(() => {
     if (Array.isArray(internalEditToken.value)) {
       handleBoxShadowValueChange([...internalEditToken.value, newTokenValue]);
-    } else if (typeof internalEditToken.value === 'undefined') {
+    } else if (typeof internalEditToken.value === 'undefined' || typeof internalEditToken.value === 'string') {
       handleBoxShadowValueChange(compact([newTokenValue, newTokenValue]));
     } else if (typeof internalEditToken.value !== 'string') {
       handleBoxShadowValueChange(compact([internalEditToken.value, newTokenValue]));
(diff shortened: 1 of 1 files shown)


six7:
Can we change it so when you detach the values of the aliased tokens get inserted? In that case `2 2 2 2 etc` ?

six7 (changes requested):
Two things:

- When clicking the `+` initially, can we make it so that there's a second form added? Right now I have to press twice to add the second form
- There's a strange case where when going to reference mode and then back to input mode I can no longer press the +. see video below
- Would it be possible that we just allow users to also use multiple when using aliases? So we can do a shadow that consists of many aliases to other shadows? That way you could keep the + in alias mode, but it'd just give you multiple of these single-reference field inputs?

https://user-images.githubusercontent.com/4548309/218258590-999a37e9-9fe1-4345-bc26-75cf896dc90d.mp4

swordEdge:
> Two things:
> 
> * When clicking the `+` initially, can we make it so that there's a second form added? Right now I have to press twice to add the second form
> * There's a strange case where when going to reference mode and then back to input mode I can no longer press the +. see video below
> * Would it be possible that we just allow users to also use multiple when using aliases? So we can do a shadow that consists of many aliases to other shadows? That way you could keep the + in alias mode, but it'd just give you multiple of these single-reference field inputs?


> @six7 Implemented all features you mentioned except the third part (add button on alias mode). Do you think it is necessary? And If it is really neccessary, I'd like you to create a new issue about that one.

src/app/components/BoxShadowInput.tsx
```diff
@@ -102,11 +109,11 @@ export default function BoxShadowInput({
         {
           mode === 'input' ? (
             <DndProvider backend={HTML5Backend}>
-              {Array.isArray(internalEditToken.value) ? (
-                internalEditToken.value.map((token, index) => (
+              {Array.isArray(mappedItems) ? (
+                mappedItems.map((token, index) => (
```
six7, line 113: instead of casting all values `as TokenBoxShadowValue` can you just specify a return type here?

six7 (changes requested):

six7 (approved):

PR #1583, opened 2023-02-01
Second screen

roppazvan:


src/app/store/models/tokenState.tsx
```diff
@@ -105,6 +105,14 @@ export const tokenState = createModel<RootModel>()({
       ...state,
       activeTokenSet: data,
     }),
+    setUsedTokenSet: (state, data: UsedTokenSetsMap) => ({
```
six7, line 108: Why was this changed?

src/types/SecondScreen.ts
```diff
@@ -0,0 +1,25 @@
+import { ThemeObject } from './ThemeObject';
+import { SingleToken } from './tokens';
+import { UsedTokenSetsMap } from './UsedTokenSetsMap';
+
+export type Token = SingleToken;
```
six7, line 5: Why not use SingleToken in the first place?

src/types/SecondScreen.ts
```diff
@@ -0,0 +1,25 @@
...
+export enum Clients {
+  PLUGIN = 'plugin',
+  SECOND_SCREEN = 'second-screen',
+}
+
+export interface TokenData {
+  id: string | number | null;
+  owner_email?: string;
+  created_at?: string;
+  ftData: {
```
six7, line 18: Can we change this? We are no longer called Figma Tokens

src/types/SecondScreen.ts
```diff
@@ -0,0 +1,25 @@
...
+export type TokenSets = Record<string, Token[]>;
+
+export enum Clients {
+  PLUGIN = 'plugin',
+  SECOND_SCREEN = 'second-screen',
+}
+
+export interface TokenData {
+  id: string | number | null;
+  owner_email?: string;
```
six7, line 16: When do we use snake_case and when camelCase?

src/types/Auth.ts
```diff
@@ -0,0 +1,32 @@
...
+  expires_in: number;
+  refresh_token: string;
+  token_type: string;
+  user: UserData;
+}
+
+export interface AuthInfo {
+  email?: string;
+  password?: string;
+  repeatedPassword?: string;
```
six7, line 20: Again feels weird to mix cases like that

src/context/AuthContext.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setAuthInProgress(false);
+  }, []);
+
+  const signUp = useCallback(async (signUpInfo: AuthInfo) => {
+    setAuthInProgress(true);
+    const { data, error } = await supabase.signUp(signUpInfo);
+    if (error) {
+      setAuthError(error.msg || error.error_description);
+    } else {
+      handleLogin(data);
```
six7, line 70: Can you add mixpanel track() events for all these?

src/app/components/SecondScreen/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const secondScreenButton = (
+    <Box
+      css={{
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'space-between',
+        gap: '$3',
+        padding: '$2 $4',
+        background: isEnabled ? 'rgba(65, 63, 63, 1)' : 'transparent',
```
six7, line 38: Can you use a token here? We need to support both light and dark mode, we cannot use rgb values

src/app/components/SecondScreen/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+        borderRadius: '6px',
```
six7, line 39:
```suggestion
        borderRadius: '$button',
```

src/app/components/AuthModal/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [mode, setMode] = useState<AuthModes>(AuthModes.LOGIN);
+  const secondScreenEnabled = useSelector(secondScreenSelector);
+
+  const [values, setValues] = React.useState({
+    email: '',
+    password: '',
+  });
+
+  const handleChange = useCallback(
+    (e: any) => {
```
six7, line 32: Can we remove the any?

src/app/components/AuthModal/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      </>
+    );
+  };
+
+  return (
+    <Modal close={dispatch.uiState.toggleSecondScreen} isOpen={secondScreenEnabled && !user}>
+      <Box css={{ display: 'flex', flexDirection: 'column', gap: '$5' }}>
+        <Input name="email" type="email" value={values.email} onChange={handleChange} label="Email" />
+        <Input name="password" type="password" value={values.password} onChange={handleChange} label="Password" />
+        <Button variant="primary" onClick={onSubmitButtonClick}>
```
six7, line 86: Can we still integrate a Forgot password link?

six7 (changes requested):
Nice!

src/app/store/models/tokenState.tsx
roppazvan: I needed a way to update the `usedTokenSet` field in the state with the data from second-screen. And there was no way to do that in a straightforward way. same for themes.

src/types/SecondScreen.ts
roppazvan: snake case is the recommended way for posgtres . Camel case is for the actual data inside ftData, I'll rename these to be more consistent
Screenshot 2023-02-03 at 10 17 11

src/types/Auth.ts
roppazvan: yeah, this is a mistake. theres no refresh_token there.

PR #1584, opened 2023-02-02
1203 refactor the tokenpicker component

swordEdge:
https://www.loom.com/share/2153ad7b83bb49e3b0b0e970016e025e


six7 (changes requested):
in the dropdown we need to be able to use `arrow up / down` to navigate the list, and use `enter` to select it.

when clicking outside the picker the picker should close

when clicking `esc` the picker should close (not the modal, the modal should close when there's no picker open)

the list of `Fonts` feels a bit slow when hovering or typing, lets optimize performance

when there's no tokens to suggest in `Tokens`, let's show a text `No suggestions found` (padding: 3, color: fgMuted, fontSize: small)

when there's just the `Tokens` (no Font), let's drop the word `Tokens` and start with the search input, feels cleaner

when the window is smaller, make sure the popover never grows beyond screen size so this can be avoided:
https://user-images.githubusercontent.com/4548309/218258926-92de5fb5-4973-4f56-9579-4668ad5c7941.mp4

the `mention` doesnt seem to trigger when i type `rgba({` - it seems to require a space in the beginning or be the first word. lets also support it when typed in `rgba({`

six7:
@swordEdge nice improvements! 🎉

one thing i noticed on this current build is that when using the `Load preset` and you create a new spacing token, the suggestions sometimes show as `empty lines`:

image

swordEdge:
> @swordEdge nice improvements! 🎉
> 
> one thing i noticed on this current build is that when using the `Load preset` and you create a new spacing token, the suggestions sometimes show as `empty lines`:
> 
> image

@six7 Fixed it.

six7:
@swordEdge is the dependency on fuse.js required? I'm asking because it doesn't have MIT license. Is there a way around it?

six7 (approved):

PR #1591, opened 2023-02-06
82 if the license key is invalid you cannot remove it

roppazvan:
* The remove license key should be there even if the license key is invalid
* The add license key will be there only if theres no key added
* Fixed tests to reflect the new functionality

### DIFF
--- a/src/app/components/AddLicenseKey/AddLicenseKey.test.tsx
+++ b/src/app/components/AddLicenseKey/AddLicenseKey.test.tsx
@@ -6,7 +6,9 @@ import {
   createMockStore, fireEvent, render, resetStore, screen, waitFor,
 } from '../../../../tests/config/setupTest';
 import AddLicenseKey from './AddLicenseKey';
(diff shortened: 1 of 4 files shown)


six7:
One thing that is still not optimal:

- Have no license key entered
- Enter a wrong license key, e.g. `asdf`
- Notice the plugin gives validation error
- "Remove key" is shown. But: I can't remove it!

I think ideally when the license key wasn't valid, the `Remove key` button should just remove it, without even showing the modal to confirm (given it was a wrong one, I dont need to save it). If the user removes the key without a validation error, we need to still show the modal.

six7 (changes requested):
One more bit weird scenario:

When you add a random key that does not work, e.g. `asdf`, (so the validation error says No license was found)
and you then change the contents of the input,
the remove key is disabled, until you actually type `asdf` again.

We should always allow you to remove it, it shouldnt have to match the previously entered key

roppazvan:
Fixed the remove key button disabled. Indeed it was a bug

six7 (approved):

PR #1598, opened 2023-02-08
fix tokenset tree breaking

swordEdge:


six7 (changes requested):
Can you also fix the performance of the tree? It feels really slow right now, every time an item moves it seems to stutter for a bit there.

You could also refactor the tree component itself, it still has some logic in it about the `TokenSetListOrTree` - we *always* use the Tree since a few versions ago, no need for the List anymore.

https://user-images.githubusercontent.com/4548309/218257833-5670c8d0-6212-42ef-bea8-f1ad43cbf515.mp4

src/app/components/ManageThemesModal/CreateOrEditThemeForm.tsx
```diff
@@ -102,7 +102,6 @@ export const CreateOrEditThemeForm: React.FC<Props> = ({
       {activeTab === ThemeFormTabs.SETS && (
         <Box css={{ paddingTop: '$4' }}>
           <TokenSetListOrTree
```
six7, line 104: The name of this component also no longer makes sense as all we have is a Tree, no `or tree`

src/app/components/TokenSetListOrTree/TokenSetListOrTree.tsx
```diff
@@ -17,7 +17,6 @@ type TreeOrListItem<ItemType = unknown> = {
 } & ItemType;
 
 type SharedProps<T extends TreeOrListItem> = {
```
six7, line 19: the name of this type as well

src/app/components/TokenSetListOrTree/TokenSetListOrTree.tsx
```diff
@@ -26,7 +25,6 @@ type SharedProps<T extends TreeOrListItem> = {
 type Props<T extends TreeOrListItem> = SharedProps<T>;
```
six7, line 25: here too

src/app/components/TokenSetListOrTree/TokenSetListOrTree.tsx
```diff
@@ -26,7 +25,6 @@ type SharedProps<T extends TreeOrListItem> = {
...
 
 export function TokenSetListOrTree<T extends TreeOrListItem>({
```
six7, line 27: here too

six7 (changes requested):

src/app/components/TokenSetListOrTree/TokenSetListOrTree.tsx
swordEdge: @six7 Could you merge this pr if functionality works. Test coverage is still failed but it requires me to test drag function. You know grab and drag.

six7 (approved):

PR #1600, opened 2023-02-09
fix nested referece broking

swordEdge:
https://www.loom.com/share/c416851ba7884210920f2940467ebeee


six7 (changes requested):
can you add a test case for that specific scenario i was mentioning in the issue?

six7 (approved):

PR #1605, opened 2023-02-10
1577-support-sorting-of-themes

swordEdge:
@six7 Updated it.
https://www.loom.com/share/25e659a5a24a438a9020692489e6ec14


src/selectors/themeObjectsSelector.ts
```diff
@@ -3,5 +3,5 @@ import { tokenStateSelector } from './tokenStateSelector';
 
 export const themeObjectsSelector = createSelector(
   tokenStateSelector,
-  (state) => state.themes,
+  (state) => state.themes.sort((a, b) => a.name.localeCompare(b.name)),
```
six7, line 6: instead of sorting by name, can we allow users to sort them themselves? like have the same grab handle we have for `Sets` to the left of their name when hovered. and then we'd just remember that sort.

src/selectors/themeOptionsSelector.ts
```diff
@@ -7,6 +7,6 @@ export const themeOptionsSelector = createSelector(
     Object.values(state.themes).map(({ id, name }) => ({
       value: id,
       label: name,
-    }))
+    })).sort((a, b) => a.label.localeCompare(b.label))
```
six7, line 10: see above

src/selectors/themesListSelector.ts
```diff
@@ -3,5 +3,5 @@ import { tokenStateSelector } from './tokenStateSelector';
 
 export const themesListSelector = createSelector(
   tokenStateSelector,
-  (state) => state.themes,
+  (state) => state.themes.sort((a, b) => a.name.localeCompare(b.name)),
```
six7, line 6: see above

six7 (changes requested):

six7 (changes requested):
can you move the reorder to the `Manage themes` modal? Let's keep the dropdown clean and rather have that functionality in the modal.

six7:
tests are failing @swordEdge

six7:
@swordEdge just some conflicting files

six7 (approved):

PR #1611, opened 2023-02-14
1045 ability to add line height token in rem units

MightyDevKnight:
https://user-images.githubusercontent.com/102773603/218662952-0fa4609d-ce1d-4f01-a90b-d901d01eaa3c.mp4

### DIFF
--- a/src/plugin/figmaTransforms/lineHeight.ts
+++ b/src/plugin/figmaTransforms/lineHeight.ts
@@ -4,7 +4,7 @@ export function convertLineHeightToFigma(inputValue: string, baseFontSize: strin
   let lineHeight: LineHeight | null = null;
   const value = inputValue.toString();
   const numbers = /^\d+(\.\d+)?$/;
-  if (value.match(numbers) || value.endsWith('px')) {
+  if (value.match(numbers) || value.endsWith('px') || value.endsWith('rem')) {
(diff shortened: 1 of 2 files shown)


src/plugin/figmaTransforms/lineHeight.ts
```diff
@@ -14,6 +14,11 @@ export function convertLineHeightToFigma(inputValue: string, baseFontSize: strin
       unit: 'PERCENT',
       value: Number(value.slice(0, -1)),
     };
+  } else if (value.match(numbers)) {
+    lineHeight = {
+      unit: 'PERCENT',
```
six7, line 19: If the value is a number it should be treated like a px value like currently

six7 (changes requested):

src/plugin/figmaTransforms/lineHeight.ts
MightyDevKnight: @six7 Please review this pic
image

six7: yes, lets not do this now. changing it means all tokens of users become changed.

MightyDevKnight: okay

six7 (approved):

PR #1623, opened 2023-02-17
fix removing extension in token transformer

swordEdge:


token-transformer/output/dark.json
```diff
@@ -79,7 +79,14 @@
     "backgroundDark": {
       "value": "#949DA7",
       "description": "Darkened color value",
-      "type": "color"
+      "type": "color",
+      "$extensions": {
```
six7, line 83: the $extensions modify object should only be there when `resolveReferences` = false

src/utils/convertTokensToGroupedObject.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const resolvedTokens = [
+      {
+        name: 'color.first',
+        value: '#ffffff',
+        type: TokenTypes.COLOR,
+      },
+      {
+        name: 'color.modify',
+        value: '#ffffff',
+        $extensions: {
```
six7, line 16: only when `resolveReferences=false` should the $extensions object be kept

six7 (changes requested):

six7 (approved):

PR #1625, opened 2023-02-17
1338 make it possible to rename a tokenset with sub levels

MightyDevKnight:
https://user-images.githubusercontent.com/102773603/219541981-54feb051-dea6-4a63-9fa0-7e85c9f9ff8d.mp4


src/app/components/TokenSetSelector.tsx
```diff
@@ -106,7 +106,11 @@ export default function TokenSetSelector({ saveScrollPositionSet }: { saveScroll
 
   const handleRenameTokenSetSubmit = React.useCallback((e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
-    dispatch.tokenState.renameTokenSet({ oldName: tokenSetMarkedForChange, newName: newTokenSetName.trim() });
+    Object.keys(tokens).forEach((key) => {
+      if (!key.indexOf(tokenSetMarkedForChange)) {
+        dispatch.tokenState.renameTokenSet({ oldName: key, newName: key.replace(tokenSetMarkedForChange, newTokenSetName.trim()) });
```
six7, line 111: instead of iterating over all sets here, can we create an action in `tokenState`? So that we call `tokenState.renameTokenSetFolder` with the folder path and then it handles the logic for us? I'd like to keep logic like this out of the components

six7 (changes requested):

six7 (changes requested):
Right now if you have the following token sets

```
light/base
light/foo
lighter/base
```

and you rename `light` to `asdf`

it will also rename the `lighter` token set (because string matches), which it shouldnt

src/app/store/models/reducers/tokenState/renameTokenSetFolder.ts
```diff
@@ -0,0 +1,27 @@
...
+export function renameTokenSetFolder(state: TokenState, data: Payload): TokenState {
+  if (
+    Object.keys(state.tokens).includes(data.newName)
+    && data.oldName !== data.newName
+  ) {
+    notifyToUI('Folder name already exists', { error: true });
+    return state;
+  }
+
+  return updateTokenSetsInState(
```
six7, line 19: nice! can you add tests for this file still? should be isolated and easy to test 👍

src/app/store/models/reducers/tokenState/renameTokenSetFolder.ts
```diff
@@ -0,0 +1,27 @@
...
+    state,
+    (setName, tokenSet) => (
+      setName.includes(data.oldName) && setName[data.oldName.length] === '/'
```
six7, line 22: can you explain what this does? wouldnt that mean that it would still rename other sets as long as they have children (which we dont want, we just want to rename the selected set)

e.g. 

```
light/base
lighter/base
```
?

six7 (changes requested):

src/app/store/models/reducers/tokenState/renameTokenSetFolder.ts
```diff
@@ -0,0 +1,27 @@
+import { notifyToUI } from '@/plugin/notifiers';
+import { updateTokenSetsInState } from '@/utils/tokenset/updateTokenSetsInState';
+import type { TokenState } from '../../tokenState';
+
+type Payload = {
+  oldName: string
+  newName: string
+};
+
+export function renameTokenSetFolder(state: TokenState, data: Payload): TokenState {
```
six7, line 10: Can you add a unit test for this like tokenState.test.ts ?

Including the example I gave in https://github.com/tokens-studio/figma-plugin/pull/1625#pullrequestreview-1316248870

six7 (changes requested):

six7 (approved):

PR #1640, opened 2023-02-24
Fix color modifier not working correctly

swordEdge:


six7 (approved):
I think for now this is OK, but it's not ideal. We're basically making the `getAliasValue` function more complex than it needs to be. Can we refactor this later? For now I'll release given we want this fixed, but I feel this isn't optimal.

PR #1646, opened 2023-02-27
Fix supabase subscribe issue

roppazvan:
* In some cases an infinite loop was caused, and the connection lost.

### DIFF
--- a/src/app/components/SecondScreenSync/index.tsx
+++ b/src/app/components/SecondScreenSync/index.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useEffect } from 'react';
+import * as Sentry from '@sentry/react';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


src/app/components/SecondScreenSync/index.tsx
```diff
@@ -99,9 +98,11 @@ export default function SecondScreenSync() {
 
       channel.subscribe(async (status) => {
         if (status === 'SUBSCRIBED' && channel) {
-          await channel.track({ online_at: new Date().toISOString() });
-        } else if (status === 'CLOSED') {
-          createChannel();
+          try {
+            await channel.track({ online_at: new Date().toISOString() });
+          } catch (error) {
+            console.log('channel error', error);
```
six7, line 104: Should we rather track that with Sentry?

six7 (approved):

PR #1648, opened 2023-02-28
duplicate token across sets

swordEdge:
https://www.loom.com/share/b31faf78475f4272a1610263a650df88


src/app/components/EditTokenForm.tsx
```diff
@@ -243,6 +246,7 @@ function EditTokenForm({ resolvedTokens }: Props) {
   );
 
   // @TODO update to useCallback
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
six7, line 249: can you add a comment in code why this was disabled? otherwise we wont remember in a few months why its there and remove it + change the deps

six7 (changes requested):

six7 (approved):

PR #1657, opened 2023-03-01
 feat: add support expressions for additional math logic

SorsOps:
Adds the following functions to the underlying math layer 

`clamped`  - Note not `clamp` as this would clash with the existing css expression which might be being used
`lerp` - Linear interpolation
`norm`- Normalization of a value between a scale
`cubicBezier1D` -  A cubic bezier that operats in a single dimension. Again changed to reflect its 1D nature and to not collide with the css cubic-bezier expression
`sample` - Samples a function. Currently only supports the cubicBezier, but is intended for first class function support in the future 

Notes :

The functions themselves are unitless. Units cannot be added directly in the expression at the moment due to the internals of expr-eval not supporting string manipulation nor sufficient support for unary expressions to convert the unit into an object that can be manipulated. Trying to assign to the `unaryOps` variable on the parser whilst also enabling it seemed to have not effect. 

This means that units need to be added through the use of an intermediate token eg 

x = `sample(cubicBezier1D(1,0),0.5)`
y = `{x}px`


six7 (commented):
Looks good! One thing I wonder.. why `clamped` and not `clamp`? If someone's been using clamp it'd still work for them, and they'd be able to decide how they want to transform their output (e.g. they could leave clamp() in place if they're using css?

SorsOps:
I set it as `clamped` and not `clamp` in case someone was using the existing css clamp function which supports units whereas the `clamped` function is unitless. Unfortunately the `expr-eval` lib does not seem to support custom unary ops so I can't get it to support units like `px` , etc using some kind of wrapper.

So since it might cause a regression I changed the name

src/utils/math/checkAndEvaluateMath.ts
```diff
@@ -24,13 +75,13 @@ export function checkAndEvaluateMath(expr: string) {
   let evaluated: number;
 
   try {
-    evaluated = parser.evaluate(unitlessExpr);
+    evaluated = parser.evaluate(`${unitlessExpr}`);
```
six7, line 78: Out of interest why is that change needed?

jorenbroekema:
Mind if I adopt this PR to https://github.com/tokens-studio/sd-transforms ?

The `checkAndEvaluateMath` function can then be imported in this project:
```js
import { checkAndEvaluateMath } from '@tokens-studio/sd-transforms';
```

My checkAndEvaluateMath already supports multi-value values as well so I'm assuming it's a bit ahead of this one now, just judging by the tests.

six7:
> Mind if I adopt this PR to https://github.com/tokens-studio/sd-transforms ?
> 
> The `checkAndEvaluateMath` function can then be imported in this project:
> 
> ```js
> import { checkAndEvaluateMath } from '@tokens-studio/sd-transforms';
> ```
> 
> My checkAndEvaluateMath already supports multi-value values as well so I'm assuming it's a bit ahead of this one now, just judging by the tests.

just bumping this for @SorsOps - i think it'd be great to combine those. 

i think we can release this as `clamped` - if we later find a way to release a proper `clamp` that supports units, we're not blocked by having the same naming as that fn.

let me know if you'd like to consolidate it as Joren suggested, would love to release early of the week (as a silent release)

SorsOps:
Yeah that would be great, especially as I'd also like to test this for the tokens-studio API.

I'm still using `token-transformer` but I'd like to move over to `@tokens-studio/sd-transforms` for a better local token resolution experience

PR #1665, opened 2023-03-06
Fixes second-screen data sync issues

roppazvan:
* Because of multiple dispatches the sync component sent updates too often to second screen & there were some issues with inline editing because of this.
* Moved the updates to a single action & dispatch


roppazvan:
@six7 Can you have a look ? Thanks

PR #1672, opened 2023-03-08
1671 base font size token doesnt pick up tokens

swordEdge:
https://user-images.githubusercontent.com/25951419/223715617-4b34653c-3dd1-4e94-aaae-cc92fb0fcedd.mp4

### DIFF
--- a/src/app/store/models/settings.tsx
+++ b/src/app/store/models/settings.tsx
@@ -196,7 +196,7 @@ export const settings = createModel<RootModel>()({
     },
     setBaseFontSize: (payload, rootState) => {
       setUI(rootState.settings);
-      dispatch.tokenState.updateDocument({ shouldUpdateNodes: true });
+      dispatch.tokenState.updateDocument({ shouldUpdateNodes: false });
     },
     setAliasBaseFontSize: (payload, rootState) => {
       setUI(rootState.settings);
--- a/src/app/store/models/settings.tsx
+++ b/src/app/store/models/settings.tsx
@@ -155,7 +155,7 @@ export const settings = createModel<RootModel>()({
       };
     },
   },
-  effects: () => ({
+  effects: (dispatch) => ({
     setWindowSize: (payload) => {
       AsyncMessageChannel.ReactInstance.message({
         type: AsyncMessageTypes.RESIZE_WINDOW,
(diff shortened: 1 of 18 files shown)


src/plugin/node.ts
```diff
@@ -221,14 +221,23 @@ async function migrateTokens(entry: NodeManagerNode, values: MapValuesToTokensRe
...
-  const { ignoreFirstPartForStyles, prefixStylesWithThemeName, baseFontSize } = settings ?? {};
+  const { ignoreFirstPartForStyles, prefixStylesWithThemeName, aliasBaseFontSize } = settings ?? {};
   const figmaStyleMaps = getAllFigmaStyleMaps();
   const themeInfo = await AsyncMessageChannel.PluginInstance.message({
     type: AsyncMessageTypes.GET_THEME_INFO,
   });
-
+  let nameToLookFor = '';
+  if (aliasBaseFontSize) {
+    if (aliasBaseFontSize.startsWith('{')) {
```
six7, line 233: shouldnt we already have the value of `aliasBaseFontSize` in settings? you query on line 197 of Tokens.tsx the aliasValue? That should be the real value to be used, right? No need to further query the alias value of it'd already be in store?

six7 (changes requested):

src/plugin/node.ts
swordEdge: > shouldnt we already have the value of `aliasBaseFontSize` in settings? you query on line 197 of Tokens.tsx the aliasValue? That should be the real value to be used, right? No need to further query the alias value of it'd already be in store?

@six7 Yes. We store the real value to be used. But when we switch the theme, it causes the document update. But at this time, the real value has not been updated yet. It will be updated after this process. That's why I resolved the aliasValue here.

six7: Then dont we have a function to getAliasVallue? I feel we shouldnt have alias resolve logic in node.ts - we should import a function that does it.

Or even better, make sure that by the time the document updates the base font size token has been updated

src/app/store/models/settings.tsx
```diff
@@ -196,6 +196,7 @@ export const settings = createModel<RootModel>()({
     },
     setBaseFontSize: (payload, rootState) => {
       setUI(rootState.settings);
+      dispatch.tokenState.updateDocument({ shouldUpdateNodes: true });
```
six7, line 199: can you set this to `shouldUpdateNodes: false`?

We wouldnt want a change in the base font size trigger a document wide update, which could mean the users would have to wait a long time. So instead lets just update the document, but not the nodes.

six7 (changes requested):

PR #1688, opened 2023-03-15
Address code vulnerabilities

SorsOps:
Optimize pipelines
Fix npm audit problems
Reshuffle devDeps in prod deps
Update some modules according to snyk


SorsOps:
Note that the two remaining high priority vulnerabilities are minimatch and set-value

There is a major version 4 of set-value that could fix the vulnerability detected, however this is a major change. Naive installation of the new major version causes regression

SorsOps:
Note this will cause a failure in the CI 

```
      - name: Run test coverage report
        id: testCoverage
        uses: anuraag016/Jest-Coverage-Diff@master
        with:
          fullCoverageDiff: false
          runCommand: 'LAUNCHDARKLY_FLAGS=tokenThemes,gitBranchSelector,multiFileSync,tokenFlowButton npx jest --collectCoverageFrom=''["src/**/*.{js,jsx,ts,tsx}"]'' --coverage --collectCoverage=true --coverageDirectory=''./'' --coverageReporters=''json-summary'' --forceExit --detectOpenHandles'
          total_delta: 1
          delta: 1
          afterSwitchCommand: npm ci
          useSameComment: true
          ````
          
          It attempts to pull the current main and then run npm ci. Until this is merged in, main will use yarn, We don't want this to stay yarn as it will cause failures for other PRs once this is in

six7 (approved):
🙏

PR #1700, opened 2023-03-18
Introduce Monaco JSON editor

swordEdge:


src/AsyncMessageChannel.ts
```diff
@@ -63,7 +63,7 @@ export class AsyncMessageChannel {
 
   public connect() {
     return this.attachMessageListener(async (msg: { id?: string; message?: AsyncMessages }) => {
-      if (!msg.id || !msg.message || !msg.message.type.startsWith('async/')) return;
+      if (!msg?.id || !msg.message || !msg.message.type.startsWith('async/')) return;
```
six7, line 66: shouldn't `msg` always be present?

src/app/components/JSONEditor.tsx
```diff
@@ -1,24 +1,39 @@
+/*
```
six7, line 1: we can remove those comments

src/app/components/JSONEditor.tsx
```diff
@@ -1,24 +1,39 @@
...
   handleChange: (tokens: string) => void;
-  hasError: boolean;
 };
 
 function JSONEditor({
   stringTokens,
   handleChange,
-  hasError,
 }: Props) {
-  const editProhibited = useSelector(editProhibitedSelector);
```
six7, line 20: we will still need `editProhibited` for the case where users are in read-only mode

src/app/components/JSONEditor.tsx
```diff
@@ -1,24 +1,39 @@
...
   const { handleJSONUpdate } = useTokens();
+
+  const handleJsonEditChange = React.useCallback((value?: string) => {
+    if (value) handleChange(value);
```
six7, line 34: is there a case where value is null? shouldnt the editor *always* have a value? even when its `""` ?

src/app/components/JSONEditor.tsx
```diff
@@ -37,13 +52,11 @@ function JSONEditor({
...
-        isDisabled={editProhibited}
-        placeholder="Enter JSON"
-        rows={21}
-        onChange={handleChange}
+      <Editor
+        language="json"
+        onChange={handleJsonEditChange}
         value={stringTokens}
-        css={{ paddingBottom: hasError ? '$9' : '0' }}
+        theme="vs-dark"
```
six7, line 59:
```suggestion
        theme="vs-dark"
        options={{
          minimap: {
            enabled: false,
          },
          lineNumbers: 'off',
          wordWrap: 'on',
          contextmenu: false,
        }}
```

six7 (changes requested):

src/app/components/JSONEditor.tsx
swordEdge: The type of parameter of onChange function in Moncaco editor is string | undefined.

six7 (approved):
🙌

PR #1714, opened 2023-03-22
preserve resolve reference border token

swordEdge:


six7 (changes requested):
Can you also add a token like that to the token-transformer tests? Given this was an issue created around that. Just one border token in the tokens.json file for example

six7 (approved):

PR #1717, opened 2023-03-22
1673-wrong-order-in-metadata-tokensetorder

swordEdge:


src/app/components/TokenSetTree.tsx
```diff
@@ -43,12 +45,20 @@ type TokenSetTreeProps = {
...
   tokenSets, onReorder, onRename, onDelete, onDuplicate, saveScrollPositionSet,
 }: TokenSetTreeProps) {
+  const tokens = useSelector(tokensSelector);
   const activeTokenSet = useSelector(activeTokenSetSelector);
   const usedTokenSet = useSelector(usedTokenSetSelector);
   const editProhibited = useSelector(editProhibitedSelector);
   const collapsed = useSelector(collapsedTokenSetsSelector);
   const items = tokenSetListToTree(tokenSets);
 
+  React.useEffect(() => {
```
six7, line 55: Wouldn't a better place to fix this bug be where we create and rename token sets? So we make sure that it is already inserted into the right place when we create token sets? Instead of relying on the UI to do that as a side effect?

six7 (changes requested):

src/app/components/TokenSetTree.tsx
swordEdge: @six7 We can fix this bug in the place where we create and rename token sets. But how about the users with old data like the user who mentioned this issue. He will get this error even though after fixing. Because tokenSets are ordered when we add or rename. What's your opinion? Still can we fix in the place where we create and rename token sets?

six7: I agree, this sounds like a good solution that also fixes older sets of users.

Can you still additionally add the logic to insert in the right place when creating and renaming? That way we do not have to rely on the UI to do the magic for us.

six7 (approved):

PR #1757, opened 2023-03-27
1697-ado-on-initial-setup-the-tokensjson-file-is-not-created-resulting-in-an-error

swordEdge:
https://user-images.githubusercontent.com/25951419/227830240-9285ad22-0926-4192-94db-08ad46dae912.mp4

### DIFF
--- a/src/storage/ADOTokenStorage.ts
+++ b/src/storage/ADOTokenStorage.ts
@@ -321,6 +321,9 @@ export class ADOTokenStorage extends GitTokenStorage {
           })),
         ];
       }
+      if (/could not be found in the repository/i.test(singleItem.message)) {
+        return [];
+      }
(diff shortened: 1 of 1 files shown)


src/storage/ADOTokenStorage.ts
```diff
@@ -321,6 +321,9 @@ export class ADOTokenStorage extends GitTokenStorage {
           })),
         ];
       }
+      if (/could not be found in the repository/i.test(singleItem.message)) {
```
six7, line 324: Any way we can make this not dependent on the actual text used but rather an error code? If they change their error messages this won't work anymore

six7 (changes requested):

six7 (approved):
✨

PR #1784, opened 2023-04-03
1609 show diff before pushing and pulling changes

swordEdge:


src/app/components/ChangedStateList.tsx
```diff
@@ -0,0 +1,133 @@
...
+                  gap={1}
+                  css={{ padding: '$2 $4' }}
+                >
+                  <Text bold size="small">{theme.name}</Text>
+                  {
+                    theme.importType === 'REMOVE' && (
+                    <Text
+                      size="small"
+                      css={{
+                        padding: '$2',
```
six7, line 57: i think it could make it a bit cleaner to move this into a StyledText and then reuse that one with a `variant` prop of either `success` or `danger`

src/app/components/ChangedTokenItem.tsx
```diff
@@ -0,0 +1,105 @@
...
+                fontSize: '$xsmall',
+                backgroundColor: '$bgDanger',
+                color: '$fgDanger',
+              }}
+              >
+                {token.oldDescription}
+              </Box>
+            ) : null}
+            <Box css={{
+              padding: '$2',
```
six7, line 70: same here, theres lots of these success / danger text items that could be a reusable component instead

src/app/components/ChangeStateListingHeading.tsx
```diff
@@ -0,0 +1,44 @@
...
+  tokenKey,
+  onCollapse,
+}: Props) {
+  const handleCollapse = React.useCallback((e: React.MouseEvent<HTMLButtonElement>) => {
+    onCollapse(e, tokenKey);
+  }, [onCollapse, tokenKey]);
+
+  return (
+    <Stack direction="row" align="center" justify="between" gap={4} css={{ position: 'relative' }}>
+      <StyledTokenGroupHeadingButton
```
six7, line 30: i think it could be better not to reuse the token group heading here and instead have its own component? if we later decide we want to change the token group heading component this could have unexpected implications here

src/app/components/PullDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <Modal
+          title={`Pull from ${transformProviderName(storageType.provider)}`}
+          showClose
+          large
+          isOpen
+          close={onCancel}
+        >
+          <Stack direction="column" gap={4}>
+            <Stack direction="row" gap={2}>
+              This will override your current tokens. Make sure you copy your changes if you want to preserve theme.
```
six7, line 41: slight typo: should be `preserve them`

src/app/components/PullDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return (
+        <Modal
+          large
+          isOpen
+          close={onCancel}
+          title={`Pull from ${transformProviderName(storageType.provider)}`}
+        >
+          <Stack direction="column" gap={4} justify="center" align="center">
+            <Stack direction="row" gap={2}>
+              This will override your current tokens. Make sure you copy your changes if you want to preserve theme.
```
six7, line 73: can we change this so it only changes the part that need to be changed for Loading ? right now it seems to be duplicate content.

Also, can we remove this sentence from the loading state? I know it was in the initial designs, but from testing it it feels like it's better if it doesn't show it during Loading

src/app/components/PushJSON.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <Stack
+      direction="column"
+      gap={1}
+    >
+      {Object.entries(tokens).length > 0 && Object.entries(tokens)?.map(([tokenSet, tokenList]) => (
+        tokenList.length > 0 && (
+        <>
+          <ChangeStateListingHeading onCollapse={handleSetIntCollapsed} tokenKey={tokenSet} label={tokenSet} isCollapsed={collapsedTokenSets.includes(tokenSet)} />
+          {!collapsedTokenSets.includes(tokenSet) && tokenList && (
+            <pre style={{ padding: '$2 $4' }}>
```
six7, line 34: one thing i noticed here in the JSON (either of them) is that when you have a very long description it currently will not wrap and will lead to horizontal scroll. we should make it so that text wraps

(to reproduce load the example Preset (Tools > Load preset) and then push this

Image

src/app/components/PushJSON.tsx
```diff
@@ -0,0 +1,70 @@
...
+              {stringifyTokens(tokens, tokenSet)}
+            </pre>
+          )}
+        </>
+        )
+      ))}
+      {
+        themes.length > 0 && (
+          <>
+            <ChangeStateListingHeading onCollapse={handleSetIntCollapsed} tokenKey="$themes" label="$themes" isCollapsed={collapsedTokenSets.includes('$themes')} />
```
six7, line 44: one thing that doesnt yet work is when i `Alt + Click` on an expanded token set i would expect that **all** token sets would collapse. much like when i do that on a collapsed on all expand

src/app/store/remoteTokens.tsx
```diff
@@ -65,11 +74,7 @@ export default function useRemoteTokens() {
     context = api, featureFlags, usedTokenSet, activeTheme, collapsedTokenSets,
   }: PullTokensOptions) => {
     track('pullTokens', { provider: context.provider });
-    dispatch.uiState.startJob({
-      name: BackgroundJobs.UI_PULLTOKENS,
-      isInfinite: true,
-    });
-
+    pullDialog('loading');
```
six7, line 77: should this be renamed to `showPullDialog` ? to match what we have for `closePullDialog`? that way it's more descriptive

src/app/components/PullDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+              padding: '$4',
+              borderTop: '1px solid',
+              borderColor: '$borderMuted',
+            }}
+            >
+              <Button variant="secondary" id="pullDialog-button-close" onClick={handleClose}>
+                Cancel
+              </Button>
+              <Button variant="primary" id="pullDialog-button-override" onClick={handleOverrideClick}>
+                Override Tokens
```
six7, line 56:
```suggestion
                Override tokens
```

six7 (changes requested):
Very cool!

I added some comments.

A slightly bigger request: Can we change the `Push` logic so that it's not separate into 2 steps? Instead, can we just add a `Commit` tab to the list of tabs? And then we have `Commit | Diff | JSON` as Tabs. That way we save users a click when they want to push changes, but they still will be able to see their changes? And if the `branch` was changed, we'd retrigger the comparison for the diff/json in the tabs?

six7 (changes requested):
Nice! 1 minor thing:

image

the horizontal scroll is still there in the JSON tab when there's a really long text string in the JSON

swordEdge:
> Nice! 1 minor thing:
> 
> image
> 
> the horizontal scroll is still there in the JSON tab when there's a really long text string in the JSON

> On my side, it was fixed already. Could you try again on your side? @six7 
image

six7 (approved):
🔥

PR #1790, opened 2023-04-05
1058-allow-token-names-and-style-names-to-be-different-while-still-in-sync

swordEdge:
https://user-images.githubusercontent.com/25951419/230179449-e3f5797f-34e5-4bbe-9c6d-6632eca18054.mp4


six7 (changes requested):
One thing:

- It seems like it's now no longer possible to update styles if you're not using Themes. To reproduce: Load the preset, create styles from it, then change the values of the tokens (without ever creating a Theme). This should still be possible

Otherwise this is great! 🎉

six7 (approved):
Nice! 💖

PR #1793, opened 2023-04-07
1326-filtering-of-git-branches-in-the-selector

swordEdge:


six7:
Not sure if this is ready for review yet, but:

- Can we change the logic so it only displays the search bar if there's more than 5 branches? Right now it displays even when there's just 1

There seems to be some padding issue on the right of the second dropdown that appears. Let's make this look like the first

image

There seems to be another padding issue with the `branch` item compared to the `Create new branch from..`
image
(notice the blue background doesnt span to the right)
image

src/app/components/BranchSelector.tsx
```diff
@@ -184,12 +198,30 @@ export default function BranchSelector() {
...
   }, []);
 
+  const handleSearchTextForBaseBranchChange = React.useCallback<React.ChangeEventHandler<HTMLInputElement>>((e) => {
+    setSearchTextForBaseBranch(e.target.value);
+  }, []);
+
+  const handleSearchTextForCreateBranchChange = React.useCallback<React.ChangeEventHandler<HTMLInputElement>>((e) => {
+    setSearchTextForCreateBranch(e.target.value);
+  }, []);
+  // TODO: I'm not sure there is another way instead of stopImmediatePropagation to prevent option selection when typing the first letter of the option
```
six7, line 220: this seems to select the first item for me as soon as i start typing 

image

six7 (changes requested):

src/app/components/BranchSelector.tsx
six7: also got an error in console:

image

swordEdge: Did you upgrade the node module? This error is happening in old version so upgrated it.

swordEdge: What I'm not sure is that all test cases are passed with old version but failed with upgraded version.

six7: You're right, this is gone now 👍

six7: i'm still getting that error in the console, i did install, yes

image

six7 (changes requested):
when you set figma to light mode it currently displays the search field with wrong colors. you can just use raw hex values for these instead, as the context menu has the same colors both in light and dark mode. but create a token for that, so that it's stored in stitches.config

image

six7:
just want to check in on this one: is there anything i can help with? as far as i can see its just the error in the console that's still there, the rest is done 🙏

PR #1798, opened 2023-04-10
Json schema and backstage

SorsOps:
Add support for our json schema for autocomplete and adds backstage config

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,25 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: figma-plugin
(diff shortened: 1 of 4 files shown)


six7 (commented):
Not sure what I should focus on in this review. Should I see anything in Monaco?

SorsOps:
Hey @six7  I just wanted you to check that the behaviour with the intellisense of the schema fits with your idea of how it should work :)

six7 (approved):
Awesome, tested - thank you! Would be awesome to set the width of intellisense to something less wide. Right now it's not visible when your plugin window is more narrow

PR #1809, opened 2023-04-14
Supernova.io Integration

JiriTrecak:
This PR adds support for the Supernova data provider. The provider works analogous to every other git-type provider on the platform, minus the support for branches which is not a concept Supernova has.

An upcoming docs page for this integration [can be found here](https://learn.supernova.io/latest/design-systems/add-design-system-content/tokens-studio.html).

This integration is currently not testable with the production environments of Supernova (we need to merge prod API support for it first), but should be available in next few days as well. In a meantime, I'll share access for the development realm privately.

Start by adding a new provider:

tokens


src/app/store/remoteTokens.tsx
```diff
@@ -55,294 +70,336 @@ export default function useRemoteTokens() {
...
 
-  const pullTokens = useCallback(async ({
-    context = api, featureFlags, usedTokenSet, activeTheme,
-  }: PullTokensOptions) => {
-    track('pullTokens', { provider: context.provider });
-    dispatch.uiState.startJob({
-      name: BackgroundJobs.UI_PULLTOKENS,
-      isInfinite: true,
-    });
+  const pullTokens = useCallback(
```
six7, line 89: any chance we can avoid these large diffs as in code style changes? our `next` branch is quite different from `main`, so i'd like to avoid large merge conflicts due to these

src/app/store/remoteTokens.tsx
```diff
@@ -55,294 +70,336 @@ export default function useRemoteTokens() {
...
+        }
+        case StorageProviderType.BITBUCKET: {
+          content = await syncTokensWithBitbucket(context);
+          break;
+        }
+        case StorageProviderType.ADO: {
+          content = await syncTokensWithADO(context);
+          break;
+        }
+        case StorageProviderType.SUPERNOVA: {
```
six7, line 194: like ideally this file only adds the supernova stuff, not reshuffles whitespaces entirely (just purely from a merge conflict perspective), but also these should already be linted correctly

src/constants/ErrorMessages.ts
```diff
@@ -6,6 +6,8 @@ export enum ErrorMessages {
   ADO_CREDENTIAL_ERROR = 'Error syncing with ADO, check credentials',
   URL_CREDENTIAL_ERROR = 'Error fetching from URL, check console (F12)',
   JSONBIN_CREDENTIAL_ERROR = 'Error fetching from JSONbin, check console (F12)',
+  SUPERNOVA_CREDENTIAL_ERROR = 'Error syncing with Supernova, check credentials or your mapping configuration',
+  SUPERNOVA_NOT_UPLOADED_YET_ERROR = 'There were no tokens imported from Token Studio to Supernova previously, please push your tokens first',
```
six7, line 10:
```suggestion
  SUPERNOVA_NOT_UPLOADED_YET_ERROR = 'There were no tokens imported from Tokens Studio to Supernova previously, please push your tokens first',
```

src/icons/supernova.svg
```diff
@@ -0,0 +1,15 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g clip-path="url(#clip0_348_146)">
+<mask id="mask0_348_146" style="mask-type:luminance" maskUnits="userSpaceOnUse" x="0" y="0" width="16" height="16">
+<path d="M16 0H0V16H16V0Z" fill="white"/>
```
six7, line 4: we'd want to make sure this uses `currentColor` as we have to support both light and dark theme

src/storage/SupernovaTokenStorage.ts
```diff
@@ -0,0 +1,164 @@
...
+      const designSystemVersion = await designSystem!.activeVersion();
+      if (designSystem && designSystemVersion) {
+        return {
+          ds: designSystem,
+          version: designSystemVersion,
+        };
+      }
+    } catch (error) {
+      // Will throw always when something goes wrong
+      console.log(error);
```
six7, line 160: should we track these in Sentry? would help with any issues that might occur

webpack-transform.config.js
```diff
@@ -48,4 +48,4 @@ module.exports = (env, argv) => ({
     library: 'tokenTransformer',
     libraryTarget: 'commonjs2',
   },
-});
+});
```
six7, line 51:
```suggestion
});

```

six7 (changes requested):
haven't tested it yet, but here's a preliminary code review that surfaces a couple points. in general let's try to avoid code style changes as these will lead to merge conflicts down the road as we have other dev branches that touch a lot of this stuff as well

src/app/components/PushDialog.tsx
```diff
@@ -58,6 +59,15 @@ function PushDialog() {
         default:
           break;
       }
+    } else if (localApiState && 'designSystemUrl' in localApiState) {
```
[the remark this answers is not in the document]
JiriTrecak: Yes, this is good point. I have changed it :)

src/app/store/providers/supernova/supernova.tsx
```diff
@@ -0,0 +1,221 @@
...
+    }
+    return null;
+  }, [
+    storageClientFactory,
+  ]);
+
+  // Function to initially check auth and sync tokens with Supernova
+  const syncTokensWithSupernova = useCallback(
+    async (context: SupernovaCredentials): Promise<RemoteResponseData> => {
+      console.log('SYNC TOKENS');
```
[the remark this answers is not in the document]
JiriTrecak: Removed all I could find, I believe rest is there by default now

src/app/store/remoteTokens.tsx
JiriTrecak: ^ Same as above

src/constants/ErrorMessages.ts
JiriTrecak: Good catch - I also found second situation where Token(s) were used incorrectly, fixed both

six7 (approved):

PR #1816, opened 2023-04-17
Fix/change mention to rc mention

swordEdge:


six7 (approved):
👏 This is so much better, awesome job @swordEdge 🎉

PR #1824, opened 2023-04-21
1515 multi dimensional themes

swordEdge:


src/app/components/ThemeSelector/ThemeSelector.tsx
```diff
@@ -25,60 +25,110 @@ const ThemeDropdownLabel = styled(Text, {
...
+    return newArray;
+  }, [availableThemes]);
   const handleClearTheme = useCallback(() => {
-    dispatch.tokenState.setActiveTheme({ themeId: null, shouldUpdateNodes: true });
+    dispatch.tokenState.setActiveTheme({ newActiveTheme: {}, shouldUpdateNodes: true });
   }, [dispatch]);
 
   const handleSelectTheme = useCallback((themeId: string) => {
-    const nextTheme = (activeTheme === themeId) ? null : themeId;
+    const groupOfTheme = availableThemes.find((theme) => theme.value === themeId)?.group ?? 'noGroup';
```
six7, line 52: would this work if a user named their group `noGroup` ? should we use something here that might be more like `INTERNAL_THEMES_NO_GROUP` or so?

six7 (changes requested):
Amazing work on this, @swordEdge 👏 

Few things, no biggies:

## Reordering themes doesn't work anymore

Let's allow users to reorder themes inside the same theme group. We don't have to support moving to a different theme group for now (unless it's easy?)

## Allow reordering theme groups

Let's allow users to reorder whole theme groups so they can manage order

## Having a group should not be mandatory

Right now the `group` input is mandatory. Let's make it optional. If there's no group set, show a button `+ Add group`. Once clicked, this will turn into the group input field. This input should be optional, meaning you can save even if it is empty.

six7 (changes requested):
Nice. Few more things:

### Restarting the plugin should remember last used Themes

In the production version of the plugin we remembered the last used Theme and activated that. Let's bring this back, this seems to no longer work.

### Having a group should not be mandatory

I added this as a comment to the last review but I think it slipped through:

If there's no group set, show a button + Add group. Once clicked, this will turn into the group input field. This input should be optional, meaning you can save even if it is empty.

image

## Minor stuff
- Change `no group` to `No group`

six7 (approved):
🎉 Nice!

PR #1832, opened 2023-04-26
fix: allow "treat as source" without edit rights

mihkeleidast:
We use URL Sync, and with URL sync editing of the tokens is disabled inside Tokens Studio. However, currently, this also hides away the "Treat as source" functionality, which has nothing to do with editing the tokens, rather it's a special checked state of the tokenset. Since you can check/uncheck the tokensets even when you can't edit, I believe you should also be able to treat the tokenset as source.

PS. Sorry for jumping directly to a PR, seemed like a small enough change so we can discuss directly here if necessary?
PPS. There are some unrelated autoformatting changes in here, I can undo them manually, if necessary, but it's a bit cumbersome so I did not do that right now. Not sure what is going on with the prettier config in this repo, does not seem to be enforced.

### DIFF
--- a/src/app/components/TokenSetItem/TokenSetItem.tsx
+++ b/src/app/components/TokenSetItem/TokenSetItem.tsx
@@ -57,9 +57,7 @@ export function TokenSetItem({
   onTreatAsSource,
   onDragStart,
 }: TokenSetItemProps) {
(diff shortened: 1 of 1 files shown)


six7:
Awesome, thank you for this! 🙏

six7 (approved):

PR #1873, opened 2023-05-09
1612 when no styles were created due to ignore first part of token name for styles inform the user via notification

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/25951419/9e490f7e-de45-48c2-a9a9-d89253d7435d

### DIFF
--- a/src/plugin/updateStyles.ts
+++ b/src/plugin/updateStyles.ts
@@ -9,6 +9,7 @@ import updateColorStyles from './updateColorStyles';
 import updateEffectStyles from './updateEffectStyles';
 import updateTextStyles from './updateTextStyles';
 import { TokenSetStatus } from '@/constants/TokenSetStatus';
+import { notifyUI } from './notifiers';
 
 export default async function updateStyles(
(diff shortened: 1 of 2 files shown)


src/plugin/updateStyles.ts
```diff
@@ -53,5 +54,9 @@ export default async function updateStyles(
     ...(textTokens.length > 0 ? updateTextStyles(textTokens, settings.baseFontSize, shouldCreate) : {}),
     ...(effectTokens.length > 0 ? updateEffectStyles(effectTokens, settings.baseFontSize, shouldCreate) : {}),
   };
+  if (styleTokens.length < tokens.length && shouldCreate) {
+    notifyUI('Some styles were not created due to your user Settings. Make sure Settings > Ignore first part of token name doesn\'t conflict', { error: true });
```
six7, line 58:
```suggestion
    notifyUI('Some styles were not created due to your settings. Make sure Ignore first part of token name doesn\'t conflict', { error: true });
```

PR #1881, opened 2023-05-10
1610-expandtypography-in-token-transformer-returns-inconsistent-lineheight-type

swordEdge:


six7 (approved):
can you add a test ase for

PR #1888, opened 2023-05-11
fix not visible of manageThemeModal in jsonEditor

swordEdge:


six7:
I'm a bit confused, in the video it seems to jump back to the List view? We should stay on JSON view with the modal on top, can you confirm that is the case?

swordEdge:
> I'm a bit confused, in the video it seems to jump back to the List view? We should stay on JSON view with the modal on top, can you confirm that is the case?

@six7 Just the record was incorrect. Updated the video.

six7 (approved):

PR #1889, opened 2023-05-11
Upgrade storybook, cleanup core-js and update monaco

SorsOps:


.storybook/main.js
```diff
@@ -1,53 +1,70 @@
...
 module.exports = {
   stories: ['../src/**/*.stories.mdx', '../src/stories/*.stories.@(js|jsx|ts|tsx)'],
-  addons: [
-    '@storybook/addon-links',
-    '@storybook/addon-essentials',
-    '@storybook/addon-interactions',
-  ],
-  framework: '@storybook/react',
-  webpackFinal: async (config, { configType }) => {
-    config.resolve.alias['@'] = path.resolve(__dirname, '../src/');
```
six7, line 11: Does this mean our aliases still work? Wasnt able to see if you reintroduced this alias somewhere else - or is that a default now?

src/app/index.tsx
```diff
@@ -1,5 +1,3 @@
-import 'core-js/stable';
```
six7, line 1: Should we expect any side effects here?

package.json
```diff
@@ -89,7 +90,7 @@
     "postcss": "^8.2.13",
     "postcss-calc-ast-parser": "^0.1.4",
     "rc-mentions": "^2.2.0",
-    "react": "17.0.0",
+    "react": "^18.2.0",
```
six7, line 93: Should we test anything specific with the upgrade? No breaking changes?

six7 (commented):
Just some questions, given all tests are failing

src/app/index.tsx
SorsOps: We shouldn't assuming that we are using ES6 features. Most of what core-js was offering us was basic ES6 functionality, async await, rest spreads ,etc 

As long as it fits within ES6 we should be fine, otherwise we will continue to need to transpile down using babel

package.json
SorsOps: I've reverted this change, there is indeed a regression for redux that has not been addressed

PR #1949, opened 2023-06-08
1845-dont-overwrite--key

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/25951419/fc21bf51-d437-4d09-91dd-77adc20d5440


six7 (changes requested):
could you also change it so that we do not reset studio.tokens?

right now if i have something inside `studio.tokens` that is not modify, the plugin resets this. to reproduce:

have a json like this:


```
'text': {
  value: '#000000',
  type: 'color',
  '$extensions': {
    'studio.tokens': {
      'id': 'foobar'
    }
  }
}
```

and i then edit that token in the plugin ui

we remove the 'id' and just keep the 'studio.tokens'.

however, it would be good if we just respected whatever users have stored in there.

six7 (approved):

PR #1963, opened 2023-06-20
Plugin forgets credentials

roppazvan:
* When the token was refreshed, the new auth information was not saved in the Figma client storage. Updated the code to save it now. (hopefully, it should solve the plugin forgetting credentials issue). As long as the refresh token is valid it should be ok.
* As an improvement., if the user needs to re-authenticate i have saved the user email in client storage and its automatically populated in the login form

https://github.com/tokens-studio/figma-plugin/assets/21245398/c677e223-62c1-41fd-b7ca-01a66a90daf1


SorsOps (commented):
If we can just add a test to be sure that this is working with the errror handling

SorsOps (approved):

PR #1976, opened 2023-06-22
feat: connect created styles to local variables

mihkeleidast:
In relation to this question: https://tokens-studio.slack.com/archives/C03C60J1XUP/p1687428596648179?thread_ts=1687367476.523889&cid=C03C60J1XUP

This is probably a very simplified implementation, as it should probably consider some other settings as well, like whether the first part of token path is removed from style names or not. But seemed to work okay for my test, I now have styles connected to vars! 

Feel free to iterate on, orpoint out what I should do additionally to get this merged, or close at some point :)


six7:
I'll be closing this, we're working on something similar for our current epic: https://github.com/tokens-studio/figma-plugin/issues/2250

PR #1980, opened 2023-06-22
Fixes the 1mb limit on a file by swapping to raw values

SorsOps:
Fixes the 1mb limit on a single file https://octokit.github.io/rest.js/v18#repos-get-content by swapping to getting a raw value.

This should work for smaller files as well

### DIFF
--- a/src/storage/GithubTokenStorage.ts
+++ b/src/storage/GithubTokenStorage.ts
@@ -206,7 +206,7 @@ export class GithubTokenStorage extends GitTokenStorage {
               const filePath = path.startsWith(normalizedPath) ? path : `${normalizedPath}/${path}`;
               let name = filePath.substring(this.path.length).replace(/^\/+/, '');
               name = name.replace('.json', '');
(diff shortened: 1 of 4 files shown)


src/storage/GithubTokenStorage.ts
```diff
@@ -192,12 +202,11 @@ export class GithubTokenStorage extends GitTokenStorage {
               path
               && fileContent?.data
               && !Array.isArray(fileContent?.data)
-              && 'content' in fileContent.data
             ) {
               const filePath = path.startsWith(normalizedPath) ? path : `${normalizedPath}/${path}`;
               let name = filePath.substring(this.path.length).replace(/^\/+/, '');
               name = name.replace('.json', '');
-              const parsed = JSON.parse(decodeBase64(fileContent.data.content)) as GitMultiFileObject;
+              const parsed = JSON.parse(fileContent.data as unknown as string) as GitMultiFileObject;
```
six7, line 209: no chance we can properly type this?

six7 (approved):

Issue #1981, opened 2023-06-22
All themes loaded ungrouped from storage, despite having groups defined


beauroberts:
**Describe the bug**

I created a couple of themes in a theme group called "Color". When I load the JSON content attached below, the theme selector does not show the group name "Color" like I expected - instead the themes appear ungrouped.

I was able to track this down to [`src/storage/schemas/themeObjectSchema.ts`](https://github.com/tokens-studio/figma-plugin/blob/main/src/storage/schemas/themeObjectSchema.ts#L4), which does not declare `group` as a valid key. When the JSON file is validated against this schema, `group` gets filtered out.

**To Reproduce**
1. Load the JSON content attached below
2. In the theme selector, look at the hierarchy. Both the "Light" and "Dark" themes are in the ungrouped, under the heading "No group".

**Expected behavior**
The "Light" and "Dark" themes should be in a theme group called "Color".

**Screenshots or Screencasts**
<img width="583" alt="Screenshot 2023-06-22 at 17 24 43" src="https://github.com/tokens-studio/figma-plugin/assets/337712/e1fbf20b-9422-4729-b0c1-96ad11cfd77d">

**Figma file (optional)**
Send us the link to a Figma file containing a reproducible example

**JSON (optional)**

```json
{
  "global": {
    "primaryBackgroundColor": {
      "value": "#fff",
      "type": "color"
    }
  },
  "dark": {
    "primaryBackgroundColor": {
      "value": "#000",
      "type": "color"
    }
  },
  "$themes": [
    {
      "id": "light",
      "name": "Light",
      "group": "Color",
      "selectedTokenSets": {
        "global": "enabled",
        "dark": "disabled"
      }
    },
    {
      "id": "dark",
      "name": "Dark",
      "group": "Color",
      "selectedTokenSets": {
        "global": "enabled",
        "dark": "enabled"
      }
    }
  ],
  "$metadata": {
    "tokenSetOrder": [
      "global",
      "dark"
    ]
  }
}
```

thaoula:
Hi Team,

I also commented on the pull request #1981. 

We are loading from Azure Devops and the themes are ungrouped.

Regards,
Tarek

beauroberts:
[...]

PR #1982, opened 2023-06-22
Fix theme groups not being loaded (#1981)

beauroberts:
This should fix the issue identified in #1981 where theme group metadata is ignored when loading tokens from storage.

### DIFF
--- a/src/storage/__tests__/FileTokenStorage.test.ts
+++ b/src/storage/__tests__/FileTokenStorage.test.ts
@@ -15,6 +15,7 @@ const json = JSON.stringify({
     {
       id: '8722635276827d42671ab23df835867c9e0024dd',
       name: 'Light',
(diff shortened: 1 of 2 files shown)


six7 (approved):
Thank you!

thaoula:
Hi @beauroberts, @six7 

Do you know if this fix also applies to loading from Git specifically Azure DevOps. Our designers are using 1.37.1

We seem to be loosing Theme groups on each load so syncing / creating Figma variables break. 

Example of theme -
```
  {
    "id": "1196a8c17e93c98178f2e8ec40dc1cd03940cd79",
    "name": "light",
    "selectedTokenSets": {
      "light": "enabled",
      "core": "source"
    },
    "$figmaStyleReferences": {},
    "group": "Color",
}
```

Regards,
Tarek

PR #1985, opened 2023-06-23
Feat: Add source map support for Sentry tracking 

SorsOps:
This results in proper source maps when uploaded to Sentry instead of the reconstructed stack trace


manifest.json
```diff
@@ -6,5 +6,6 @@
   "ui": "dist/index.html",
   "editorType": ["figma"],
   "permissions": ["currentuser"],
-  "relaunchButtons": [{ "command": "edit", "name": "Tokens" }]
+  "relaunchButtons": [{ "command": "edit", "name": "Tokens" }],
+  "enableProposedApi": true
```
six7, line 10:
```suggestion
  "relaunchButtons": [{ "command": "edit", "name": "Tokens" }]
```

six7 (changes requested):
minor

PR #1989, opened 2023-06-26
1975-color-modifiers-do-not-carry-across-to-figma-variables

swordEdge:
image

### DIFF
--- a/src/plugin/setValuesOnVariable.ts
+++ b/src/plugin/setValuesOnVariable.ts
@@ -28,19 +28,29 @@ export default function setValuesOnVariable(
       if (!variable && shouldCreate) {
         variable = figma.variables.createVariable(t.path, collection.id, variableType);
       }
-      if (variable && variableType === 'BOOLEAN' && typeof t.value === 'string') {
-        setBooleanValuesOnVariable(variable, mode, t.value);
-      }
-      if (variable && variableType === 'COLOR' && typeof t.value === 'string') {
-        setColorValuesOnVariable(variable, mode, t.value);
-      }
-      if (variable && variableType === 'FLOAT') {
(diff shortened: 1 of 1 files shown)


src/plugin/setValuesOnVariable.ts
```diff
@@ -21,39 +21,42 @@ export default function setValuesOnVariable(
...
-      variableObj[t.name] = {
-        variable,
-        modeId: mode,
-        referenceVariable: referenceTokenName.split('.').join('/'),
-        shouldReferenceToVariable: checkIfContainsAlias(t.rawValue),
-      };
-    }
-  });
-
+      if (variable && variableType === 'BOOLEAN' && typeof t.value === 'string') {
```
SorsOps, line 31: We should probably just have a single `if (variable)` statement with a switch inside of it

src/plugin/setValuesOnVariable.ts
```diff
@@ -21,39 +21,42 @@ export default function setValuesOnVariable(
...
+        if (t.rawValue && t.rawValue?.toString().startsWith('{')) {
+          referenceTokenName = t.rawValue?.toString().slice(1, t.rawValue.toString().length - 1);
+        } else {
+          referenceTokenName = t.rawValue!.toString().substring(1);
+        }
+        variableObj[t.name] = {
+          variable,
+          modeId: mode,
+          referenceVariable: referenceTokenName.split('.').join('/'),
+          shouldReferenceToVariable: checkIfContainsAlias(t.rawValue) && !t.$extensions?.['studio.tokens']?.modify,
```
SorsOps, line 54: Is it only modify that would result in a reference?

SorsOps:
Also need to just fix the tests

SorsOps (approved):

PR #1991, opened 2023-06-26
23-variables-in-manage-themes-when-trying-to-detach-variables-we-do-not-seem-to-update-the-ui

swordEdge:
This one is for [this ticket](https://github.com/tokens-studio/figma-plugin-internals/issues/23)

https://github.com/tokens-studio/figma-plugin/assets/25951419/ad41dc6a-54db-42b7-87c0-56250aaa768d


SorsOps (approved):

src/plugin/asyncMessageHandlers/resolveVariableInfo.ts
```diff
@@ -12,8 +12,8 @@ export const resolveVariableInfo: AsyncMessageChannelHandlers[AsyncMessageTypes.
   await Promise.all(msg.variableIds.map(async (variableId) => {
```
SorsOps, line 12: We should probably add a small test around this for the future, so we don't miss this again in the future

PR #1992, opened 2023-06-26
feat: ignore first part of token name for variables

mihkeleidast:
Applies the same behavior to variables as could be previously done to styles.

Let me know if this should be done differently in some way (e.g. a separate settings value?).


six7:
Sorry for the delay on this, we're currently focused on performance. I'll make sure to focus on this for a release after the performance focus is over 🙏

railstride:
Hey @six7, were you able to make any progress on this? Very keen to "just" check the option rather than rewriting the core of a style dictionary :D

six7:
> Hey @six7, were you able to make any progress on this? Very keen to "just" check the option rather than rewriting the core of a style dictionary :D

hey, no, we're still doing the perf release

mihkeleidast:
@six7 Any updates, when should we expect progress on this? It's getting a bit annoying to have to rely on a feature branch to sync variables properly...

src/plugin/setValuesOnVariable.test.ts
```diff
@@ -48,7 +49,53 @@ describe('SetValuesOnVariable', () => {
...
+        setValueForMode: mockSetValueForMode,
+      } as unknown as Variable,
+      {
+        id: 'VariableID:309:16432',
+        key: '124',
+        name: 'primary/height',
+        setValueForMode: mockSetValueForMode,
+      } as unknown as Variable,
+    ] as Variable[];
+    const settings = { ignoreFirstPartForStyles: true } as SettingsState;
```
mihkeleidast, line 71: Variables have been in use for quite a while, so making the same setting change behavior may be unexpected. Would it be better to create a new setting for this?

mihkeleidast:
@six7 Cheers, done! I have one open question regarding whether this feature should use the existing setting, or a new setting?

Also, not sure if I have to do anything to get the failing checks to succeed, does not seem like an issue I have created?

LukeFinch:
> @six7 Cheers, done! I have one open question regarding whether this feature should use the existing setting, or a new setting?
> 
> Also, not sure if I have to do anything to get the failing checks to succeed, does not seem like an issue I have created?

HI @mihkeleidast I'll check the tests for you. As for whether it should be a new setting. I think for now - let's keep it as a separate setting. It's easier to merge them in the future if we want to. There may also be a discovery benefit from us having it called out individually.

mihkeleidast:
Refactored it to a separate settings value.

six7 (changes requested):
Thanks for all this! Code looks fine, but: As you have it right now, this setting will not be remembered each time the user restarts the plugin.

Look into `updateUISettings` to see how we do this, there you'll notice we pass through the `ignoreFirstPartForStyles`. As a user I'd expect to set this once and forget about it, not have to recheck that everytime I need it

six7 (approved):
Nice, thank you! I'll merge this in - we'll combine that with another release this week

railstride:
been following this issue and am a bit confused about the current state 😄 
so the branch got merged, then reverted due to a bug of this branch and again merged with the bug fix (1.38.6)
… so the functionality is there, right? Now the UI is still missing the option to actually set this feature, or am I missing something?

esthercheran:
Hey @railstride we reverted this as it wasn't working as expected some additional work is needed. Will keep you updated.

PR #2010, opened 2023-06-29
Feat/automated releases

SorsOps:
- Adds support for improving the release flow through github workflows
- Adds additional sentry tracking 
- Setups up changesets in anticipation of monorepo support for the tokens-transformer
- Removes old unnecessary shell scripts for Sentry and includes them into the webpack build process


src/app/sentry.ts
```diff
@@ -0,0 +1,51 @@
...
+  switch (process.env.ENVIRONMENT) {
+    case 'production':
+    case 'beta':
+      Sentry.addTracingExtensions();
+      Sentry.init({
+        dsn: DSN,
+        release: `figma-tokens@${pjs.version}`,
+        environment: process.env.ENVIRONMENT,
+        tracesSampleRate: SAMPLING,
+        // @ts-ignore This is correct, but the types are wrong.
```
roppazvan, line 39: I've recently read about ts-expect-error. Wouldn't this be more suitable for these cases?

src/app/store/models/settings.tsx
```diff
@@ -30,6 +31,10 @@ export interface SettingsState {
   shouldSwapStyles: boolean;
   baseFontSize: string;
   aliasBaseFontSize: string;
+  /**
+   * Whether the user has opted in for session recording in Sentry
+   */
+  sessionRecording:boolean;
```
roppazvan, line 37: do you have an issue with prettier ? it's weird that theres no space here

roppazvan (approved):
LGTM

src/app/store/models/settings.tsx
SorsOps: Yeah that is strange, I don't get any indication about it in any of my prettier linting either. Regardless I will amend

src/app/sentry.ts
roppazvan: yes, and when using ts-expect-error ts will let us know when the types are fixed, and we will have to remove the ts-expect-error

PR #2019, opened 2023-07-03
1996 when working with variables bordercolor is applied as style instead of variable

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/25951419/6a001094-aab0-4d96-a5f1-91827efa79d3

### DIFF
--- a/src/plugin/setBorderColorValuesOnTarget.ts
+++ b/src/plugin/setBorderColorValuesOnTarget.ts
@@ -4,7 +4,6 @@ import { trySetStyleId } from '@/utils/trySetStyleId';
 import { paintStyleMatchesColorToken } from './figmaUtils/styleMatchers';
 import { clearStyleIdBackup, getNonLocalStyle, setStyleIdBackup } from './figmaUtils/styleUtils';
 import setColorValuesOnTarget from './setColorValuesOnTarget';
-import { tryApplyColorVariableId } from '@/utils/tryApplyColorVariableId';
 
 type Props = {
   data: string;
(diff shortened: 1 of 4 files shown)


src/plugin/setBorderColorValuesOnTarget.ts
```diff
@@ -13,39 +14,45 @@ type Props = {
...
+  data, value, node, stylePathPrefix = null, stylePathSlice, styleReferences, paintStyles, figmaVariableReferences, figmaVariableMaps,
 }: Props) {
-  const pathname = convertTokenNameToPath(data, stylePathPrefix, stylePathSlice);
-  let matchingStyleId = matchStyleName(
-    data,
-    pathname,
-    styleReferences,
-    paintStyles,
-  );
+  if ('strokes' in node) {
```
SorsOps, line 24: So if I'm reading this correctly the initial bug was occuring because of  it not attempting to apply the color variable and defaulting to style application first? If so we should probably add a test case for this.

src/plugin/setValuesOnNode.ts
```diff
@@ -280,7 +280,7 @@ export default async function setValuesOnNode(
       // FILL
       if (values.fill && typeof values.fill === 'string') {
         if ('fills' in node && data.fill) {
-          if (!(await tryApplyColorVariableId(node, data.fill, figmaVariableReferences, figmaVariableMaps))) {
+          if (!(await tryApplyColorVariableId(node, data.fill, figmaVariableReferences, figmaVariableMaps, 'fills'))) {
```
SorsOps, line 283: Can we rather replace these with enums instead of strings?

SorsOps (approved):

PR #2022, opened 2023-07-04
Public Beta for Second Screen & Flow

LukeFinch:
Moves the controls for Second Screen into the Nav Bar, and creates a tab with the Second Screen controls.
Removes flag checks for launchdarkly

image
image


SorsOps:
I'm seeing an issue with the layout when running locally 
image

SorsOps:
Forgot password does not appear to be working 

The Open second screen button should likely be disabled if a user hasn't logged in already

SorsOps (approved):

PR #2036, opened 2023-07-07
33-when-users-rename-a-token-that-exists-with-the-same-old-name-across-multiple-sets-offer-rename-in-other-sets-in-the-rename-dialog

swordEdge:
[This issue](https://github.com/tokens-studio/figma-plugin-internals/issues/33)

https://github.com/tokens-studio/figma-plugin/assets/25951419/be00ff29-5c68-4f15-8b34-36c68cc9f377


src/app/components/EditTokenForm.tsx
```diff
@@ -360,6 +362,17 @@ function EditTokenForm({ resolvedTokens }: Props) {
               key: 'rename-variable', label: 'Rename variable',
             });
           }
+          const tokenSetsContainsSameToken: string[] = [];
```
SorsOps, line 365: 2 Questions. 

1. This is a slow lookup that has to iterate across all tokens to find a token. Can we not do a direct lookup on the token object ?

2. The next step seems like we might need to rename aliases as well as part of this editing. I don't think you need to address this in this PR , but we should note it as a future addition

SorsOps (approved):

PR #2043, opened 2023-07-07
1932-using-border-token-in-composition-is-not-working-properly

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/25951419/4b22b955-05ef-43c4-aab3-e87c2d5996a7

### DIFF
--- a/src/plugin/node.ts
+++ b/src/plugin/node.ts
@@ -161,6 +161,13 @@ export function destructureToken(values: MapValuesToTokensResult): MapValuesToTo
   Record<TokenTypes, SingleToken['value']>
   & Record<Properties, SingleToken['value']>
   > = {};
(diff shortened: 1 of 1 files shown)


src/plugin/node.ts
```diff
@@ -161,6 +161,13 @@ export function destructureToken(values: MapValuesToTokensResult): MapValuesToTo
   Record<TokenTypes, SingleToken['value']>
   & Record<Properties, SingleToken['value']>
   > = {};
+  if (values && values.composition) {
```
SorsOps, line 164: So the only change here was the ordering ? Were there other properties here on the values that took priority over `composition`?

SorsOps (approved):

PR #2075, opened 2023-07-18
2056-add-id-to-track-token-usage

swordEdge:


src/utils/updateTokenPayloadToSingleToken.ts
```diff
@@ -3,24 +3,24 @@ import { SingleToken } from '@/types/tokens';
 
 export function updateTokenPayloadToSingleToken(
   payload: UpdateTokenPayload,
+  id?: string,
 ): SingleToken {
   return {
     name: payload.name,
     value: payload.value,
     type: payload.type,
+    $extensions: {
```
SorsOps, line 12: I see we're missing  `...(payload.$extensions ? {` does this cause an issue if we are inadvertently passed a token without an extension property?

src/utils/updateTokenPayloadToSingleToken.ts
```diff
@@ -3,24 +3,24 @@ import { SingleToken } from '@/types/tokens';
 
 export function updateTokenPayloadToSingleToken(
   payload: UpdateTokenPayload,
+  id?: string,
```
SorsOps, line 6: Probably makes more sense to have 
```suggestion
  id: string = uuidv4(),
```
then we don't have to fix the old calls to this function

src/app/store/models/tokenState.tsx
```diff
@@ -130,7 +132,7 @@ export const tokenState = createModel<RootModel>()({
       usedTokenSet: data.usedTokenSets || state.usedTokenSet,
       themes: data.themes || state.themes,
       activeTheme: data.activeTheme || state.activeTheme,
-      tokens: data.sets || state.tokens,
+      tokens: addIdPropertyToTokesn(data.sets ?? {}) || addIdPropertyToTokesn(state.tokens),
```
SorsOps: The second part will never fire. This should probably be 
```
addIdPropertyToTokesn(data.sets || state.tokens),
```

SorsOps (changes requested):

SorsOps:
Failing tests are primarily because of `storeTokenIdInJsonEditor` now being added

src/utils/updateTokenPayloadToSingleToken.ts
swordEdge: if we should change the id as default parameter, then when we edit the token, tokenId will be changed. That is not what we expect.

src/utils/updateTokenPayloadToSingleToken.ts
swordEdge: About this comment, rightnow, we added the utility to add Id property to all tokens. In the old version, if the payload doesn't have $extension property, we don't add tokenId. But we should add tokenId even though the payload doesn't have $extensions property.

roppazvan (approved):
LGTM

PR #2103, opened 2023-07-26
Performance release

six7:
## 1.38 - Performance Improvements

We’ve changed the way the plugin stores, applies and resolves tokens for the fastest Tokens Studio experience to date.

We’ve been working on performance patch releases since 1.37 to improve performance, but 1.38 brings the removal of internal cache that led to a lot of crashes and instability.

This means switching themes, activating a new token set, making changes to large documents, or applying tokens in bulk happens up to 300x faster.

### :rocket: Notable changes:

- We have completely revamped internal token storage and removed the cache we were using which has improved performance.
- Figma finally released an update to their plugin API that we requested which makes finding layers containing tokens MUCH faster.
- Utilizing the new API to find layers with tokens on them, we no longer rely on an internal cache on nodes that requires the plug-in to keep and scan all Tokens Studio data for every operation you perform.

### :eyes: What you’ll see:

- A short “working” phase when applying tokens to layers
  - The plug-in is finding layers because it no longer has cached data to check
- It is still faster than the previous working phase to check the cache.
  - A node counter and estimated time to complete your action
- Size tokens remain the longest to update in Figma.
  - Also true with variables natively in Figma, so we can only assume this is a current limitation on the Figma API.
- Enhanced data has been added to DevMode.
  - The names of all tokens applied to the element are visible in the inspect panel
  - [Here’s how to set up dev mode](https://docs.tokens.studio/dev-mode)
- The plug-in data preview in the bottom right of the Figma Design panel has been removed.
  - The data stored there was time-consuming to populate, and limits on what we can show (type of token applied and not the value) made it not very helpful.
  - There is richer data in DevMode to replace this feature.
  - NOTE: We will not remove this data automatically. If you want to remove it:
    - Plugin Settings > Scroll to Debugging > Remove Deprecated Data
    - This only removes the preview in the design panel in Figma. It will NOT harm any tokens.
    - There is no harm in leaving the data there. But new tokens applied after this update will not show the preview.

### 😍  What you’ll notice

- Faster time to complete large operations
  - Not having to scan all cache data means we can pinpoint which nodes to work on with each click you make
- Less crashing on large token files
  - Storing cache data was eating away at limited memory which was causing the plug-in to crash
- Improved responsiveness of the plug-in
  - Using Figma’s API to look for layers containing Tokens Studio plug-in data instead of scanning the entire cache data is way less intensive on the plug-in.
- Making token changes on COMPONENTS will trickle down to INSTANCES
  - Previously we had to “clear cache” on components and that action didn’t always flow through to the instances. Now that we don’t use cache, this shouldn’t be an issue.
- Working with ALIAS TOKENS that have many levels of REFERENCES resolves much faster
  - A backend logic change simplifies resolving references. One test case went from 3 seconds to 10 milliseconds, which is 300x faster.
- EX: A button-color token which references a
  - theme-color token which references a
  - brand-color token which references a
  - core-color token which references a
  - global-color token.

### :bulb: New functionality:

- You can now clear the filter input by pressing the x button next to it when you have added a filter.

### :bug: Other smaller bugfixes:

- Changed default Apply to setting to Selection, this should not change any already stored Apply to setting
- You can now "unset" dimension tokens by either removing their token on Inspect or on Tokens view.
- We now no longer ignore invisible instance children from updating, as we're no longer relying on a Cache. This means you can start using boolean tokens in instances.
- Fixes a bug that caused nodes to not get updates when min/max width tokens were applied on instances, which caused any subsequent updates to this node to fail. We now ignore instances for min/max width tokens, as they are unsupported by Figma.
- Fixed a bug that caused Inspect to not show any results when you've used min/max width tokens on instances.
- Fixed a bug that caused Composition token's scroll menu to behave weird when scrolling down
- Fixes a bug that caused shadow tokens that were using array values to not get a value if they were redefined in another token set
- Fixed issue that caused theme groups to not play nice with branches, where we'd previously disable the theme groups after branch switching
- Fixed an issue that caused the plugin to crash when typing [ in the search input.
- Fixed performance of dragging token sets around on the left token set navigation.
- Fixed an issue with the mentions input not showing active colors introduced in the last release.
- When removing a sync provider that was the active one and then closing the plugin, we change token storage to Local. Previously, this led to Invalid credentials.


samiamdesigns:
1.38 - Performance Improvements

We’ve changed the way the plugin stores, applies and resolves tokens for the fastest Tokens Studio experience to date. 

We’ve been working on performance patch releases since 1.37 to improve performance, but 1.38 brings the ​​removal of internal cache that led to a lot of crashes and instability.

This means switching themes, activating a new token set, making changes to large documents, or applying tokens in bulk happens up to 300x faster. 

:rocket: Notable changes:

- We have completely revamped internal token storage and removed the cache we were using which has improved performance.
- Figma finally released an update to their plugin API that we requested which makes finding layers containing tokens MUCH faster. 
- Utilizing the new API to find layers with tokens on them, we no longer rely on an internal cache on nodes that requires the plug-in to keep and scan all Tokens Studio data for every operation you perform.

:eyes: What you’ll see: 

- A short “working”phase when applying tokens to layers
-- The plug-in is finding layers because it no longer has cached data to check
- It is still faster than the previous working phase to check the cache. 
-- A node counter and estimated time to complete your action
- Size tokens remain the longest to update in Figma.
-- Also true with variables natively in Figma, so we can only assume this is a current limitation on the Figma API. 
- Enhanced data has been added to DevMode. 
--The names of all tokens applied to the element are visible in the inspect panel 
-- Here’s how to set up dev mode
-- https://docs.tokens.studio/dev-mode
- The plug-in data preview in the bottom right of the Figma Design panel has been removed.
-- The data stored there was time consuming to populated, and limits on what we can show (type of token applied and not the value) made it not very helpful. 
-- There is richer data in DevMode to replace this feature. 
-- NOTE: We will not remove this data automatically. If you want to remove it:
-- Plugin Settings > Scroll to Debugging> Remove Deprecated Data
-- This only removes the preview in the design panel in Figma. It will NOT harm any tokens. 
-- There is no harm in leaving the data there. But new tokens applied after this update will not show the preview. 

:face_with_monocle: What you’ll notice 

- Faster time to complete large operations 
-- Not having to scan all cache data means we can pin-point which nodes to work on with each click you make
- Less crashing on large token files
-- Storing cache data was eating away at limited memory which was causing the plug-in to crash
- Improved responsiveness of the plug-in 
-- Using Figma’s API to look for layers containing Tokens Studio plug-in data instead of scanning the entire cache data is way less intensive on the plug-in. 
- Making token changes on COMPONENTS will trickle down to INSTANCES
-- Previously we had to “clear cache” on components and that action didn’t always flow through to the instances. Now that we don’t use cache, this shouldn’t be an issue. 
- Working with ALIAS TOKENS that have many levels of REFERENCES resolves  much faster
-- A backend logic change simplifies resolving references. One test case went from 3 seconds to 10 milliseconds, which is 300x faster.
- EX: A button-color token which references a 
-- theme-color token which references a
-- brand-color token which references a 
-- core-color token which references a 
-- global-color token. 

:bulb:New functionality:

- You can now clear the filter input by pressing the x button next to it when you have added a filter.

:bug: Other smaller bugfixes:

- Changed default Apply to setting to Selection, this should not change any already stored Apply to setting
- You can now "unset" dimension tokens by either removing their token on Inspect or on Tokens view.
- We now no longer ignore invisible instance children from updating, as we're no longer relying on a Cache. This means you can start using boolean tokens in instances.
- Fixes a bug that caused nodes to not get updates when min/max width tokens were applied on instances, which caused any subsequent updates to this node to fail. We now ignore instances for min/max width tokens, as they are unsupported by Figma.
- Fixed a bug that caused Inspect to not show any results when you've used min/max width tokens on instances.
- Fixed a bug that caused Composition token's scroll menu to behave weird when scrolling down
- Fixes a bug that caused shadow tokens that were using array values to not get a value if they were redefined in another token set
- Fixed issue that caused theme groups to not play nice with branches, where we'd previously disable the theme groups after branch switching
- Fixed an issue that caused the plugin to crash when typing [ in the search input.
- Fixed performance of dragging token sets around on the left token set navigation.
- Fixed an issue with the mentions input not showing active colors introduced in the last release.
- When removing a sync provider that was the active one and then closing the plugin, we change token storage to Local. Previously, this led to Invalid credentials.

PR #2110, opened 2023-07-27
improve create variable function

swordEdge:


six7 (changes requested):
Before we merge this in, can we add tests for the functions here? It feels risky to have all these things change without any changed test. We need to avoid having to test manually.

Let's write tests for these functions that are not yet tested related to creating variables, as this is an important piece.

six7 (approved):
🎉

PR #2135, opened 2023-08-03
analysis of the main functions

swordEdge:
Investigation result:
Plugin performance depends on some main functions. Most of features like updateDocuments and switchThemes call those heavy functions together.

resolveTokenValues, updateNodes, updatePluginData, sendSelectionChange
We should optimize those functions.
There is no big O (n^3), but some functions call above big functions together.


src/plugin/generateTokensToCreate.ts
```diff
@@ -5,7 +5,7 @@ import { AnyTokenList } from '@/types/tokens';
 import { mergeTokenGroups, resolveTokenValues } from './tokenHelpers';
 
 export function generateTokensToCreate(theme: ThemeObject, tokens: Record<string, AnyTokenList>, availableTokenTypes: TokenTypes[]) {
-  // Big O(n * m)
+  // Big O(resolveTokenValues + mergeTokenGroups)
```
SorsOps, line 8:
```suggestion
  // Big O(resolveTokenValues * mergeTokenGroups)
```

SorsOps (approved):

PR #2138, opened 2023-08-04
2058-add-sentry-metrics

swordEdge:


src/app/components/Tokens.tsx
```diff
@@ -99,6 +100,9 @@ function Tokens({ isActive }: { isActive: boolean }) {
   const tokenDiv = React.useRef<HTMLDivElement>(null);
   const { t } = useTranslation(['tokens']);
 
+  const transaction = getCurrentHub().getScope().getTransaction();
```
SorsOps, line 103: We shouldn't add these as part of a component. It's render cycle might either not get called in a real transaction or get called multiple times. We should rather focus this on the actual transactions on the redux models or on the actions of say a bulk remap

src/plugin/NodeManager.ts
```diff
@@ -273,7 +274,12 @@ export class NodeManager {
     } else if (updateMode === UpdateMode.SELECTION) {
       relevantNodes = findAll(figma.currentPage.selection, true);
     } else {
-      relevantNodes = findAll([figma.root], false);
+      relevantNodes = await wrapTransaction({
```
SorsOps, line 277: The backend of the plugin doesn't have sentry added because it can't emit the messages because fetch,xhr request etc aren't present in the backend sandbox. None of these would be called or tracked

SorsOps:
We should focus on annotating the existing transactions , so these 
image

with additional metrics data

src/app/store/models/tokenState.tsx
```diff
@@ -589,6 +589,10 @@ export const tokenState = createModel<RootModel>()({
       try {
         wrapTransaction({
           name: 'updateDocument',
+          statExtractor: (result, transaction) => {
+            transaction.setMeasurement('tokenSets', Object.keys(rootState.tokenState.tokens).length, '');
```
SorsOps, line 593: We could probably add tokens here as well

src/plugin/tokenHelpers.ts
```diff
@@ -1,4 +1,5 @@
 import omit from 'just-omit';
+import { startTransaction } from '@sentry/react';
```
SorsOps, line 2: Isn't this still being imported on the backend and not just the frontend?

src/app/store/models/tokenState.tsx
```diff
@@ -590,6 +589,10 @@ export const tokenState = createModel<RootModel>()({
         wrapTransaction({
           name: 'updateDocument',
           statExtractor: (result, transaction) => {
+            transaction.setMeasurement('tokens', Object.entries(rootState.tokenState.tokens).reduce((acc, curr) => {
```
SorsOps, line 592:
```suggestion
             transaction.setMeasurement('tokens', Object.entries(rootState.tokenState.tokens).reduce((acc, [key,tokens]) => {
              acc += tokens.length;
              return acc;
            }, 0), '');
```

src/plugin/tokenHelpers.ts
```diff
@@ -1,5 +1,5 @@
 import omit from 'just-omit';
-import { startTransaction } from '@sentry/react';
+import { startTransaction } from '@sentry/browser';
```
SorsOps, line 2: We will need to omit this completely, transactions cant be profiled on the backend without the bridge

SorsOps:
Just the fix to the tests remaining

SorsOps (approved):

PR #2150, opened 2023-08-08
2050-backend-bridge-to-sentry

swordEdge:


SorsOps (approved):

SorsOps:
I would not count this as full completion of the Sentry bridge however, that would require us running Sentry in the backend but proxying the the network requests using a custom transport

SorsOps:
As Jan mentioned, this will likely result in a flood of unnecessary errors which will affect tracking . I think we should recreate a new PR with the actual bridge code

PR #2159, opened 2023-08-11
finished all Chinese translation

MrBiscuit:


src/i18n/lang/zh/tokens.json
```diff
@@ -70,9 +70,9 @@
   "createStyles": "创建样式",
   "createVariables": "创建变量",
   "sets": {
-    "title": "套",
+    "title": "集",
     "description": "集允许您将令牌拆分为多个文件。您可以激活不同的集来控制主题。",
```
LukeFinch, line 74: Should this be

集允许您将 Token 拆分为多个文件。您可以激活不同的集来控制主题。

To be in line with the rest of the styling of 'Token' being in english?

PR #2162, opened 2023-08-14
backend bridge

six7:
initial PR to introduce a way for us to measure perf of functions happening in the sandbox


src/profiling/backendTransaction.ts
```diff
@@ -0,0 +1,24 @@
...
+  switch (process.env.ENVIRONMENT) {
+    case 'production':
+    case 'beta':
+      return !!process.env.SENTRY_DSN;
+    default:
+      return false;
+  }
+};
+
+export const wrapBackendTransaction = async <U>(name: string, fn: () => U) => {
```
SorsOps, line 13: So this is what I thought of initially , but I think when we start talking about how to handle spans ,etc that this implementation ends up being too crude.

I think best case would be we create a custom transport for the server Sentry, that will proxy the values to this reportFromPlugin function  and it _should_ fix the issue with fetch not being available. 

We might still have to do some basic mocking/polyfilling on other functions that sentry expects.

If we don't do this we won't be able to connect existing transactions on the front end with the backend as we'll want to track server interactions as if they were spans as part of a transaction started on the front end.

To also better support the manual transactions we're currently doing, we should look at using the decorator pattern to identify which functions should be profiled and then use decorators inside to handle the stats extraction we are currently doing

PR #2172, opened 2023-08-18
fix apply performance

six7:
Fixes https://github.com/tokens-studio/figma-plugin/issues/2129

Removes `updatePluginData` calls from the `update` call, as we only used that for migration purposes. We would still do that when users apply tokens or remap tokens, but we won't do anymore on every Apply to.. call

* `src/plugin/asyncMessageHandlers/update.ts`: The `updatePluginData` function is no longer used in the `update.ts` file, so the import statement and call to the function have been removed. (remove `updatePluginData` references in `update.ts`)

https://github.com/tokens-studio/figma-plugin/assets/4548309/57759166-42be-4d1e-878a-89ecc63301f7

### DIFF
--- a/src/plugin/node.ts
+++ b/src/plugin/node.ts
@@ -31,6 +31,14 @@ export function returnValueToLookFor(key: string) {
   }
 }
 
+const borderPropertyMap = new Map<Properties, string>([
+  [Properties.border, 'border'],
+  [Properties.borderTop, 'borderTop'],
+  [Properties.borderRight, 'borderRight'],
+  [Properties.borderBottom, 'borderBottom'],
+  [Properties.borderLeft, 'borderLeft'],
+]);
+
 type MapValuesToTokensResult = Record<string, string | number | SingleToken['value'] | {
   property: string
   value?: SingleToken['value'];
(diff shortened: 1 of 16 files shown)


src/plugin/node.ts
```diff
@@ -49,22 +36,53 @@ type MapValuesToTokensResult = Record<string, string | number | SingleToken['val
...
         const modifier = resolvedToken.$extensions?.['studio.tokens']?.modify;
         if (modifier) {
           acc[key] = modifier.type === ColorModifierTypes.MIX ? `${resolvedToken.rawValue} / mix(${modifier.color}, ${modifier.value}) / ${modifier.space}` : `${resolvedToken.rawValue} / ${modifier.type}(${modifier.value}) / ${modifier.space}`;
         }
+      } else if (key === TokenTypes.COMPOSITION) {
+        Object.entries(resolvedToken.value).forEach(([property, value]) => {
+          // Assign the actual value of a composition token property to the applied values
+          acc[property as CompositionTokenProperty] = value;
+          // If we're dealing with border tokens we want to extract the color part to be applied (we can only apply color on the whole border, not individual sides)
+          if (typeof value === 'object' && ['border', 'borderTop', 'borderRight', 'borderBottom', 'borderLeft'].includes(property) && 'color' in value && typeof value.color === 'string') {
```
SorsOps, line 58: Microtweak, but we don't want to iterate through an array, just use a map for this lookup. It should probably also use the `Properties` values as well instead of raw strings

src/plugin/node.ts
```diff
@@ -49,22 +36,53 @@ type MapValuesToTokensResult = Record<string, string | number | SingleToken['val
...
+            acc.borderColor = value.color;
+          }
+        });
+        // Same as above, if we're dealing with border tokens we want to extract the color part to be applied (we can only apply color on the whole border, not individual sides)
+      } else if (key === Properties.border && resolvedToken.type === TokenTypes.BORDER && typeof resolvedToken.value === 'object' && 'color' in resolvedToken.value && resolvedToken.value.color) {
```
SorsOps, line 63: This if statement block has pretty much the exact same predicates.  lets rather compress this into a single if statement

SorsOps (approved):

Issue #2130, opened 2023-08-03
[Performance]: updatePluginData function is taking too much time


esthercheran:

six7:
@esthercheran can you flesh this out still?

Issue #2128, opened 2023-08-03
[Performance]: updateDocument functions are taking too much time


esthercheran:

six7:
@esthercheran can you flesh this out still?

Issue #1901, opened 2023-05-15
[Performance]: Improve performance for large documents


six7:
Next would be how we handle updates to large documents. There are 2 pieces to this, 1 maintaining some form of index so we don't have to walk the entire tree each time if nothing has changed. This would need to listen to events to determine if a change has occurred and then invalidate the index if need be. Using a weakmap would be the best here to make sure we don't keep references in memory. I know we currently have the NodeManager for this but we iterate across all nodes on the update. While we do have O(1) lookups, we do O(n) changes by iterating through every node when making changes.

SorsOps:
We should also be adding async messages to push metrics like the amount of nodes. time for walk etc  to the front end to report to Sentry

SorsOps:
Added the above as #2058

SorsOps:
Note for the initial task, we should consider this to be dependent on #2053 , #2058  and #2057   to give us additional insights on tasks and follow up for this issue

PR #2176, opened 2023-08-21
fix: make apply faster (the one where we remove the cache)

six7:
Fixes https://github.com/tokens-studio/figma-plugin/issues/1901
Fixes https://github.com/tokens-studio/figma-plugin/issues/2129
Fixes https://github.com/tokens-studio/figma-plugin/issues/2130
Fixes https://github.com/tokens-studio/figma-plugin/issues/2128

This PR does dramatic things:

- Removed the cache in favor of Figma's new plugin api feature to more performantly get all nodes that have plugin data
- Refactor the way we apply nodes to be faster
- Refactor a few usages of updatePluginData and updateNodes, two slow functions
- Removed relaunchButtons and any updates to that plugindata. Users have voices confusion with these, and they don't add much value, can become out of sync and cost performance.

Overall performance boost is significant:
- Old prod build took 1min 30 to apply to 24 pages with complex tokens inside, subsequent runs take around 1min
- New build takes 16-20 seconds, on every run


src/utils/formatNumber.ts
```diff
@@ -0,0 +1,8 @@
+export function formatNumber(num: number): string {
+  if (num >= 1000000) {
+    return num.toString();
+  } if (num >= 1000) {
+    return `${(num / 1000).toFixed(1)}k`;
```
SorsOps, line 5: This is very weird that we have this strange gap between 1000 to 1000000 expressed in k. Can you add a comment as to why?

six7: thanks for catching that one, removed that one for now - tbh it's even unnecessary as we likely won't run 1M nodes (plus i think i just forgot to handle that case proper)

src/plugin/asyncMessageHandlers/getThemeReferences.ts
```diff
@@ -16,13 +16,13 @@ export async function getThemeReferences(prefixStylesWithThemeName?: boolean) {
   const activeThemes = themeInfo.themes?.filter((theme) => Object.values(themeInfo.activeTheme).some((v) => v === theme.id));
   const stylePathPrefix = prefixStylesWithThemeName && activeThemes.length > 0 ? activeThemes[0].name : undefined;
 
-  activeThemes?.forEach((theme) => {
-    Object.entries(theme.$figmaVariableReferences ?? {}).forEach(async ([token, variableId]) => {
+  activeThemes?.forEach(async (theme) => {
```
SorsOps, line 19: If we don't await it at the top level it won't block the execution of this function till after all have been completed

```suggestion
  await Promise.all(activeThemes?.map(async (theme) => {
```

Issue #2099, opened 2023-07-25
Cancel button does not work when an operation has started


esthercheran:
**Describe the bug**
When a task has started in the plugin (change theme, bulk remap etc) clicking on the cancel button does not stop the task

**To Reproduce**
1. Change theme in the plugin
2. The plugin shows the ongoing task on the top bar. 
3. Click on the cancel button
4. The plugin freezes, Figma file freezes and the task doesn't go ahead anymore

**Expected behavior**
On clicking cancel the ongoing task should stop immediately and freeze the document or the plugin 

**Screenshots or Screencasts**

https://github.com/tokens-studio/figma-plugin/assets/66366384/9d9607b3-cfb5-401d-8587-533fe69d8878

six7:
I started work on this in #2177 but had to skip as I was not able to get this done in a performant way.

The TLDR is that currently Cancel waits for the remainders of the items to run through once, even though it doesn't execute on the current job. If there's a followup-job, it will still do that with the remaining items, which is an issue.

Ideally if I click on Cancel it should just flush the entire job pool and stop everything it's doing immediately.

PR #2177, opened 2023-08-21
fix cancel

six7:
fixes an issue where `Cancel` wouldn't reliably cancel the plugin, but rather freeze it :o 

Fixes https://github.com/tokens-studio/figma-plugin/issues/2099

### DIFF
--- a/src/plugin/Worker.ts
+++ b/src/plugin/Worker.ts
@@ -58,8 +58,6 @@ export class Worker {
 
   public cancel = () => {
     this.pool.clear();
-    this.emitter.emit('canceled', Date.now());
-    this.emitter.removeAllListeners();
   };
 
   public start = () => {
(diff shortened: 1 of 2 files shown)


src/plugin/Worker.ts
```diff
@@ -58,8 +58,6 @@ export class Worker {
 
   public cancel = () => {
     this.pool.clear();
-    this.emitter.emit('canceled', Date.now());
```
SorsOps, line 61: Are you sure we don't need to clean this emitter of listeners if cancelled during a partial application ? This feels like it would result in some difficult to detect bugs if we didn't.

I'm assuming this wasn't the main culprit of the freeze as these should be pretty simple operations( Assuming that a `canceled` listener doesn't do some blocking operation)

six7: Been cracking my head with this for too long now. I think you're right, but I have no better solution. If you find some time could be interesting to dig in to this, for now I move it back to ready for dev.

Issue #2097, opened 2023-07-25
Allow To Manually Select Input-field


keeganedwin:
**Is your feature request related to a problem? Please describe.**
When editing a token, the user selects the value field to edit, but the plugin automatically selects (auto-focus) the name-field for the user to edit.

**Key information:**
User only seems to be facing this issue on a Windows system. 

**Describe the solution you'd like**
Allow the user to select the field they want to edit.

**Additional context**
https://github.com/tokens-studio/figma-plugin/assets/93133012/84afa36b-de36-47dd-9b83-e340ad206ebf

six7:
Reason for that is likely the 50ms timeout we have on focusing the first input. I've never been able to work around that, but maybe someone else wants to pick that up.

https://github.com/tokens-studio/figma-plugin/blob/4a2253d1375a8d3dc2f9018321d1d5f6df437a63/src/app/components/Input.tsx#L187

six7:
Shipped in 1.37.10

Issue #2031, opened 2023-07-06
Remapping Tokens Changes the 'Apply To' Setting


keeganedwin:
**Describe the bug**
After remapping a token and selecting an option to update all token layers, the _`Apply To`_ setting is changed according to the option selected in the modal. 

**To Reproduce**
1. Select an initial 'Apply to' setting
2. Rename a token
3. Check Page or Document
4. See error - 'Apply to' changes to the option checked

**Expected behavior**
Have the plugin retain the initial selected setting, and not change based on the option selected when remapping a token. 

**Screenshots or Screencasts**
https://github.com/tokens-studio/figma-plugin/assets/93133012/d970e3c4-3775-4d47-b6b8-269522d20256

six7:
Shipped in 1.37.10

Issue #1799, opened 2023-04-11
Remember `Rename styles` choice


XavierVilars:
**Describe the bug**
In the studio, renaming a token will not necessarely rename the according style. It seems the checkbox "Rename style" is unticked by default. it should ideally remember the last used choice.

**To Reproduce**
1. Rename a token binded with a style in Figma
2. click on validate
3. check the rename style textbox
4. Validate. At this point all is ok.
5. Take another token binded with a style in Figma and rename it
6. validate
7. the tick box "rename style" is now already checked.
8. Validate
9. the style is not modify.

**Expected behavior**
It is good to keep the previous choice of the user. I'm modifying some token's name and I don't want to check and check again the same settings to validate. So, you just need to correct this bug.

**Screenshots or Screencasts**
![image](https://user-images.githubusercontent.com/104128778/231090804-e056660f-c1ee-4463-b60b-7722e67514fc.png)

six7:
Shipped in 1.37.10

PR #2199, opened 2023-08-30
fix rename/updatemode and input issues

six7:
Fixes https://github.com/tokens-studio/figma-plugin/issues/2031
Fixes https://github.com/tokens-studio/figma-plugin/issues/2097
Fixes https://github.com/tokens-studio/figma-plugin/issues/1799

This set of changes includes improvements to the user experience and state management, as well as bug fixes. The `ConfirmDialog` and `Input` components were updated to ensure immediate focus on the input element after rendering. The `useTokens` module and `EditTokenForm` component now store the last used update mode when renaming tokens, improving state management and UI. Two bug fixes were also included, related to issues when renaming tokens or token groups.

* `src/app/components/Input.tsx`: `setTimeout` function removed and `focus()` called directly on `htmlInputRef.current` to ensure immediate focus on input element after rendering. (F6c8a7b5L3)
* `src/app/components/ConfirmDialog.tsx`: Improved UX by focusing on first input field after `ConfirmDialog` is shown by removing setTimeout function and adding `firstInput` to useEffect dependencies in `ConfirmDialog.tsx`. (F7d9e2c8L1)
* `src/app/store/useTokens.tsx`: Stores last used update mode when renaming tokens and groups, and UI now defaults to last used update mode instead of current mode. (F5a6b1e4L1)
* `src/app/components/EditTokenForm.tsx`: Stores last used update mode when renaming tokens, ensuring it is selected by default instead of current mode. Improves state management and UI. (F2b5c8f1L1)
* `.changeset/curly-ravens-prove.md` and `.changeset/young-kangaroos-brake.md`: Updated `@tokens-studio/figma-plugin` to fix bug causing issues when renaming tokens or token groups. (Bug fixes)

### DIFF
--- a/src/app/components/TokenTooltip/TokenTooltip.tsx
+++ b/src/app/components/TokenTooltip/TokenTooltip.tsx
@@ -1,7 +1,9 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
 import { SingleToken } from '@/types/tokens';
 import Tooltip from '../Tooltip';
 import { TokenTooltipContent } from './TokenTooltipContent';
+import { showEditFormSelector } from '@/selectors';
 
 type Props = {
   token: SingleToken;
@@ -11,6 +13,9 @@ export const TokenTooltip: React.FC<Props> = ({
   children,
   token,
 }) => {
+  // When we open the token edit form we don't want tooltips to show through, which is happening sometimes
+  const showEditForm = useSelector(showEditFormSelector);
+
   if (!children || !React.isValidElement(children)) {
     return null;
   }
@@ -18,7 +23,7 @@ export const TokenTooltip: React.FC<Props> = ({
   return (
     <Tooltip
       side="bottom"
-      label={(
+      label={showEditForm ? '' : (
         <TokenTooltipContent
           token={token}
         />
--- a/.changeset/curly-ravens-prove.md
+++ b/.changeset/curly-ravens-prove.md
@@ -1,0 +1,5 @@
+---
+'@tokens-studio/figma-plugin': patch
+---
+
+Fixes an issue where renaming a token or token group would cause `Apply to` to be changed
--- a/.changeset/rare-cups-join.md
+++ b/.changeset/rare-cups-join.md
@@ -1,0 +1,5 @@
+---
+'@tokens-studio/figma-plugin': patch
+---
+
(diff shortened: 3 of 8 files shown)


src/app/components/TokenTooltip/TokenTooltip.tsx
```diff
@@ -11,14 +13,17 @@ export const TokenTooltip: React.FC<Props> = ({
   children,
   token,
 }) => {
+  // When we open the token edit form we don't want tooltips to show through, which is happening sometimes
+  const showEditForm = useSelector(showEditFormSelector);
+
   if (!children || !React.isValidElement(children)) {
     return null;
   }
```
roppazvan, line 21:
```suggestion
  if (!children || !React.isValidElement(children) || showEditForm) {
    return null;
  }
```

src/app/components/TokenTooltip/TokenTooltip.tsx
```diff
@@ -11,14 +13,17 @@ export const TokenTooltip: React.FC<Props> = ({
...
 
   return (
     <Tooltip
       side="bottom"
-      label={(
+      label={showEditForm ? '' : (
```
roppazvan, line 26:
```suggestion
      label={(
```

roppazvan: wouldn't it be better to add the condition above and just return null when the editForm is visible ?

roppazvan (approved):

src/app/components/TokenTooltip/TokenTooltip.tsx
six7: that would cause the children to also not be rendered though (Tooltip wraps the token button)

src/app/components/TokenTooltip/TokenTooltip.tsx
six7: same as further down, that would cause jumps in layout as the button would not be rendered anymore. we just dont want the tooltip to be renderd

src/app/components/TokenTooltip/TokenTooltip.tsx
roppazvan: ah got it.  then the best way would be something like if (showEditForm) return children

six7: I'll include that in a followup 👍

PR #2217, opened 2023-09-04
introduce remove deprecated data feature

six7:
Introduces an option `Remove deprecated data` in `Settings` near the other debug / reset features where users can remove the deprecated `relaunchData` from their nodes.


src/app/store/useDebug.tsx
```diff
@@ -0,0 +1,35 @@
...
+import useConfirm from '../hooks/useConfirm';
+import { AsyncMessageChannel } from '@/AsyncMessageChannel';
+import { wrapTransaction } from '@/profiling/transaction';
+import { UpdateMode } from '@/constants/UpdateMode';
+
+export function useDebug() {
+  const { confirm } = useConfirm();
+  async function removeRelaunchData() {
+    const userDecision = await confirm({
+      text: 'Remove relaunch data?',
```
SorsOps, line 11: Translations needed for this

src/app/store/useDebug.tsx
```diff
@@ -0,0 +1,35 @@
+import { AsyncMessageTypes } from '@/types/AsyncMessages';
+import useConfirm from '../hooks/useConfirm';
+import { AsyncMessageChannel } from '@/AsyncMessageChannel';
+import { wrapTransaction } from '@/profiling/transaction';
+import { UpdateMode } from '@/constants/UpdateMode';
+
+export function useDebug() {
+  const { confirm } = useConfirm();
+  async function removeRelaunchData() {
```
SorsOps, line 9: Lets use a callback around this

src/plugin/asyncMessageHandlers/removeRelaunchData.ts
```diff
@@ -0,0 +1,61 @@
...
+      timePerTask: 1,
+      completedTasks: 0,
+      totalTasks: relevantNodes.length,
+    },
+  });
+
+  const tracker = new ProgressTracker(BackgroundJobs.PLUGIN_REMOVE_RELAUNCH_DATA);
+  const promises: Set<Promise<void>> = new Set();
+
+  relevantNodes.forEach((node) => {
```
SorsOps, line 38: Probably cleaner to just do  this ?
```suggestion
  await Promise.all(relevantNodes.map(node)=> defaultWorker.schedule(async () => {
        try {
          node.setRelaunchData({});
        } catch (e) {
          console.log('got error', e);
        } finally {
          tracker.next();
          tracker.reportIfNecessary();
          Promise.resolve();
        }
      }))
```

src/app/store/useDebug.tsx
```diff
@@ -30,6 +34,6 @@ export function useDebug() {
         area: userDecision.data[0],
       }));
     }
-  }
+  }, [confirm]);
```
SorsOps, line 37: Need to add `t` here to as it changes when the user switches language

PR #2219, opened 2023-09-04
fix variables perf on epic branch

six7:
This takes the changes from the [hotfix](https://github.com/tokens-studio/figma-plugin/pull/2216) to `main` and bases them on the perf branch


src/plugin/asyncMessageHandlers/getThemeReferences.ts
```diff
@@ -1,30 +1,27 @@
...
     await Promise.all(Object.entries(theme.$figmaVariableReferences ?? {}).map(async ([token, variableId]) => {
-      try {
-        const foundVariableId = await figma.variables.importVariableByKeyAsync(variableId);
-        if (foundVariableId) {
-          figmaVariableReferences.set(token, foundVariableId);
-        }
-      } catch (e) {
-        console.error(e);
+      if (!figmaVariableReferences.has(token)) {
+        figmaVariableReferences.set(token, variableId);
```
SorsOps, line 24: If we do this , we don't need the overhead of the promises wrapping around this

src/utils/tryApplyColorVariableId.ts
```diff
@@ -1,32 +1,49 @@
...
+  if (!hasCachedVariable && typeof variableMapped === 'string') {
+    try {
+      const foundVariable = await figma.variables.importVariableByKeyAsync(variableMapped);
+      if (foundVariable) {
+        resolvedVariableReferences.set(token, foundVariable);
+        variable = foundVariable;
+      }
+    } catch (e) {
+      console.log('error importing variable', e);
+      Promise.reject();
```
SorsOps, line 26: We should be rejecting with an error  even if we don't throw it

src/plugin/updatePluginDataAndNodes.ts
```diff
@@ -78,6 +78,7 @@ export async function updatePluginDataAndNodes({
     }));
   });
   await Promise.all(promises);
+  resolvedVariableReferences.clear();
```
SorsOps, line 81: I think this would be easier to read if we had a function called start and stop for this to clear the values. Whilst it would do exactly the same thing as clearing the values, it would be significantly easier to know when a transaction scope begins and ends with these

PR #2220, opened 2023-09-04
Add support for sentry backend

### LINKED ISSUE #2050: Backend bridge to Sentry
We currently have an issue with tracking errors and transactions that occur in the backend controller portion of the plugin. This is because the backend is a treated as a completely separate process in the plugin.

We do however need to track how transactions are down between the UI and controller of the plugin.

Only the UI is currently capable of interacting with the sentry service as the backend cannot perform any network requests.

To circumvent this we should be able to create a transport on the Sentry instance in the backend that can be used to proxy the requests to the front end so that it can handle sending the transaction data.

 NOTE , since the backend is so stripped down in terms of what is available, we might need to shim some functionality just for basic tracing . Effort has been set to medium assuming everything goes well, but if shimming becomes an issue, it should be considered to be timeboxed for a single day

SorsOps:
Added support for a seperate instance of Sentry on the backend that can be used for transactions and profiling. It proxies all egress through the front end which could be used for enriching transactions prior to being sent or flushed

Additionally updated package manager to prefer yarn with deduping 

Closes #2050


README.md
```diff
@@ -36,7 +36,7 @@ Whenever you apply a token to a layer, the plugin will store hidden information
 Most information can be found in the [docs](https://docs.tokens.studio) or [our website](https://tokens.studio).
 
 # Contribute
-* Run `npm ci` to install dependencies.
+* Run `yarn  --frozen-lockfile --prefer-offline` to install dependencies.
```
six7, line 39: if we're using `yarn` to install deps, shouldnt we use it to run as well? feels a bit weird to have one for installing and the other for running?

six7 (approved):
Looking great 🎉  feel free to ignore my one comment and merge

six7:
Merging this in, let's address my comments in a different PR if you think that makes sense

PR #2226, opened 2023-09-06
16 variables write tests for new functions added

swordEdge:
Add new tests for variable releated functions


package-lock.json
SorsOps, line 1: Why are we getting a change in the lock file ? Are we remediating a vulnerability on anything

SorsOps:
Could you increase the benchmark `HARD_PERFORMANCE_LEEWAY` to be 15 for now. I will make a PR later to try get the speed tests to be more consistent

Issue #2224, opened 2023-09-06
Typing [ in the Search field Crashes the plugin


keeganedwin:
**Describe the bug**
Typing ` [ ` in the plugin crashes the plugin.

**Expected behavior**
Have the plugin not crash when users by mistake type in any character in the search field. After testing, the following characters don't work:

- Backslash `\`
- Brackets `( )`
- Question mark `?`
- Asterisk `*`
- Forward bracket`[`
- Add sign `+`

**Screenshots or Screencasts**
![CleanShot 2023-09-06 at 14 23 23](https://github.com/tokens-studio/figma-plugin/assets/93133012/1d79edce-7154-436e-9bae-6e99f4a500c1)

six7:
Fixed in 1.38 🚢

PR #2229, opened 2023-09-08
2224-typing-in-the-search-field-crashes-the-plugin

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/25951419/9305cfb3-7f7a-4a59-baf5-9e004c52a665

[This issue](https://github.com/tokens-studio/figma-plugin/issues/2224)


SorsOps:
@six7 This feels like this should be merged into main not the performance branch correct?

six7 (approved):

PR #2234, opened 2023-09-11
2059-calculated-values-are-not-respected-in-variables-1

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/25951419/ffad2024-4e44-4f31-beef-54a0ec5ccae5


src/plugin/updateVariablesFromPlugin.ts
```diff
@@ -61,6 +61,7 @@ export default async function updateVariablesFromPlugin(payload: UpdateTokenVari
             case TokenTypes.BORDER_WIDTH:
             case TokenTypes.SPACING:
             case TokenTypes.NUMBER:
+              console.log('spacing');
```
SorsOps, line 64: Remove

SorsOps (changes requested):

PR #2235, opened 2023-09-12
Fix/When we create the variable, collections are created 2 times

swordEdge:
Fixed the bug where the variable collections are created 2 times in the epic branch.
In the epic branch, 2 collections are created.

image
image

Fixed so that there will be only one collection.
image

### DIFF
--- a/src/plugin/createLocalVariablesInPlugin.ts
+++ b/src/plugin/createLocalVariablesInPlugin.ts
@@ -20,9 +20,8 @@ export default async function createLocalVariablesInPlugin(tokens: Record<string
   });
   const allVariableCollectionIds: Record<string, LocalVariableInfo> = {};
   let referenceVariableCandidates: ReferenceVariableType[] = [];
-  const existingVariableCollections = figma.variables.getLocalVariableCollections();
   themeInfo.themes.forEach((theme) => {
-    const collection = existingVariableCollections.find((vr) => vr.name === (theme.group ?? theme.name));
+    const collection = figma.variables.getLocalVariableCollections().find((vr) => vr.name === (theme.group ?? theme.name));
     if (collection) {
       const mode = collection.modes.find((m) => m.name === theme.name);
       const modeId: string = mode?.modeId ?? createVariableMode(collection, theme.name);


src/plugin/createLocalVariablesInPlugin.ts
```diff
@@ -20,9 +20,8 @@ export default async function createLocalVariablesInPlugin(tokens: Record<string
   });
   const allVariableCollectionIds: Record<string, LocalVariableInfo> = {};
   let referenceVariableCandidates: ReferenceVariableType[] = [];
-  const existingVariableCollections = figma.variables.getLocalVariableCollections();
   themeInfo.themes.forEach((theme) => {
-    const collection = existingVariableCollections.find((vr) => vr.name === (theme.group ?? theme.name));
+    const collection = figma.variables.getLocalVariableCollections().find((vr) => vr.name === (theme.group ?? theme.name));
```
six7, line 24: wouldnt this be slower? why dont we just fetch the collections outside like before?

SorsOps: This is likely because of   `const newCollection = figma.variables.createVariableCollection(theme.group ?? theme.name);` where we're mutating the collections. Assuming that the call is not expensive this fix is fine. If it isn't we would need to check ahead of time for overlap on collections

SorsOps: @swordEdge  does this seem to have any performance impact?

swordEdge: The count of collection is same as theme group. And normally theme group is not so much. So It won't affect any performance.
And for the logic, this find function should be inside of the loop. If it is outside, then we try to find the matching collection once when there is no collection.
So it will cause to create 2 or 3 collections for the same group.
E.g Group: Color, Theme: light, Theme: Dark
Group: Mode, Theme: Desktop, Theme: Mobile
These themes will create 4 collections. -> Color, Color, Mode, Mode.
But the result should be 2 collections Color, and Mode which have 2 modes. So It should be inside of the loop.

six7: Got it 👍

six7 (approved):

PR #2242, opened 2023-09-15
1925-scrolling-down-the-dropdown-menu-also-scrolls-the-plugin-screen

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/25951419/bd1f92f5-5540-44e5-9378-04b1c6a9fa24

### DIFF
--- a/src/tokens/core.ts
+++ b/src/tokens/core.ts
@@ -22,7 +22,7 @@ export const sizes = {
   8: '48px',
   10: '28px',
   controlMedium: '28px',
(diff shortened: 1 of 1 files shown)


six7 (commented):
One thing while you're here: The position of that menu looks weird. It should be placed below the trigger. Not to the right of it - is that an easy fix?

six7:
I also dont think the max height fix alone solves the problem 🤔 did you try on different window sizes?

SorsOps (approved):

PR #2244, opened 2023-09-18
583-add-an-easy-way-to-clear-search-field01

swordEdge:
https://github.com/tokens-studio/figma-plugin/assets/103296157/afad66d3-7370-449c-aab8-9a2ff6617b36

### DIFF
--- a/src/app/components/TokenFilter.tsx
+++ b/src/app/components/TokenFilter.tsx
@@ -3,6 +3,7 @@ import { useDebouncedCallback } from 'use-debounce';
 import { useDispatch, useSelector } from 'react-redux';
 import { MagnifyingGlassIcon } from '@radix-ui/react-icons';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


src/app/components/TokenFilter.tsx
```diff
@@ -61,6 +66,7 @@ const TokenFilter = () => {
         onChange={handleChange}
         placeholder={t('search') as string}
       />
+      <Box css={{ position: 'absolute', right: '$2', cursor: 'pointer' }} onClick={handleResetSearchString}><XCircleFillIcon /></Box>
```
six7, line 69: can you make this an `IconButton`? we should never use a `div` (Box) for an `onClick`. A `<div>` does not include focus handling like `<button>` does and will not be focusable with the keyboard

six7 (changes requested):

six7 (approved):

PR #2259, opened 2023-09-21
1821 cant rename stylesvariables after changing the parent name of a token group

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/262ddee1-7d21-4cf5-b324-04f9facd069e


src/app/store/models/reducers/tokenState/renameStyleIdsToCurrentTheme.ts
```diff
@@ -1,14 +1,19 @@
+import { TokensToRenamePayload } from '@/app/store/useTokens';
 import type { TokenState } from '../../tokenState';
 
-export function renameStyleIdsToCurrentTheme(state: TokenState, styleIds: string[], newName: string): TokenState {
+export function renameStyleIdsToCurrentTheme(state: TokenState, styleIds: string[], tokensToRename: TokensToRenamePayload[]): TokenState {
   const updatedThemes = [...state.themes];
+  const oldNameMapNewName = tokensToRename.reduce<Record<string, string>>((acc, curr) => {
```
six7, line 6: Instead of building the object like this here, why don't we just pass the object in like that? Instead of passing in an array like now. That would make the code more readable IMO

src/app/components/EditTokenForm.tsx
```diff
@@ -392,7 +392,7 @@ function EditTokenForm({ resolvedTokens }: Props) {
               renameTokensAcrossSets(oldName, newName, type, tokenSetsContainsSameToken);
             }
             if (confirmData.data.includes(StyleOptions.RENAME)) {
-              renameStylesFromTokens({ oldName, newName, parent: activeTokenSet });
+              renameStylesFromTokens([{ oldName, newName }], activeTokenSet);
```
six7, line 395: Like here, why don't we just do an object with `{ oldName: newName }` ?

six7 (changes requested):

keeganedwin:
Tested and confirming that this works when renaming the parent level of a token-group, then checking `sync variables` & `sync styles` in the modal which results in successful renaming. Even maintains the connection after renaming on elements that have variables or styles applied.

src/app/store/models/reducers/tokenState/renameStyleIdsToCurrentTheme.ts
robinhoodie0823: The structure of `renameStylesFromTokens` has been designed to get array params which can handle multiple styles changes. So it would be better to use array params.

src/app/components/EditTokenForm.tsx
robinhoodie0823: And here the same reason.

PR #2260, opened 2023-09-22
1841-token-transformer-expands-typography-tokens-even-when-resolvereferences=false

robinhoodie0823:


six7 (changes requested):
can you add a test for this case?

PR #2261, opened 2023-09-22
1930-active-theme-disabled-upon-editing-and-saving

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/8f011893-9678-4a00-9c31-bd1325a4f037

### DIFF
--- a/src/app/store/models/reducers/tokenState/saveTheme.ts
+++ b/src/app/store/models/reducers/tokenState/saveTheme.ts
@@ -33,7 +33,7 @@ export function saveTheme(state: TokenState, data: Payload): TokenState {
   });
 
   const newActiveTheme = state.activeTheme;
-  if (isActiveTheme) {
+  if (!isActiveTheme) {
     Object.keys(newActiveTheme).forEach((group) => {
       if (newActiveTheme[group] === themeId) {
         delete newActiveTheme[group];


src/app/store/models/reducers/tokenState/saveTheme.ts
```diff
@@ -33,7 +33,7 @@ export function saveTheme(state: TokenState, data: Payload): TokenState {
   });
 
   const newActiveTheme = state.activeTheme;
-  if (isActiveTheme) {
+  if (!isActiveTheme) {
```
six7, line 36: were we just doing it wrong before? or why is the fix to this so simple? any side effects that could be caused?

keeganedwin:
Tested and confirming that this works when:
1. Editing theme name 
2. Editing set status: source, enabled, disabled.

src/app/store/models/reducers/tokenState/saveTheme.ts
robinhoodie0823: Yes. The root cause was when try to edit active theme, then it all clears its activity inside the condition.

six7 (changes requested):
can you base the change on `main` and not `epic/performance` ? may be best to redo the branch given the change is so small

PR #2265, opened 2023-09-25
1935-unable-to-remove-license-key

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/6a5021af-6337-4653-ae68-8fdb88c32042


.changeset/applyto.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Changed default `Apply to` setting to `Selection`, this should not change any already stored `Apply to` setting
```
SorsOps, line 5: We should add a changeset to mention the issue with removing the licence key

src/app/components/AddLicenseKey/AddLicenseKey.tsx
```diff
@@ -49,15 +49,15 @@ export default function AddLicenseKey() {
 
   const removeKey = useCallback(async () => {
     if (licenseKeyError) {
-      dispatch.userState.removeLicenseKey('');
+      dispatch.userState.removeLicenseKey('empty');
```
SorsOps, line 52: I see we're using a hardcoded control value here, can we not rather draw a better distinction using the null type and then just doing a type comparison?

PR #2267, opened 2023-09-26
2151-adjust-shorthand-for-quick-edit-on-windows-to-ctrl+click

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/6483264c-1874-4257-b40b-755c5eb84005

### DIFF
--- a/src/app/components/MoreButton/MoreButton.tsx
+++ b/src/app/components/MoreButton/MoreButton.tsx
@@ -122,7 +122,8 @@ export const MoreButton: React.FC<Props> = ({ token, type, showForm }) => {
   );
 
   const handleTokenClick = React.useCallback((event: React.MouseEvent<HTMLButtonElement>) => {
-    if (event.metaKey || event.ctrlKey) {
+    const isMacBrowser = /Mac|iPod|iPhone|iPad/.test(navigator.platform);
+    if ((isMacBrowser && event.metaKey) || (!isMacBrowser && event.ctrlKey)) {
       handleEditClick();
     } else {
       handleClick(properties[0]);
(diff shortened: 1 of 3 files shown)


src/app/components/MoreButton/MoreButton.tsx
```diff
@@ -122,7 +122,7 @@ export const MoreButton: React.FC<Props> = ({ token, type, showForm }) => {
   );
 
   const handleTokenClick = React.useCallback((event: React.MouseEvent<HTMLButtonElement>) => {
-    if (event.metaKey || (event.ctrlKey && event.altKey)) {
+    if (event.metaKey || event.ctrlKey) {
```
six7, line 125: i think this wont work as it will disable `ctrl + click` on macos which is often used to right-click the context menu.

can we not check the operating system somehow?

six7 (changes requested):

src/app/components/MoreButton/MoreButton.tsx
```diff
@@ -122,7 +122,7 @@ export const MoreButton: React.FC<Props> = ({ token, type, showForm }) => {
...
       handleEditClick();
```
six7, line 126: test

six7 (changes requested):

src/app/components/MoreButton/MoreButton.tsx
SorsOps: hmmm, we should be able to do this with 
`const isMacBrowser = /Mac|iPod|iPhone|iPad/.test(navigator.platform);`

I'm just not sure if this is exposed correctly in the electron wrapper, we would need to verify

six7 (approved):

PR #2285, opened 2023-10-01
Add rotation property to number & dimension tokens

LukeFinch:
Currently, we can't set the rotation of a layer with a token.
This PR adds a new rotation property to numbers.

image

Rotation happens around the centre point of the layer - emulating the behaviour of setting the behaviour in the UI.


src/app/hooks/usePropertiesForType.test.ts
```diff
@@ -308,6 +308,10 @@ describe('usePropertiesForTokenType', () => {
           label: 'Background blur',
           name: Properties.backgroundBlur,
         },
+        {
+          label: 'Rotation',
+          name: Properties.rotation,
+        },
```
six7, line 314: Let's remove those from `dimension`. `rotation` is not a valid dimension unit, let's just keep it in `number`.

src/app/hooks/usePropertiesForType.ts
```diff
@@ -197,6 +197,10 @@ export function usePropertiesForTokenType(type: TokenTypes, value?: SingleToken[
             label: 'Background blur',
             name: Properties.backgroundBlur,
           },
+          {
+            label: 'Rotation',
+            name: Properties.rotation,
+          },
```
six7, line 203: This currently also makes it appear for `dimension`. Let's change it so it only appears for `number`.

src/plugin/removeValuesFromNode.ts
```diff
@@ -84,6 +85,11 @@ export default function removeValuesFromNode(node: BaseNode, prop: Properties) {
         node.effects = node.effects.filter((effect) => effect.type !== 'BACKGROUND_BLUR');
       }
       break;
+    case 'rotation':
+      if ('relativeTransform' in node && typeof node.relativeTransform !== 'undefined') {
```
six7, line 89: are we able to change rotation for instances / instance children? or would that lead to an error?

src/plugin/rotateNodes.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    expect(node.rotation).toBe(45);
+  });
+
+  it('should return undefined if node has no absoluteBoundingBox', () => {
+    expect(getCenterOfNode(nodeWithoutBounds)).toBeUndefined();
+  });
+
+  it('should parse string angle', () => {
+    // @ts-ignore next-line
+    rotateNode(node, '45px');
```
six7, line 85: this feels wrong. why would we want to use 45px as a rotation?

six7 (changes requested):

package-lock.json
six7, line 1: @LukeFinch can you remove this one? Given we didnt install anything this feels like a mistake

src/plugin/removeValuesFromNode.ts
LukeFinch: Oooh this is a very good point! It would lead to an error.
How's best to handle those kinds of errors? - I guess its the same as trying to change the size on an instance's child?

src/plugin/rotateNodes.test.ts
LukeFinch: Yeah this is a little wrong.
It seems like the number tokens are passing a string when they get sent across, and we need to parse them back into numbers.  This should really be '45' not '45px' to catch this issue

package-lock.json
LukeFinch: Not sure how it ended up changing. But resolved.

six7:
@LukeFinch can you get a changeset in this one and then resolve the merge conflict? then let's get that merged

six7 (approved):

PR #2289, opened 2023-10-01
prepare monorepo

### LINKED ISSUE #2288: Introduce Monorepo
Before starting Variables M2, we want to introduce a monorepo using `Turborepo`. This will allow us to separate the plugin's core functions into smaller pieces we can then reuse as individual packages.

In this first step, just move everything to a figma-plugin package folder in the monorepo and keep things as they are. We can then incrementally extract pieces into dedicated isolated packages.

six7:
Fixes #2288

Introduces Monorepo (basic) to better allow us split things into smaller packages.

This PR is best reviewed commit by commit.


turbo.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "globalDependencies": ["**/.env.*local"],
+  "pipeline": {
+    "build": {
+      "dependsOn": ["^build"],
+      "outputs": [".next/**", "!.next/cache/**"]
```
SorsOps, line 7: This probably needs dist for this to help it cache

turbo.json
```diff
@@ -0,0 +1,15 @@
...
+    },
+    "lint": {},
+    "dev": {
+      "cache": false,
+      "persistent": true
```
SorsOps, line 12: This sometimes causes an issue if it leaves a node process remaining for dependencies. We removed it to promote consistency in dev.

.github/workflows/node.js.yml
```diff
@@ -154,15 +154,15 @@ jobs:
 
       # 4: Build package
       - name: build:transform
-        run: npm run build-transform
+        run: cd packages/tokens-studio-for-figma && npm run build-transform
```
SorsOps, line 157: probably better to use ` working-directory` for this

turbo.json
six7: Added `dist/**` in latest commit

turbo.json
six7: Removed that line in latest commit

.github/workflows/node.js.yml
six7: Changed to working-directory for all instances where I cd'd before in latest commit

PR #2293, opened 2023-10-02
1821-cant-rename-stylesvariables-after-changing-the-parent-name-of-a-token-group

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/262ddee1-7d21-4cf5-b324-04f9facd069e

### DIFF
--- a/src/app/components/DownshiftInput/mentions.css
+++ b/src/app/components/DownshiftInput/mentions.css
@@ -69,7 +69,7 @@
   border: 1px solid var(--colors-borderDefault);
   border-radius: var(--radii-medium);
   box-shadow: var(--shadows-contextMenu);
-  background: var(--colors-contextMenuBg);
+  background: var(--colors-bgDefault);
   cursor: pointer;
   z-index: 10;
 }
--- a/src/app/components/DownshiftInput/mentions.css
+++ b/src/app/components/DownshiftInput/mentions.css
@@ -69,7 +69,7 @@
   border: 1px solid var(--colors-borderDefault);
   border-radius: var(--radii-medium);
   box-shadow: var(--shadows-contextMenu);
-  background: var(--colors-bgDefault);
+  background: var(--colors-contextMenuBg);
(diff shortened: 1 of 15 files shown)


src/app/components/DownshiftInput/mentions.css
```diff
@@ -69,7 +69,7 @@
   border: 1px solid var(--colors-borderDefault);
   border-radius: var(--radii-medium);
   box-shadow: var(--shadows-contextMenu);
-  background: var(--colors-bgDefault);
+  background: var(--colors-contextMenuBg);
```
six7, line 72: this seems like a mistake, we only recently changed this

src/app/store/models/reducers/tokenState/renameVariableNamesToThemes.ts
```diff
@@ -1,13 +1,18 @@
...
       Object.entries(updatedTokens).forEach(([key]) => {
-        if (key === oldName) {
-          updatedTokens[newName] = updatedTokens[oldName];
-          delete updatedTokens[oldName];
+        if (oldNameMapNewName[key]) {
+          updatedTokens[oldNameMapNewName[key]] = updatedTokens[key];
+          delete updatedTokens[key];
         }
       });
     }
```
six7, line 18: lets rename this variable to something more understandable: `oldToNewNameMap` for example

six7 (changes requested):

src/app/components/DownshiftInput/mentions.css
robinhoodie0823: Reverted.

src/app/store/models/reducers/tokenState/renameVariableNamesToThemes.ts
robinhoodie0823: Updated variable name to `oldToNewNameMap`

six7 (approved):

PR #2294, opened 2023-10-03
1935-unable-to-remove-license-key-rebase

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/6a5021af-6337-4653-ae68-8fdb88c32042

### DIFF
--- a/src/app/components/AddLicenseKey/AddLicenseKey.tsx
+++ b/src/app/components/AddLicenseKey/AddLicenseKey.tsx
@@ -49,7 +49,7 @@ export default function AddLicenseKey() {
 
   const removeKey = useCallback(async () => {
     if (licenseKeyError) {
-      dispatch.userState.removeLicenseKey('empty');
+      dispatch.userState.removeLicenseKey(undefined);
     } else {
       const confirmation = await confirm({
         text: t('confirmRemove') as string,
@@ -57,7 +57,7 @@ export default function AddLicenseKey() {
         confirmAction: t('removeKey') as string,
       });
       if (confirmation) {
-        dispatch.userState.removeLicenseKey('empty');
+        dispatch.userState.removeLicenseKey(undefined);
         removeAccessToFeatures();
       }
     }
@@ -129,11 +129,7 @@ export default function AddLicenseKey() {
             type="password"
             inputRef={inputEl}
             isMasked
-            value={
-              newKey === 'empty'
-                ? ''
-                : newKey || ''
-            }
+            value={newKey || ''}
             onChange={onLicenseKeyChange}
             error={licenseKeyError}
           />
--- a/src/app/components/AddLicenseKey/AddLicenseKey.tsx
+++ b/src/app/components/AddLicenseKey/AddLicenseKey.tsx
@@ -49,7 +49,7 @@ export default function AddLicenseKey() {
 
   const removeKey = useCallback(async () => {
     if (licenseKeyError) {
-      dispatch.userState.removeLicenseKey('');
+      dispatch.userState.removeLicenseKey('empty');
     } else {
       const confirmation = await confirm({
         text: t('confirmRemove') as string,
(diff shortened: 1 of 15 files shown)


src/app/components/AddLicenseKey/AddLicenseKey.tsx
```diff
@@ -129,7 +129,11 @@ export default function AddLicenseKey() {
             type="password"
             inputRef={inputEl}
             isMasked
-            value={newKey || ''}
+            value={
```
SorsOps, line 132: Instead of using this hardcoded value, can we instead use null and the empty string to properly distinguish between the two ?

robinhoodie0823: @SorsOps I also tried to use `null` and empty string to distinguish the two situations. But the empty string automatically converted to `null` and it's difficult to distinguish by using the two values you proposed. That is why I used the hardcoded value.

six7: can you change the logic in `[removeLicenseKey](https://github.com/tokens-studio/figma-plugin/blob/abf1cb0be9605817287f1e988e84529793cfb902/src/app/store/models/userState.ts#L128)` so it takes the string as optional?

if there is a license key, we trigger `[removeLicense](https://github.com/tokens-studio/figma-plugin/blob/abf1cb0be9605817287f1e988e84529793cfb902/src/app/store/models/userState.ts#L132C38-L132C38)` on line 132 of `userState.tsx

and then we continue as before, we reset the local license key.

that would mean we dont need to pass `empty`, and we would still try to detach the key, but it is no longer a requirement. we would still continue with clearing the key locally, which is what they want to do.

six7 (changes requested):

src/app/components/AppContainer/startupProcessSteps/addLicenseFactory.ts
```diff
@@ -8,16 +8,16 @@ export function addLicenseFactory(dispatch: Dispatch, params: StartupMessage) {
   return async () => {
     const { user } = params;
     let { licenseKey } = params;
-    const isFirstLoading = licenseKey !== 'empty';
+    const { islicenseKeyRemoved } = params;
```
SorsOps, line 11: This feels weird that we're tracking the licence key being removed instead of the initial loading  ? We never actually change the value of isLicenseKeyRemoved, so it feels kind of messy that we have 

1. The initial load (assuming empty)
2. Adding a key
3. A user removes the licence key 

If a user restarts now we have stored the isLicenseKeyRemoved value and the startup logic is now different. I feel like we should rather send an `initialLoad` state var, and then have a side effect of turning this off after we perform the first execution of this function

src/app/components/AppContainer/startupProcessSteps/addLicenseFactory.ts
```diff
@@ -8,7 +8,9 @@ export function addLicenseFactory(dispatch: Dispatch, params: StartupMessage) {
   return async () => {
     const { user } = params;
     let { licenseKey } = params;
-    if (!licenseKey) {
+    const { initialLoad } = params;
+
+    if (licenseKey === null && initialLoad.toString() !== 'true') {
```
SorsOps, line 13: Can't we do something like ` if ((licenseKey === null|| licenseKey ===undefined) && !initialLoad) {`

Or do we gain anything from the stringification of the boolean value ?

robinhoodie0823: The value of `initialLoad` that we get is the string type.

src/app/components/AppContainer/startupProcessSteps/addLicenseFactory.ts
```diff
@@ -8,9 +8,9 @@ export function addLicenseFactory(dispatch: Dispatch, params: StartupMessage) {
   return async () => {
     const { user } = params;
     let { licenseKey } = params;
-    const { initialLoad } = params;
+    const { initialLoad } = JSON.parse(JSON.stringify(params));
```
SorsOps, line 11: Why are we doing this ? This should result in no change

robinhoodie0823: Resolved.

SorsOps (approved):

six7 (approved):

PR #2296, opened 2023-10-03
1841-token-transformer-expands-typography-tokens-even-when-resolvereferences=false

robinhoodie0823:


SorsOps (commented):
This looks fine. If tests pass, lets merge

six7 (changes requested):
can you add a change of the test files for this? token-transformer has quite good test coverage so ideally we'd add a new test case for this.

six7 (approved):

PR #2322, opened 2023-10-13
Text type token text string being treated as number and not being set properly in variable

### LINKED ISSUE #2100: "Text" type token text string being treated as number and not being set properly in variable
**Describe the bug**
messing around with components, variables, and tokens; i created several "text" tokens and put hex values in them. most of them work fine, but a couple values are being treated as numbers.
"003e78" is being resolved to "3e+78"
"000000" is being resolved to "0"
then when variables are created from these, neither of those values are being used in created variable. rather it seems the variable is falling back to a different token with a "valid" value from a different set.
to ensure this was not a file dependent issue, i created a brand new design file with an empty tokens file, made a single token with one of the offending values, then created variables from the tokens. the issue was present (see second screenshot)

**To Reproduce**
1. create a text token with a value that would be converted to a number if you typed it in excel, like "003e78" or "000000"
2. create variable from the token
3. see that the desired token value did not correctly get placed in the variable

**Expected behavior**
i would expect the text token value to be treated as a string without trying to reduce or resolve it

image

image

**Figma file (optional)**
https://www.figma.com/file/TlqW1qEpxJYksDfVbTxfae/Untitled?type=design&node-id=0%3A1&mode=design&t=PT3z9vqHpPtWicuq-1

**JSON (optional)**
{
  "texttest": {
    "value": "003e78",
    "type": "text"
  }
}

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/8cc4e13f-7bc8-4b8c-a966-d0bff04af7df

Closes #2100


src/utils/TokenResolver.ts
```diff
@@ -45,7 +46,7 @@ class TokenResolver {
     const resolvedTokens: ResolveTokenValuesResult[] = [];
 
     for (const token of this.tokens) {
-      const resolvedValue = this.resolveReferences(token);
+      const resolvedValue = token.type === TokenTypes.TEXT ? token : this.resolveReferences(token);
```
SorsOps, line 49: Would this not break if you wanted to have a text string that had references to other text strings?

src/utils/TokenResolver.ts
```diff
@@ -45,7 +46,7 @@ class TokenResolver {
     const resolvedTokens: ResolveTokenValuesResult[] = [];
 
     for (const token of this.tokens) {
-      const resolvedValue = this.resolveReferences(token);
```
SorsOps, line 48: Could you add a changeset for this PR?

robinhoodie0823: Added

six7:
Tests are failing here

six7:
Tests are failing here

src/utils/TokenResolver.ts
```diff
@@ -46,6 +46,7 @@ class TokenResolver {
 
     for (const token of this.tokens) {
       const resolvedValue = this.resolveReferences(token);
+      // const resolvedValue = token.type === TokenTypes.TEXT ? token : this.resolveReferences(token);
```
SorsOps, line 49: Why is this still uncommented?

PR #2336, opened 2023-10-25
When switching themes switch the figma native theme mode selector if possible

### LINKED ISSUE #2187: When switching Themes, switch the Figma native theme mode selector if possible

ultraprinse:
https://github.com/tokens-studio/figma-plugin/assets/103296157/8769ec50-0d13-4b1c-9d0a-6797e8f65115

Closes #2187


src/AsyncMessageChannel.ts
```diff
@@ -122,6 +122,16 @@ export class AsyncMessageChannel {
       this.attachMessageListener((msg: IncomingMessageEvent<AsyncMessageResults & { type: Message['type'] }>['data']['pluginMessage']) => {
         if (msg.id === messageId) {
           if ('message' in msg) {
+            if (msg.message.type === 'async/get-theme-info') {
+              if ('activeTheme' in msg.message && 'themes' in msg.message) {
+                const { activeTheme } = msg.message;
```
SorsOps, line 127:
```suggestion
                  const { activeTheme, themes } = msg.message;
```

src/AsyncMessageChannel.ts
```diff
@@ -122,6 +122,16 @@ export class AsyncMessageChannel {
...
+                const { themes } = msg.message;
```
SorsOps, line 128:
```suggestion
const themeToFind = Object.values(activeTheme)[0];
```

src/AsyncMessageChannel.ts
```diff
@@ -122,6 +122,16 @@ export class AsyncMessageChannel {
...
+                const activeMode = themes.filter((theme) => theme.id === Object.values(activeTheme)[0])[0];
```
SorsOps, line 129:
```suggestion
                const activeMode = themes.find((theme) => theme.id === themeToFind);
```

SorsOps (changes requested):

SorsOps:
Can you add a changeset so we can inform the users about this change?

Check to see if the fixes for the lookup fix the OOM error, otherwise we'll have to double check why that's occurring

six7:
Tests are failing here

src/AsyncMessageChannel.ts
```diff
@@ -122,6 +122,18 @@ export class AsyncMessageChannel {
       this.attachMessageListener((msg: IncomingMessageEvent<AsyncMessageResults & { type: Message['type'] }>['data']['pluginMessage']) => {
         if (msg.id === messageId) {
           if ('message' in msg) {
+            if (msg.message.type === 'async/get-theme-info') {
```
six7, line 125: i dont think this belongs here. this file doesnt have any of the other message handlers, this needs to go elsewhere where we define those.

https://github.com/tokens-studio/figma-plugin/blob/1424c9f8e3c8959e4595d5a7a97d7309a0fdf67f/src/app/components/Initiator.tsx#L31

this feels like the right place

src/AsyncMessageChannel.ts
```diff
@@ -122,6 +122,18 @@ export class AsyncMessageChannel {
...
+              if ('activeTheme' in msg.message && 'themes' in msg.message) {
+                const { activeTheme, themes } = msg.message;
+                const themeToFind = Object.values(activeTheme)[0];
+                const activeMode = themes.find((theme) => theme.id === themeToFind);
+                if (figma.currentPage.children && figma.currentPage.children.length > 0) {
+                  figma.currentPage.children.forEach((frame) => {
+                    frame.setExplicitVariableModeForCollection(activeMode?.$figmaCollectionId ?? '', activeMode?.$figmaModeId ?? '');
```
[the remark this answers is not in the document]
six7: yeah it could be that the information is outdated and then this will error. they'll likely not be invalid or null but it could be that it doesnt exist anymore causing an error.

let's check for if a) those are set and b) wrap so that we dont crash

src/AsyncMessageChannel.ts
```diff
@@ -122,6 +122,18 @@ export class AsyncMessageChannel {
       this.attachMessageListener((msg: IncomingMessageEvent<AsyncMessageResults & { type: Message['type'] }>['data']['pluginMessage']) => {
         if (msg.id === messageId) {
           if ('message' in msg) {
+            if (msg.message.type === 'async/get-theme-info') {
+              if ('activeTheme' in msg.message && 'themes' in msg.message) {
+                const { activeTheme, themes } = msg.message;
+                const themeToFind = Object.values(activeTheme)[0];
+                const activeMode = themes.find((theme) => theme.id === themeToFind);
+                if (figma.currentPage.children && figma.currentPage.children.length > 0) {
+                  figma.currentPage.children.forEach((frame) => {
```
six7, line 131: this ignores the issue requirement completely and sets an active variable mode for **every** frame on the current page. this is not what we need. we need to apply way more logic here than that.

https://github.com/tokens-studio/figma-plugin/issues/2187

the issue states that it should have different behavior depending on the apply mode

src/AsyncMessageChannel.ts
```diff
@@ -122,6 +122,18 @@ export class AsyncMessageChannel {
       this.attachMessageListener((msg: IncomingMessageEvent<AsyncMessageResults & { type: Message['type'] }>['data']['pluginMessage']) => {
         if (msg.id === messageId) {
           if ('message' in msg) {
+            if (msg.message.type === 'async/get-theme-info') {
+              if ('activeTheme' in msg.message && 'themes' in msg.message) {
+                const { activeTheme, themes } = msg.message;
+                const themeToFind = Object.values(activeTheme)[0];
```
six7, line 128: we do not just want to switch the first theme, but all theme groups that are active

six7 (changes requested):

PR #2340, opened 2023-10-27
Some tokens are falsely labeled as broken

### LINKED ISSUE #2301: Some tokens are falsely labeled as broken
For composite token types:
- Typography
- Border
- Box shadow
There is an issue at the 3rd level of referencing which appears broken but is not really broken. The token when referenced again does not show as broken.

Detailed ways of reproduction given in the Figma file below:
**Figma File**
https://www.figma.com/file/3wBqHDuy4DnTdlcfUOQYec/Github-issue-%232301?type=design&node-id=0%3A1&mode=design&t=a0u6uR80uQEUjKkc-1

----
<details><summary>Original Issue</summary>
**Describe the bug**
I have transition tokens defined in JSON and they are used to generate CSS variables. They follow a different structure as in [Design Tokens Draft Format](https://tr.designtokens.org/format/#transition)

```JSON
    "animation-timing": {
      "50": { "value": "cubic-bezier(0.25,1,0.3,1)", "type": "animationTiming" }
    },
    "animation-duration": {
      "50": { "value": "0.25", "type": "animationDuration" },
      "100": { "value": "0.5", "type": "animationDuration" }
    },
    "animation-delay": {
      "none": { "value": "0ms", "type": "animationDelay" }
    },
    "transition": {
      "50": {
        "value": {
          "animation-timing": "{animation-timing.50}",
          "animation-duration": "{animation-duration.50}",
          "animation-delay": "{animation-delay.none}"
        },
        "type": "transition"
      },
      "100": {
        "value": {
          "animation-timing": "{animation-timing.50}",
          "animation-duration": "{animation-duration.100}",
          "animation-delay": "{animation-delay.none}"
        },
        "type": "transition"
      }
    }
```

This animation definition is then linked to the abstract `form-control` token:
```JSON
    "form-control": {
      "xs": {    
        "transition": { "value": "{transition.50}", "type": "transition" }
      },
    }
```

And then it is linked to components that share form-control definition:
```JSON
    "button": {
      "xs": {
        "transition": {
          "value": "{form-control.xs.transition}",
          "type": "transition"
        }
      },
```

When I look at the plugin, the first component to have this transition is labeled as it has a broken reference (A). The rest of the components show the reference is correct (B). 
image

When I delete the first component in the list (e.g. I delete a button), the broken reference passes to another component (e.g., icon-button) and shows other references as correct:
image

**To Reproduce**
Not sure how to reproduce it, see the description of the behavior

**Expected behavior**
The broken reference should not be shown, as the reference is not broken.

see above

**JSON (optional)**
I'm on the Tokens Studio Slack, I can send the file via Slack if needed

**Figma File**
https://www.figma.com/file/3wBqHDuy4DnTdlcfUOQYec/Github-issue-%232301?type=design&node-id=0%3A1&mode=design&t=a0u6uR80uQEUjKkc-1

</details>

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/86917098-1cba-48e8-a937-5d8e5e92cacc

Closes #2301


SorsOps:
@robinhoodie0823 Can you provide the testing for this PR inside for a valid case where the resolution algorithm  is showing a falsely labelled broken token ? This just looks like a PR regressing the functionality to ever show an example of a broken reference

six7:
Tests are failing here

six7:
@robinhoodie0823 any updates on this one?

robinhoodie0823:
No, I'm working on other tickets now. I will fix asap.

src/utils/TokenResolver.test.ts
```diff
@@ -241,7 +241,6 @@ const output = [
     value: 6,
   },
   {
-    failedToResolve: true,
```
six7, line 244: Why remove this? This token should be classified as failedToResolve because it did fail to resolve. We cant just remove this.

src/utils/TokenResolver.ts
```diff
@@ -187,10 +188,10 @@ class TokenResolver {
         }
       } else {
         // If it's not, we mark it as failed to resolve
```
six7, line 190: We cant uncomment this. This is important to understand that some tokens failed to resolve.

src/utils/TokenResolver.ts
```diff
@@ -187,10 +188,10 @@ class TokenResolver {
...
-        const hasFailingReferences = !AliasRegex.test(JSON.stringify(finalValue));
+        // const hasFailingReferences = !AliasRegex.test(JSON.stringify(finalValue));
 
         resolvedToken = {
```
six7, line 193: Same here, we need hasFailingReferences.

six7 (changes requested):
As @SorsOps already stated we need a test for this scenario. 

Also, it seems you only removed the check if something is failing. What needs to be fixed is the fact that some tokens are falsely labeled as broken, not that we dont want tokens to show up as broken anymore.

The problem according to the issue is about typography, border, shadow tokens.

src/utils/TokenResolver.ts
```diff
@@ -187,7 +190,11 @@ class TokenResolver {
         }
       } else {
         // If it's not, we mark it as failed to resolve
-        const hasFailingReferences = !AliasRegex.test(JSON.stringify(finalValue));
+
+        const stringifiedValue = JSON.stringify(finalValue);
+        const convertedValue = stringifiedValue.substring(1, stringifiedValue.length - 1);
```
six7, line 195: I think this won't work for tokens that have another object inside, such as composite tokens or shadow tokens. Is there another way you can stringify the object that does not produce curly brackets?

For example right now the transformer tests are failing because of this

six7 (changes requested):

six7 (approved):
✨

PR #2361, opened 2023-11-09
Feat/ reintroduce update on change

LukeFinch:
Brings back the updateOnChange behavior when editing a token.

- If the user has the setting for update on change AND - their apply mode is set to page or selection

When: tokens are edited, update the nodes on either the page or their selection.

### DIFF
--- a/.changeset/sweet-walls-prove.md
+++ b/.changeset/sweet-walls-prove.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fixed an issue with Update on change. It's now working again, but disabled for new users as intended.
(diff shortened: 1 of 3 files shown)


six7 (changes requested):
Can you check if this setting is enabled for new users? I'd want it to be disabled for them. We know users don't find this menu and I want to avoid a scenario where we again hear from users how bad performance is just because they have settings enabled that cause this

LukeFinch:
image

The latest commit ensures the setting is disabled by default for new users - I tested by changing the plugin ID in the manifest to something new, this creates all-new default settings. 

One future enhancement might be for us to have a 'restore default settings' option in the settings.

six7 (approved):

PR #2368, opened 2023-11-14
For-multi-value-number-or-dimension-or-spacing-or-sizing-tokens-we-should-skip-them

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/41b901f1-35cb-4b63-ae1f-66bd05c02965

### DIFF
--- a/src/app/store/useTokens.test.tsx
+++ b/src/app/store/useTokens.test.tsx
@@ -553,6 +553,9 @@ describe('useToken test', () => {
             { name: 'borderWidth-1', value: '{base.base-350} {base.base-187} {base.base-50}', type: TokenTypes.BORDER_WIDTH },
             { name: 'borderWidth-2', value: '{base.base-50} {base.base-187} {base.base-150}', type: TokenTypes.BORDER_WIDTH },
             { name: 'borderWidth-3', value: '{base.base-187}', type: TokenTypes.BORDER_WIDTH },
+            { name: 'borderWidth-3', value: '{base.base-187}', type: TokenTypes.BORDER_WIDTH },
+            { name: 'text', value: 'text value', type: TokenTypes.TEXT },
+            { name: 'color', value: 'rgba(20, 80, 200, 1)', type: TokenTypes.COLOR },
           ],
         },
       },
@@ -578,6 +581,8 @@ describe('useToken test', () => {
           { name: 'borderRadius-3', value: '{base.base-150}', type: TokenTypes.BORDER_RADIUS },
           { name: 'spacing-3', value: '{base.base-50}', type: TokenTypes.SPACING },
           { name: 'borderWidth-3', value: '{base.base-187}', type: TokenTypes.BORDER_WIDTH },
+          { name: 'text', value: 'text value', type: TokenTypes.TEXT },
+          { name: 'color', value: 'rgba(20, 80, 200, 1)', type: TokenTypes.COLOR },
         ],
       });
     });
--- a/src/app/store/useTokens.test.tsx
+++ b/src/app/store/useTokens.test.tsx
@@ -544,10 +544,15 @@ describe('useToken test', () => {
           global: [{ name: 'white', value: '#ffffff', type: TokenTypes.COLOR }, { name: 'headline', value: { fontFamily: 'Inter', fontWeight: 'Bold' }, type: TokenTypes.TYPOGRAPHY }, { name: 'shadow', value: '{shadows.default}', type: TokenTypes.BOX_SHADOW }],
           light: [
             { name: 'bg.default', value: '#ffffff', type: TokenTypes.COLOR },
-            { name: 'spacing-1', value: '{base.base-0} {base.base-25}', type: TokenTypes.SPACING },
-            { name: 'spacing-2', value: '{base.base-50} {base.base-150}', type: TokenTypes.SPACING },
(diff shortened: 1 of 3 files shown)


src/app/store/useTokens.tsx
```diff
@@ -399,6 +399,19 @@ export default function useTokens() {
       name: BackgroundJobs.UI_CREATEVARIABLES,
       isInfinite: true,
     });
+
+    const tempTokens: Record<string, AnyTokenList> = {};
+    Object.keys(tokens).forEach((tokenSetKey) => {
```
SorsOps, line 404: We can do all of this cleaner with a `Object.entries(tokens).reduce(...)`

src/app/store/useTokens.tsx
```diff
@@ -399,6 +399,19 @@ export default function useTokens() {
       name: BackgroundJobs.UI_CREATEVARIABLES,
       isInfinite: true,
     });
+
```
SorsOps, line 402: Can you link the ticket for this PR? I just want to double check the context

src/app/store/useTokens.tsx
```diff
@@ -399,6 +399,19 @@ export default function useTokens() {
...
+    const tempTokens: Record<string, AnyTokenList> = {};
+    Object.keys(tokens).forEach((tokenSetKey) => {
+      const tokenList = tokens[tokenSetKey].filter((tokenItem) => {
+        if ((typeof tokenItem.value === 'string')) {
+          const resolvedValue = getAliasValue(tokenItem.value, tokensContext.resolvedTokens) || '';
+          return !resolvedValue.toString().trim().includes(' ');
```
SorsOps, line 408: Can we test this function to assert that this is processing the tokens correctly ?

src/app/store/useTokens.tsx
SorsOps: We should also add a changeset to indicate this fix for the users

SorsOps (changes requested):

src/app/store/useTokens.test.tsx
```diff
@@ -339,19 +339,19 @@ describe('useToken test', () => {
...
-      await result.current.remapTokensInGroup({ oldGroupName: 'old.', newGroupName: 'new.' });
-    });
-    expect(messageSpy).toBeCalledWith({
-      type: AsyncMessageTypes.BULK_REMAP_TOKENS,
-      oldName: 'old.',
-      newName: 'new.',
-      updateMode: UpdateMode.SELECTION,
-    });
-  });
+  // it('remapTokensInGroup', async () => {
```
SorsOps, line 342: Can we turn these back on ?

src/app/store/useTokens.tsx
```diff
@@ -393,12 +393,31 @@ export default function useTokens() {
     });
   }, []);
 
+  const filterMultiValueTokens = useCallback(() => {
+    const tempTokens = Object.entries(tokens).reduce((tempTokens, [tokenSetKey, tokenList]) => {
+      const filteredTokenList = tokenList.filter((tokenItem) => {
+        if (typeof tokenItem.value === 'string') {
+          const resolvedValue = getAliasValue(tokenItem.value, tokensContext.resolvedTokens) || '';
```
SorsOps, line 400: Should we not be filtering based on the type of token as well. The linked issue describes its usage for dimension and sizing ,other token types will likely be allowed to have spaces in them

six7 (approved):

src/app/store/useTokens.test.tsx
```diff
@@ -530,6 +530,52 @@ describe('useToken test', () => {
...
+    });
+
+    beforeEach(() => {
+      resetStore();
+      result = renderHook(() => useTokens(), {
+        wrapper: ({ children }: { children?: React.ReactNode }) => <Provider store={tokenMockStore}>{children}</Provider>,
+      }).result;
+    });
+
+    it('skip multi value tokens', async () => {
```
six7, line 563: Can you add a multi value token here to see it gets filtered out? and one `TEXT` token that also has a space inside but is not getting filtered out due to the type?

six7 (changes requested):

src/app/store/useTokens.tsx
```diff
@@ -393,12 +393,32 @@ export default function useTokens() {
     });
   }, []);
 
+  const filterMultiValueTokens = useCallback(() => {
+    const validTokenTypes = [TokenTypes.DIMENSION, TokenTypes.BORDER_RADIUS, TokenTypes.BORDER, TokenTypes.BORDER_WIDTH, TokenTypes.SPACING];
```
SorsOps, line 397: This can just be a constant , we don't need to place this on the stack each time we call this function

src/app/store/useTokens.test.tsx
robinhoodie0823: You mean, I need to add other multi value tokens besides the `SPACING` token?

six7: ah, i didnt see that you had one in there already. i think the tests could use a specific example.

here you have it:

```
            { name: 'spacing-1', value: '{base.base-50} {base.base-150}', type: TokenTypes.SPACING },
```

if it was just 3 tokens instead of 30 then the test would be more readable

six7 (approved):

SorsOps (approved):

PR #2372, opened 2023-11-15
Upgrade to React 18

LukeFinch:
Upgrades to React 18 along with some other older dependencies.
This will enable us to use components from our shared UI package, and use more modern libraries.

I plan to keep this PR open for a short while as an internal beta, to check for any issues we may come across with unforeseen issues between react 17 and 18.


six7:
Closing as this is now part of the `release-139` branch

PR #2404, opened 2023-12-02
139: integrate @tokens-studio/ui

six7:
Integrates shared components targeted for the 139 branch instead of using custom components


packages/tokens-studio-for-figma/src/app/components/IconButton.tsx
LukeFinch, line 1: I'm guessing the DS one lacks all the props we use on the plugin?

packages/tokens-studio-for-figma/src/app/components/Input.tsx
```diff
@@ -100,20 +100,20 @@ const StyledSuffix = styled('button', {
...
     },
   },
 });
 
 const StyledPrefix = styled('div', {
   padding: '0 $3',
-  height: '28px',
+  height: '$controlSmall',
   flexShrink: 0,
   border: '1px solid $borderMuted',
```
LukeFinch, line 118: Need a border token for the width? - Think this will come with the rework Keegan is on

packages/tokens-studio-for-figma/src/app/components/InspectorMultiView.tsx
```diff
@@ -124,18 +124,21 @@ export default function InspectorMultiView({ resolvedTokens, tokenToSearch }: {
               id="selectAll"
               onCheckedChange={handleSelectAll}
             />
-            <Label htmlFor="selectAll" css={{ fontSize: '$small', fontWeight: '$bold' }}>
+            <Label htmlFor="selectAll" css={{ fontSize: '$small', fontWeight: '$sansBold' }}>
```
LukeFinch, line 127: Wonder if we should apply global css rules to Labels and similar HTML primitives? 

Feel like all labels should be small and bold 🤔

packages/tokens-studio-for-figma/src/app/components/ManageThemesModal/ThemeStyleManagementCategoryStyleEntry.tsx
```diff
@@ -1,7 +1,6 @@
 import React, { useCallback } from 'react';
-import { LinkBreak1Icon, LinkBreak2Icon } from '@radix-ui/react-icons';
-import { IconArrowRight } from '@/icons';
-import Box from '../Box';
+import { ArrowRightIcon, LinkBreak1Icon, LinkBreak2Icon } from '@radix-ui/react-icons';
```
LukeFinch, line 2: For a future issue, let's look into consolidating the icons into one single set.

packages/tokens-studio-for-figma/src/app/components/ManageThemesModal/ThemeVariableManagement.tsx
```diff
@@ -126,14 +126,15 @@ export const ThemeVariableManagement: React.FC<React.PropsWithChildren<React.Pro
       label={(
         <Stack direction="row" gap={2} align="center">
           <Heading size="medium">
-            Variable
+            Variables
```
LukeFinch, line 129: Translation?

packages/tokens-studio-for-figma/src/app/components/SecondScreen/index.tsx
```diff
@@ -26,19 +26,19 @@ export const StyledBetaBadge = styled('span', {
   backgroundColor: '$accentBg',
   lineHeight: 1,
   color: '$fgDefault',
-  fontWeight: '$bold',
+  fontWeight: '$sansBold',
   textTransform: 'uppercase',
   border: '1px solid transparent',
 });
 
 const StyledP = styled('p', {
```
LukeFinch, line 34: Would be nice to abstract all these out into a generic set of typography patterns in the future

packages/tokens-studio-for-figma/src/app/components/IconButton.tsx
six7: yeah, that + i wanted to not change all the imports, i think that's easier to review in one big PR where all we do is change all imports.

we're sometimes using `href` on a button (which we should solve with as="a") or in the old one we still have the `badge` prop (a blue dot indicating new activity, e.g. push pull icon buttons). we should likely just have a wrapping component for those that wrap an iconbutton and add the new activity dot

packages/tokens-studio-for-figma/src/app/components/Input.tsx
six7: technically it'd be better, agree (might need to do that in a dedicated PR where we change all 1px occurences)

packages/tokens-studio-for-figma/src/app/components/InspectorMultiView.tsx
six7: The `ds` Labels have that already, might need to do this in a focused PR where all we do is replace all Labels

packages/tokens-studio-for-figma/src/app/components/ManageThemesModal/ThemeVariableManagement.tsx
six7: i didnt translate what wasnt translated in this PR. maybe we should do a dedicated one checking all files to add any missing translations, seems like there's a few areas that are untranslated

packages/tokens-studio-for-figma/src/app/components/SecondScreen/index.tsx
six7: 💯  we should just be able to set these as variants on the `Text` or `Heading` components or create a dedicated component. For this one it feels like this should just be part of the `Badge` component

PR #2405, opened 2023-12-04
allow for bulk remap with variable and styles

robinhoodie0823:


.changeset/unlucky-forks-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tokens-studio/figma-plugin': patch
+---
+
+Allow bulk remapping with variables and styles
```
six7, line 5:
```suggestion
Introduces support to bulk remap variables and styles
```

src/app/components/InspectorTokenSingle.tsx
```diff
@@ -46,6 +46,10 @@ export default function InspectorTokenSingle({
 
   const tokenToDisplay = React.useMemo(() => {
     if (token.resolvedValue) {
+      const resolvedToken = getTokenValue(token.value, resolvedTokens);
+      if (resolvedToken && token.appliedType === 'variable') {
+        return { name: token.value, value: resolvedToken.value, type: property };
```
six7, line 51: Can you explain this change? What's the intent?

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -37,6 +39,44 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
...
+        if (Object.keys(tokens).length === 0) {
+          if (getAppliedVariablesFromNode(node).length > 0) {
+            const { name: variableName, type: variableType } = getAppliedVariablesFromNode(node)[0];
+
+            const savedVariableData = node.getSharedPluginData(SharedPluginDataNamespaces.VARIABLES, variableType);
+            if (savedVariableData.length > 0) {
+              if (savedVariableData.includes(oldName)) {
+                const newValue = JSON.parse(savedVariableData).replace(oldName, newName);
+                const jsonValue = JSON.stringify(newValue);
+                node.setSharedPluginData(SharedPluginDataNamespaces.VARIABLES, variableType, jsonValue);
```
six7, line 52: I think this is a misunderstanding. We should not get or set variables on shared plugin data. Variables can be set directly using boundVariables and the other variable functions. Here you've applied shared plugin data for variables instead.

There's no need to do this. Let's just read what variables are applied on the layer (Inspect already shows this!), then take the names of those, remap parts of the name, and then find the connected variables to replace it with.

For example, if a variable named colors.blue.500 is applied, and another layer has colors.blue.400, and I remap 'blue' to 'red' then we'd need to
- look up tokens named 'colors.red.400'
- and 500 and fetch the connected variables
- replace the applied variables with those

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -37,6 +39,44 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
...
+            if (savedStyleData.length > 0) {
+              if (savedStyleData.includes(oldName)) {
+                const newValue = JSON.parse(savedStyleData).replace(oldName, newName);
+                const jsonValue = JSON.stringify(newValue);
+                node.setSharedPluginData(SharedPluginDataNamespaces.STYLES, styleType, jsonValue);
+              }
+            } else if (styleName.includes(oldName)) {
+              const newValue = styleName.replace(oldName, newName);
+              const jsonValue = JSON.stringify(newValue);
+              node.setSharedPluginData(SharedPluginDataNamespaces.STYLES, styleType, jsonValue);
```
six7, line 75: Same for styles

six7 (changes requested):

package.json
```diff
@@ -192,12 +192,14 @@
...
     "html-webpack-plugin": "^5.0.0",
     "husky": "^3.0.2",
+    "i": "^0.3.7",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
     "jest-silent-reporter": "^0.5.0",
     "lint-staged": "^9.2.1",
     "loader-utils": "^2.0.4",
     "msw": "^0.42.1",
+    "npm": "^10.2.5",
```
six7, line 202: i don't think this is intended?

package.json
```diff
@@ -192,12 +192,14 @@
     "html-inline-script-webpack-plugin": "^3.2.0",
     "html-webpack-plugin": "^5.0.0",
     "husky": "^3.0.2",
+    "i": "^0.3.7",
```
six7, line 195: i don't think this is intended?

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -40,39 +50,24 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
...
+              const variableId = node.boundVariables?.fills?.[0].id;
+              if (variableId) {
+                const variable = figma.variables.getVariableById(variableId);
+                if (variable) {
+                  const newValue = variableName.replace(oldName, newName);
+                  const resolvedValue = getTokenValue(newValue, resolvedTokens ?? []);
+                  const paint = figma.util.solidPaint(resolvedValue?.value as string);
+                  const fillsCopy = clone(node.fills);
+                  fillsCopy[0] = figma.variables.setBoundVariableForPaint(paint, 'color', variable);
+                  const newLocalVariable = getLocalVariableByName(resolvedValue?.name as string);
```
six7, line 66: This will only work for local variables. In most cases, users don't have variables in the same file. Can you refactor this so it's fetching the variable by key (i think https://www.figma.com/plugin-docs/api/properties/figma-variables-importvariablebykeyasync/ should do the trick, we're using it elsewhere)

six7 (changes requested):
can you also make sure to revert the `SharedPluginDataNamespaces.STYLES` and `SharedPluginDataNamespaces.VARIABLES` code? we dont need that now that you've reverted it

Also, a test case would be great.

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -11,15 +11,15 @@ import { ProgressTracker } from '../ProgressTracker';
...
 const getTokenValue = (name: string, resolvedTokens: AnyTokenList) => (
   resolvedTokens.find((token) => token.name === name)
 );
 
-const getLocalVariableByName = (name: string) => {
-  const localVariables = figma.variables.getLocalVariables();
-  return localVariables.find((variable) => variable.resolvedType === 'COLOR' && variable.name.split('/').join('.') === name);
-};
+const getStyleByName = (name: string) => (
+  figma.getLocalPaintStyles().find((style) => style.name === name)
```
six7, line 21: same for styles, users will not have those styles local, they will be remote. we need to fetch those from the remote.

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -50,7 +50,7 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
           }
         });
 
-        if (Object.keys(tokens).length > 0) {
+        if (Object.keys(tokens).length === 0) {
```
six7, line 53: is this on purpose? this would mean this would never run

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -63,13 +63,20 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
                   const paint = figma.util.solidPaint(resolvedValue?.value as string);
                   const fillsCopy = clone(node.fills);
                   fillsCopy[0] = figma.variables.setBoundVariableForPaint(paint, 'color', variable);
-                  const newLocalVariable = getLocalVariableByName(resolvedValue?.name as string);
-                  fillsCopy[0].boundVariables.color.id = newLocalVariable?.id;
+                  const newVariable = await figma.variables.importVariableByKeyAsync(resolvedValue?.name as string);
```
six7, line 66: have you tried if it's faster if you store these in a cache? i assume if a user has 100 variables applied and remaps them, with this current approach we call `importVariableByKeyAsync` 100 times. if we had it in store we would call it 1 time.

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -63,13 +63,20 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
...
+                  fillsCopy[0].boundVariables.color.id = newVariable.id;
                   node.fills = fillsCopy;
                 }
               }
             }
           }
+          if (getAppliedStylesFromNode(node).length > 0) {
+            const { name: styleName } = getAppliedStylesFromNode(node)[0];
+            if (node.type !== 'DOCUMENT' && node.type !== 'PAGE' && 'fills' in node) {
+              const newStyle = getStyleByName(styleName.replace(oldName, newName).split('.').join('/'));
```
six7, line 76: as mentioned above, this currently will just work for local styles. users will likely almost never have these locally.

six7 (changes requested):

src/constants/SharedPluginDataNamespaces.ts
```diff
@@ -1,3 +1,5 @@
 export enum SharedPluginDataNamespaces {
   TOKENS = 'tokens',
+  VARIABLES = 'variables',
+  STYLES = 'styles',
```
six7, line 4: not needed anymore

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -29,17 +38,88 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
...
+            if (value.includes(oldName)) {
+              const newValue = value.replace(oldName, newName);
+              const jsonValue = JSON.stringify(newValue);
+              node.setSharedPluginData(namespace, key, jsonValue);
+            }
+          });
+
+          if (Object.keys(tokens).length === 0) {
+            if (getAppliedVariablesFromNode(node).length > 0) {
+              const { name: variableName } = getAppliedVariablesFromNode(node)[0];
```
six7, line 53: does this mean we only pick the first variable from all applied properties? what if there's a color, a width, a height, a spacing variable attached?

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -29,17 +38,88 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
...
+                    node.fills = fillsCopy;
+                  }
+                }
+              }
+            }
+            if (getAppliedStylesFromNode(node).length > 0) {
+              const effectStyles = figma.getLocalEffectStyles();
+              const paintStyles = figma.getLocalPaintStyles();
+              const textStyles = figma.getLocalTextStyles();
+              const allStyles = [...effectStyles, ...paintStyles, ...textStyles];
```
six7, line 81: this is something that happens on every token, and on every node now. move this code outside of the main function to just perform it once and then remember those

src/plugin/asyncMessageHandlers/bulkRemapTokens.ts
```diff
@@ -29,17 +38,88 @@ export const bulkRemapTokens: AsyncMessageChannelHandlers[AsyncMessageTypes.BULK
...
+
+              const themeInfo = await AsyncMessageChannel.PluginInstance.message({
```
six7, line 83: again, this happens on every node and on every token, this has got to be costly for performance. move it outside the main function and just trigger it once

six7 (changes requested):

PR #2418, opened 2023-12-09
2353-unable-to-un-apply-tokens-on-multiple-layers-at-once

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/5278d16c-b8ff-48d6-a92a-d937004a9cb3

### DIFF
--- a/src/plugin/updatePluginDataAndNodes.test.ts
+++ b/src/plugin/updatePluginDataAndNodes.test.ts
@@ -36,11 +36,6 @@ describe('updatePluginDataAndNodes', () => {
       entries, values, tokensMap, figmaStyleMaps, figmaVariableReferences: new Map(), figmaStyleReferences: {},
     });
 
-    console.log('entries: ', entries);
-    console.log('values: ', values);
-    console.log('tokensMap: ', tokensMap);
-    console.log('figmaStyleMaps: ', figmaStyleMaps);
-
     expect(mockSetSharedPluginData).toBeCalledWith('tokens', 'borderRadius', 'none');
     expect(mockSetSharedPluginData).toBeCalledWith('tokens', 'fill', '"red"');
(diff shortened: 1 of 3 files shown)


src/plugin/updatePluginDataAndNodes.test.ts
```diff
@@ -36,6 +36,11 @@ describe('updatePluginDataAndNodes', () => {
       entries, values, tokensMap, figmaStyleMaps, figmaVariableReferences: new Map(), figmaStyleReferences: {},
     });
 
+    console.log('entries: ', entries);
+    console.log('values: ', values);
+    console.log('tokensMap: ', tokensMap);
+    console.log('figmaStyleMaps: ', figmaStyleMaps);
```
six7, line 42: logs

src/plugin/updatePluginDataAndNodes.ts
```diff
@@ -44,7 +43,6 @@ export async function updatePluginDataAndNodes({
           const jsonValue = JSON.stringify(value);
           switch (value) {
             case 'delete':
-              delete tokenValues[key as CompositionTokenProperty];
```
six7, line 47: can you explain why this was removed?

six7 (changes requested):

src/plugin/updatePluginDataAndNodes.ts
six7: or why that particular change caused the bug to go away?

robinhoodie0823: In case of multiple layer or nodes, it needs user action needs to be applied to all of them.
But since the user action would be removed here, it could only applied to one node.

six7 (approved):

esthercheran (commented):
Feature is working as expected.

PR #2424, opened 2023-12-12
[139]: Replace buttons

six7:
Replaces all occurences of legacy buttons with shared ones


LukeFinch (approved):

packages/tokens-studio-for-figma/src/stories/ThemeStyleManagementCategory.stories.mdx
```diff
@@ -1,6 +1,5 @@
 import { Meta, Story, ArgsTable, Canvas, Preview } from '@storybook/addon-docs';
 import { ThemeStyleManagementCategory } from '@/app/components/ManageThemesModal/ThemeStyleManagementCategory';
-import ColorDisplayBubble from '@/app/components/ColorDisplayBubble';
```
SorsOps, line 3: Why was this removed?

six7: It wasn't in use anymore. We removed that a while ago but seems like we never removed the code.

PR #2425, opened 2023-12-13
Unusual remapping renaming behaviour

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/423cae0d-3718-4f53-8c84-2ae6ba0878ae

### DIFF
--- a/src/app/components/TokenGroup/TokenGroupHeading.tsx
+++ b/src/app/components/TokenGroup/TokenGroupHeading.tsx
@@ -49,8 +49,10 @@ export function TokenGroupHeading({
 
   const handleRenameTokenGroupSubmit = React.useCallback(async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
-    renameGroup(path, newTokenGroupName, type);
-    remapTokensInGroup({ oldGroupName: `${path}.`, newGroupName: `${newTokenGroupName}.`, type });
+    const tokensToRename = await renameGroup(path, newTokenGroupName, type);
+    remapTokensInGroup({
+      oldGroupName: `${path}.`, newGroupName: `${newTokenGroupName}.`, type, tokensToRename,
+    });
     setShowRenameTokenGroupModal(false);
   }, [newTokenGroupName, path, renameGroup, type, remapTokensInGroup]);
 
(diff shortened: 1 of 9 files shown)


src/app/components/TokenGroup/TokenGroupHeading.tsx
```diff
@@ -49,8 +49,10 @@ export function TokenGroupHeading({
 
   const handleRenameTokenGroupSubmit = React.useCallback(async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
-    renameGroup(path, newTokenGroupName, type);
-    remapTokensInGroup({ oldGroupName: `${path}.`, newGroupName: `${newTokenGroupName}.`, type });
+    const tokensToRename = await renameGroup(path, newTokenGroupName, type);
+    remapTokensInGroup({
```
SorsOps, line 53: This should also have an await in front of it in case someone adds on to this later and wonders why its no executing as expected

src/app/store/useTokens.tsx
```diff
@@ -203,7 +203,9 @@ export default function useTokens() {
     }));
   }, [settings.updateMode]);
 
-  const remapTokensInGroup = useCallback(async ({ oldGroupName, newGroupName, type }: { oldGroupName: string, newGroupName: string, type: string }) => {
+  const remapTokensInGroup = useCallback(async ({
+    oldGroupName, newGroupName, type, tokensToRename,
+  }: { oldGroupName: string, newGroupName: string, type: string, tokensToRename: { oldName: string, newName: string }[] }) => {
```
SorsOps, line 208: Can we make these an actual type instead of inlining in case something else needs it ?

src/app/store/useTokens.tsx
```diff
@@ -227,7 +229,9 @@ export default function useTokens() {
     });
     if (confirmData && confirmData.result) {
       if (Array.isArray(confirmData.data) && confirmData.data.some((data: string) => [UpdateMode.DOCUMENT, UpdateMode.PAGE, UpdateMode.SELECTION].includes(data as UpdateMode))) {
-        await handleBulkRemap(newGroupName, oldGroupName, confirmData.data[0]);
+        tokensToRename.forEach(async (tokenToRename) => {
```
SorsOps, line 232: This could cause an issue with race conditions if the bulk remap is slow and something else attempts to change data while its transacting.

Lets change this to 

```
  await Promise.all( tokensToRename.map(tokenToRename) =>  handleBulkRemap(tokenToRename.newName, tokenToRename.oldName, confirmData.data[0]));
)
```

SorsOps (changes requested):

keeganedwin:
Tested and confirming that this is fixed

SorsOps (approved):

PR #2430, opened 2023-12-15
[2]: Replace context menu / dropdown menu

six7:
Replaces all legacy occurences of context menu and dropdown menus in favor of the shared ui one.


packages/tokens-studio-for-figma/src/app/components/ApplySelector.tsx
```diff
@@ -53,64 +45,66 @@ export default function ApplySelector() {
         {updateMode}
       </Button>
       <DropdownMenu>
-        <DropdownMenuTrigger
+        <DropdownMenu.Trigger
           css={{
+            borderTopRightRadius: '$medium',
+            borderBottomRightRadius: '$medium',
```
roppazvan, line 51: do we still need the custom css here?

roppazvan (approved):
lgtm

packages/tokens-studio-for-figma/src/app/components/ApplySelector.tsx
six7: yeah, unfortunately. but we will soon replace this whole component with a `SplitButton` where the left side is a button and the right side is a dropdown icon (similar to the `Merge pull request` button in GitHub)

PR #2438, opened 2023-12-26
2328 plugin crash on using the wrong reference in typography tokens

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/851bd25d-4784-4abe-9473-836c1b3dfa87

### DIFF
--- a/src/utils/TokenResolver.test.ts
+++ b/src/utils/TokenResolver.test.ts
@@ -59,6 +59,13 @@ const resolvedTypographyToken = {
   },
 };
 
+const resolvedTypographyFontFamilyToken = {
+  type: TokenTypes.TYPOGRAPHY,
+  value: {
+    fontFamily: 'Arial',
+  },
+};
+
 const unResolvedTypographyToken = {
   type: TokenTypes.TYPOGRAPHY,
   value: {
@@ -287,6 +294,15 @@ const tokens = [
     name: 'pure-numeric',
     value: '0',
     type: TokenTypes.DIMENSION,
+  },
+  {
+    name: 'typography.font-family.resolved',
+    value: { ...resolvedTypographyFontFamilyToken },
+    type: TokenTypes.TYPOGRAPHY,
+  },
+  {
+    name: 'typography.font-family.resolved.reference',
+    value: '{typography.font-family.resolved}',
   },
 ];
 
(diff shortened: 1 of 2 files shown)


src/utils/TokenResolver.ts
```diff
@@ -285,7 +285,12 @@ class TokenResolver {
           if (resolvedValue.failedToResolve) {
             failedToResolve = true;
           }
+          if (typeof resolvedValue.value === 'object' && key in resolvedValue.value) {
+            failedToResolve = true;
+          }
+          //  else {
```
SorsOps, line 291: Remove

src/utils/TokenResolver.ts
```diff
@@ -285,7 +285,12 @@ class TokenResolver {
           if (resolvedValue.failedToResolve) {
             failedToResolve = true;
           }
+          if (typeof resolvedValue.value === 'object' && key in resolvedValue.value) {
```
SorsOps, line 288: If I'm reading this correctly this is checking to see if a resolved value is itself an object  which contains the key we are attempting to read. Is that really a failure state , would we not rather do  resolvedObject[key] = resolvedValue.value[key]; if this is valid?

src/utils/TokenResolver.test.ts
```diff
@@ -643,6 +659,32 @@ const output = [
...
   },
+  {
+    name: 'typography.font-family.resolved',
+    rawValue: {
+      type: 'typography',
+      value: {
+        fontFamily: 'Arial',
+      },
+    },
+    type: 'typography',
```
SorsOps, line 670: This doesn't look right,  we shouldn't be expecting a nested typography token in another one

esthercheran (commented):
Functionally working as expected

src/utils/TokenResolver.ts
robinhoodie0823: Yes. This is the what token looks like if it has wrong reference.

src/utils/TokenResolver.test.ts
robinhoodie0823: This is the case of wrong reference in typography token.

SorsOps:
@six7 I think I might have to defer to you here. Not sure I have the context on this partial resolution

src/utils/TokenResolver.test.ts
six7: So I think the problem is that technically this resolves correctly, as the reference was found. It's just that that token is not a proper reference for this kind of token.

We should establish something like rules for what a token reference can be, such as a fontFamily token should only be able to refer to tokens that are of type `string` or `fontFamily`. or, technically, anything that isn't an object.

we do not have this logic yet.. and if we find this code 6 months from now we will have no idea why this code is there.

i'd say let's add code comments here to clarify why this is done the way it's done.

then we should do a refactor of the resolution to add something where we can - for each token type - schematically detect what a "valid" reference is, and treat it not as broken, but as "invalid". this requires more work than solving the current issue, which is why i'll approve, add code comments, and create a new issue to track this work.

src/utils/TokenResolver.test.ts
```diff
@@ -643,6 +659,32 @@ const output = [
     rawValue: '0',
     type: TokenTypes.DIMENSION,
   },
+  {
```
six7, line 683:
```suggestion
  // Note: This is technically incorrect, but we're doing this because this needs to be shown as failed to resolve.
  {
```

src/utils/TokenResolver.ts
six7, line 300:
```suggestion
          // Note: This is technically incorrect. We should not be expecting this, but we need to mark those tokens that are referencing an object even though they should be expecting a string as invalid, e only have failedToResolve today.
          if (typeof resolvedValue.value === 'object' && key in resolvedValue.value) {
```

six7 (changes requested):

six7 (approved):

PR #2440, opened 2023-12-27
2427-renaming-tokens-in-the-other-type-breaks-references

robinhoodie0823:
https://github.com/tokens-studio/figma-plugin/assets/103296157/f897fb11-9e83-4bab-a376-fd1e56f75bcf


src/app/store/models/tokenState.tsx
```diff
@@ -446,6 +446,41 @@ export const tokenState = createModel<RootModel>()({
         tokens: newTokens,
       };
     },
+    updateOtherAliases: (state, data: string[]) => {
```
six7, line 449: i think all of this is not required.

all it takes to fix the issue is https://github.com/tokens-studio/figma-plugin/blob/main/src/app/store/models/tokenState.tsx#L360 and remove the check on type. this function right now looks at the type of the originating group, and only updates those that have the same. if we remove the type check there, that's all we need.

six7 (changes requested):

src/app/store/models/tokenState.tsx
robinhoodie0823: I think, what you suggest is to change the tokens in a specific group. But we also have to update tokens which have the changed tokens group as reference. Exactly `Other` type tokens.

six7: ah, i see. then we should be able to just add the logic to https://github.com/tokens-studio/figma-plugin/blob/main/src/app/store/models/tokenState.tsx#L409-L442

similar to how `updateAlias` is called from here: https://github.com/tokens-studio/figma-plugin/blob/main/src/app/store/models/tokenState.tsx#L576

src/app/store/models/tokenState.tsx
```diff
@@ -428,6 +428,21 @@ export const tokenState = createModel<RootModel>()({
...
+            if (token.$extensions?.['studio.tokens'] && token.$extensions?.['studio.tokens'].modify && token.$extensions?.['studio.tokens'].modify.value.includes(data.oldName)) {
+              return {
+                ...token,
+                $extensions: {
+                  ...token.$extensions,
+                  'studio.tokens': {
+                    ...token.$extensions['studio.tokens'],
+                    modify: {
+                      ...token.$extensions['studio.tokens'].modify,
+                      value: token.$extensions['studio.tokens'].modify.value.replace(data.oldName, data.newName),
```
six7, line 440: The token could also appear not just in the value field but in the 'amount' field (eg for mix modifier or lighten or darken

Lets add that here as well so this is covered too 👍

src/app/store/models/tokenState.tsx
```diff
@@ -428,6 +428,21 @@ export const tokenState = createModel<RootModel>()({
                 }, {}),
               } as SingleToken;
             }
+            if (token.$extensions?.['studio.tokens'] && token.$extensions?.['studio.tokens'].modify && token.$extensions?.['studio.tokens'].modify.value.includes(data.oldName)) {
```
six7, line 431: Technically if we have a faulty token like

```
studio.tokens: {
  modify: {
    'foo': 'bar'
  }
}
```

Will this crash? We do have the modify field but the value field is not defined. Or do we have a try catch around this?

Maybe add another `?.` just to make sure we dont expect that element to be there

six7 (changes requested):
Just two minor comments 👍

six7 (approved):

PR #2447, opened 2023-12-31
2: Polish settings

six7:
### Why does this PR exist?

Polishes the experience for 2.0 around the Settings area.

Introduce a new ExplainerModal for helpful hints
Moves base rem configuration to a modal
Replaces the DownshiftInput implementation to use a radix dialog - not just in Settings but anywhere used
Introduce modal={false} behavior for the modals to work around a radix popover issue

### Testing this change

Test the Settings screen and any pages using a token picker (edit token form for example, base rem config, etc)

## Screenshots

| | | |
|---|---|---|
| Image 1 | Image 2 | Image 3 |
| Image 4 | Image 5 | Image 6 |
| Image 7 | | |


roppazvan (approved):
Looks good to me, except this. the alias list overflows

Screenshot 2024-01-05 at 12 30 54

PR #2449, opened 2024-01-03
W3C DTCG format

### LINKED ISSUE #999: Support new W3C syntax for `$value`, `$type` and `$description`
The W3C format defines the use of `$` characters at the beginning of all properties such as `$value` and `$description` or `$type`. We currently do not support that. Given a lot of our users do not have this setup either, we need to be careful when we introduce this.

### Let users opt in to this format
In the credentials of each provider, let users define what format they want to use. 
- Legacy Tokens Studio format (`value`, `description`, `type`)
- DTCG Format

If a user has chosen the new format, we output the JSON in a different way. `type`, `description` and `value` properties will be prefixed by the `$` character. This should be a rather simple change, given we transform the JSON anyway when we display it to users.

- Display $ character in front of properties
- When exporting include this new format
- Use these new property names to pick up values instead of the old way

### Why store this on the provider level?
Users might opt-in to this approach on a storage level, as they convert their tokens or their build pipelines. 

### What is the default going forward?
For a migration period, we keep the current format as the default and use the new one as an opt-in format. After a while we'd treat the new format as the default (but any credentials not containing this `Token format` property will be treated like the old one.

six7:
### Why does this PR exist?

Closes #999

Introduces support for the DTCG format as in supporting `$value`, `$type` and `$description` in both input and output.

Introduces a setting that, when changed, writes and reads tokens to the `$-prefixed` props instead of the legacy ones we supported before. We'll still support both, but this will allow the new format to be chosen.

This is part of a larger epic where in the end we'll allow users to choose the format (possibly on the provider). Another thought is that we could just automatically detect the format and then use that, but then offer users a one-time "migrate to new" button where we'd convert their tokens.


packages/tokens-studio-for-figma/src/app/components/ConvertToDTCGModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const handleConvert = React.useCallback(() => {
+    dispatch.tokenState.setTokenFormat(TokenFormatOptions.DTCG);
+  }, [dispatch]);
+
+  return (
+    <Modal title="Convert to W3C DTCG format" isOpen={isOpen} close={handleClose} showClose>
+      {tokenFormat === TokenFormatOptions.Legacy ? (
+        <Stack direction="column" align="start" gap={4} css={{ color: '$fgMuted', fontSize: '$xsmall' }}>
+          Your tokens are currently stored in the legacy format. If you want to use the new W3C DTCG format, you can
+          convert them here. Read all about changing the format here, including helpful advice on how to migrate your
```
LukeFinch, line 26: Link is missing?

six7: Ah, yes. That was just a placeholder for an imaginary article 😅

packages/tokens-studio-for-figma/src/app/components/FormatSelector.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const isDTCG = tokenFormat === TokenFormatOptions.DTCG;
+
+  const handleValueChange = React.useCallback((value: TokenFormatOptions) => {
+    dispatch.tokenState.setTokenFormat(value);
+  }, [dispatch.tokenState]);
+
+  return (
+    <Stack align="center" gap={2}>
+      <Text size="xsmall" muted>Format</Text>
+      <ToggleGroup type="single" value={isDTCG ? TokenFormatOptions.DTCG : TokenFormatOptions.Legacy} onValueChange={handleValueChange}>
```
LukeFinch, line 23: We have a small one of these now in the ds if you need it.

packages/tokens-studio-for-figma/src/app/components/StartScreen.tsx
```diff
@@ -56,6 +56,7 @@ function StartScreen() {
   }, [dispatch]);
 
   const onSetDefaultTokens = React.useCallback(() => {
+    console.log('onSetDefaultTokens');
```
LukeFinch, line 59: Did you mean to keep these console logs in?

packages/tokens-studio-for-figma/src/app/components/StorageItem.tsx
```diff
@@ -31,6 +34,16 @@ const StorageItem = ({ item, onEdit }: Props) => {
   const { confirm } = useConfirm();
   const { setStorageType } = useStorage();
   const dispatch = useDispatch<Dispatch>();
+  const tokenFormat = useSelector(tokenFormatSelector);
+  const [isConvertModalOpen, setIsConvertModalOpen] = React.useState(false);
+  const closeConvertModal = React.useCallback(() => {
+    console.log('Closing!');
```
LukeFinch, line 40: Console logs.

LukeFinch (approved):

PR #2460, opened 2024-01-12
Add colorspace option to mix

### LINKED ISSUE #2402: [Bug]: Color modifier mix ignores color space
From Slack:

Color modifier question: I expected that “mix” with white would give the same result as lighten, if using the same amount and color space. But for some reason, lighten gives a better result (hue remains unchanged) while mix with white is sub par. Also, the color space setting doesn’t seem to make any difference when using mix.

https://github.com/tokens-studio/figma-plugin/blob/main/src/utils/color/mix.ts

Seems we're ignoring space for mix.

LukeFinch:
### Why does this PR exist?

Closes #2402 

The mix colormodifer ignored the colorspace option. This PR adds it.

### DIFF
--- a/.changeset/five-poets-trade.md
+++ b/.changeset/five-poets-trade.md
@@ -1,0 +1,5 @@
+---
+'@tokens-studio/figma-plugin': patch
+---
+
+Fixes an issue with the Mix color modifier ignoring colorspace options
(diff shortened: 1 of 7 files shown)


six7 (approved):
Can you add a changeset so it shows up in the release notes? Other than that let's get this merged and into release next week 👍

six7:
Tests are failing btw

PR #2466, opened 2024-01-18
2165 duplicating a token to another set displays error

### LINKED ISSUE #2165: Duplicating a Token to Another Set Displays Error
**Describe the bug**
When duplicating a token to another set, the error `Token names must be unique` is displayed. This only happens when duplicating it from the menu when right-clicking the token itself, and not the token-group.

**To Reproduce**
1. Right-click a token
2. Click duplicate token
3. Remove `copy` from the token-name
4. Unselect current set and select another set 
5. See error

**Expected behavior**
Would it be possible to throw this error _if_ and _only_ when the user has the same/current set selected when duplicating a token?
- _This error maybe shouldn't be displayed when duplicating a token to another set as the purpose of this process is to have sets overwrite tokens_

The current set is `Light`, and even when having only the `Dark` set selected the error is displayed:

CleanShot 2023-08-15 at 11 44 46@2x

robinhoodie0823:
### Why does this PR exist?
When try to duplicate token to another token set with same name, there is a error alert 'Token names must be unique'.

Closes #2165 

Resolved to duplicate token to other token sets with same name.

### Testing this change
Updated `EditTokenForm.tsx` and added a condition whether the `activeTokenSet` is included in `selectedItems` (token sets which user wants to duplicate the token to) list.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/f1af8a8f-d613-4517-8d2e-77a721fb92a5


six7 (approved):

six7 (changes requested):
can you recreate the PR? the rebase seems to have added changes that we dont want. i remember the original PR was quite small

PR #2471, opened 2024-01-19
fix letter spacing supporting rem

six7:
### Why does this PR exist?

Fixes the issue that @SamIam4Hyma raised

`Just curious if there is a reason why letter spacing is the only "dimension" property within our current typography tokens that won't accept rem values that map properly to the Figma UI?`

Adds rem to the list of transform options like we have for lineHeight

### Testing this change

Create a letter spacing token with rem at the end


six7:
I'll still expand this PR

robinhoodie0823 (approved):

PR #2473, opened 2024-01-22
2386 ability to edit tokens to add nested levels

### LINKED ISSUE #2386: Ability To Edit Tokens To Add Nested Levels
**Describe the bug**
Currently when you create a token like: `button.primary.container` and then want to edit it to add an additional level `button.primary.container.default`. The plugin throws the following error message "**Tokens can't share name with a group**".

**To Reproduce**
1. Create a token named `button.primary.container`
2. After creating, edit the token
3. Add a '**.**' to change it to `button.primary.container.default`

**Expected behavior**
Users should just be able to add additional levels if required. The current work around is to create the token from scratch.

CleanShot 2023-11-22 at 16 46 25@2x

CleanShot 2023-11-22 at 16 46 51@2x

robinhoodie0823:
### Why does this PR exist?
User can't edit token to add nested levels.

Closes #2386 

This PR gives ability to users to add nested levels to a token.

### Testing this change
The current implementation has a if condition to check the new token name starts with one of resolved tokens name to check if token share the same name with group name which is a block to add nested levels.
This PR added a new condition here to check whether the user action is `create` or `edit` token.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/f2a98cdd-a053-4487-bbea-ca493ff3c733

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/EditTokenForm.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/EditTokenForm.tsx
@@ -126,6 +126,21 @@ function EditTokenForm({ resolvedTokens }: Props) {
     return false;
   }, [internalEditToken]);
 
+  const hasPriorTokenName = React.useMemo(
+    () => {
+      const tokensWithSameParent = resolvedTokens.filter((t) => t.internal__Parent === activeTokenSet);
+      if (internalEditToken?.status === EditTokenFormStatus.CREATE) {
+        // If we are creating a new token, disallow naming it as a prefix of an existing token
+        return tokensWithSameParent.find((t) => internalEditToken.name?.startsWith(`${t.name}.`));
+      } else if (internalEditToken?.status === EditTokenFormStatus.EDIT) {
+        // If we are editing a token, only disallow the name if it's prefix matches another token and it is not the token we are currently editing
+        return tokensWithSameParent.find((t) => internalEditToken.name?.startsWith(`${t.name}.`) && internalEditToken.initialName !== t.name);
+      }
+      return false;
+    },
+    [internalEditToken, resolvedTokens, activeTokenSet],
+  );
+
   const nameWasChanged = React.useMemo(() => internalEditToken?.initialName !== internalEditToken?.name, [
     internalEditToken,
   ]);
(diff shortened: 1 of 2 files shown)


packages/tokens-studio-for-figma/src/app/components/EditTokenForm.tsx
```diff
@@ -126,13 +126,6 @@ function EditTokenForm({ resolvedTokens }: Props) {
     return false;
   }, [internalEditToken]);
 
-  const hasPriorTokenName = React.useMemo(
```
six7, line 129: this check had a reason to exist, which we should respect:

When a user has an existing token called `colors.blue.500`, and they create a new token, it cannot begin with the name of a token that already exists, e.g. if a user tried to create `colors.blue.500.primary` - this should not be allowed.

The removal of this code means that users would be able to do this.

The issue at hand describes how users need to be able to do this IF they are editing the token that causes the name collision, so in our example above if they were editing `colors.blue.500` to be renamed to `colors.blue.500.primary`

six7: Copilot suggestion:

```
const hasPriorTokenName = React.useMemo(
  () => {
    const tokensWithSameParent = resolvedTokens.filter((t) => t.internal__Parent === activeTokenSet);
    if (internalEditToken?.status === EditTokenFormStatus.CREATE) {
      // If we are creating a new token, disallow naming it as a prefix of an existing token
      return tokensWithSameParent.find((t) => internalEditToken.name?.startsWith(`${t.name}.`));
    } else if (internalEditToken?.status === EditTokenFormStatus.EDIT) {
      // If we are editing a token, only disallow the name if it's prefix matches another token and it is not the token we are currently editing
      return tokensWithSameParent.find((t) => internalEditToken.name?.startsWith(`${t.name}.`) && internalEditToken.initialName !== t.name);
    }
    return false;
  },
  [internalEditToken, resolvedTokens, activeTokenSet],
);
```

six7 (changes requested):

six7 (approved):
Changes look good from a technical side

esthercheran (commented):
Functionality is working as expected

PR #2511, opened 2024-02-23
feat: added regex support to bulk remap

xsu1010:
### Why does this PR exist?

The Bulk Remap feature is incredibly useful, as it saves a ton of time and work when renaming multiple tokens. 
By adding [regex support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions) to it, one can provide more precise patterns to the token search, allowing them to bulk remap tokens even faster.

- `bulkRemapTokens.ts`
  - Added support for regex
  - Added test for regex search

### Testing this change

1. Select multiple layers.
2. Use the bulk remap feature and enter a regular expression that matches a portion of the tokens in those layers.
3. Enter a value to change that portion to. Note how only the portion that matches the regex will change.

https://github.com/tokens-studio/figma-plugin/assets/25635678/cf2bb7a2-7f0c-4372-9147-8fe4701f6be3

### Additional Notes (if any)

- Although I changed how the code tests the token for `pattern` instead of `oldName`, this does not affect the current way of searching for a normal string, as normal strings are also valid regex.
- When creating the `pattern` const, it checks to see if the string passed is already a regex with slashes, and if so, ~~it strips the first and last slashes — if the slashes remained, then the matches wouldn't be the same anymore, as it would actually search for the slashes.~~ it uses that regex pattern with the given flags, if any. 
  - However, I also made it so that the user doesn't need to pass slashes to use a regex — if there are no slashes, it just uses that pattern with a global flag (matches all occurrences of that pattern in the token), as you can see in the video above.
  - When doing this, I thought that maybe you'd prefer to enforce the usage the slashes if the user is trying to include regex in the search. Which one would you prefer? 😅

### DIFF
--- a/.changeset/honest-clouds-add.md
+++ b/.changeset/honest-clouds-add.md
@@ -1,0 +1,5 @@
+---
+'@tokens-studio/figma-plugin': minor
+---
+
+Added regex support to bulk remap
--- a/src/plugin/asyncMessageHandlers/__tests__/bulkRemapTokens.test.ts
+++ b/src/plugin/asyncMessageHandlers/__tests__/bulkRemapTokens.test.ts
@@ -95,7 +95,7 @@ describe('bulkRemapTokens', () => {
     ]));
     await bulkRemapTokens({
       type: AsyncMessageTypes.BULK_REMAP_TOKENS,
-      oldName: 'old-size.*',
+      oldName: '/old-size.*/g',
       newName: 'new-size.1000',
       updateMode: UpdateMode.SELECTION,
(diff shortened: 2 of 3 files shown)


six7:
Thank you for this! We're currently working on a dedicated branch for the upcoming 2.0 release which I'll rebase this PR on, which shouldn't lead to any conflicts given we've not changed much in this area of the plugin.

six7:
Closing as we merged the other rebased PR 👍  Thank you!!

PR #2527, opened 2024-03-04
Fix parsing Supernova new app URL

tchief:
### Why does this PR exist?

Supernova released a new app with a different URL. Customers pasting new Supernova URL are not able to sync tokens.

Adds support for parsing new Supernova URL.


six7 (approved):
@esthercheran should be good to take forward

PR #2533, opened 2024-03-12
[749] Add support for CRUD operation for Tokens Studio

georgebuciuman:
### Why does this PR exist?

Closes https://github.com/tokens-studio/studio-app/issues/749

This is a PR to add support for basic CRUD operations for Tokens Studio source

  - It adds: create token, update token, delete token
  - added a key in tokens state to store the token set urns (`tokenSetMetadata`) where we could add other metadata in the future
  - when we pull tokens from Tokens Studio, for each token we also store the token urn in `$extensions.id`

### Testing this change
Add Tokens Studio provider, pull tokens and then start making changes (create, edit, delete token)


packages/tokens-studio-for-figma/src/app/store/models/tokenState.tsx
```diff
@@ -273,27 +281,27 @@ export const tokenState = createModel<RootModel>()({
...
               ...updateTokenPayloadToSingleToken({
-                parent: data.parent,
-                name: data.newName,
-                type: data.type,
-                value: data.value,
-                description: data.description,
-                oldName: data.oldName,
-                $extensions: data.$extensions,
-              } as UpdateTokenPayload, uuidv4()),
+                  parent: data.parent,
```
six7, line 284: Alignment feels off (i think we have 2 spaces here)

packages/tokens-studio-for-figma/src/app/store/models/tokenState.tsx
```diff
@@ -486,58 +494,58 @@ export const tokenState = createModel<RootModel>()({
     updateAliases: (state, data: TokenToRename) => {
       const newTokens = Object.entries(state.tokens).reduce<TokenState['tokens']>(
         (acc, [key, values]) => {
-          const newValues = values.map<SingleToken>((token) => {
-            if (Array.isArray(token.value)) {
-              return {
-                ...token,
-                value: token.value.map((t) => Object.entries(t).reduce<Record<string, string | number>>((a, [k, v]) => {
+        const newValues = values.map<SingleToken>((token) => {
```
six7, line 497: same, alignment issues

six7: note: those make it a bit harder to review and feel like they're not solving a problem but causing one

packages/tokens-studio-for-figma/src/app/store/models/tokenState.tsx
```diff
@@ -708,8 +732,25 @@ export const tokenState = createModel<RootModel>()({
     duplicateToken() {
       dispatch.tokenState.updateDocument({ shouldUpdateNodes: false });
     },
-    createToken() {
+    async createToken(payload: UpdateTokenPayload, rootState) {
       dispatch.tokenState.updateDocument({ shouldUpdateNodes: false });
+
+      if (rootState.uiState.api?.provider === StorageProviderType.TOKENS_STUDIO) {
+        const token = await pushToTokensStudio({
```
six7, line 739: i wonder if we can move all of this into a custom tokens studio hook so that we dont clutter up this file and rather just call a 1-liner of await createTokenInTokensStudio(...)?

six7: could clean up this file and keep it clear of any actual implementation (at least partially)

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -31,6 +49,9 @@ const tsToToken = (raw: RawToken) => {
   const combined: Token = {
     type: raw.type,
     value: null,
+    $extensions: {
+      id: raw.urn!,
```
six7, line 53: can we move this to the 'studio.tokens' namespace? the w3c recommends to keep extensions items inside namespaces and not in the $extensions property directly, so that it becomes

```
$extensions: {
  'studio.tokens': {
    id: ...
```

not sure if this has implications on other areas?

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -201,7 +187,117 @@ export class TokensStudioTokenStorage extends RemoteTokenStorage<TokensStudioSav
...
+                extensions: JSON.stringify(data.$extensions),
+              },
+            }),
+          );
+
+          if (!responseData.data) {
+            return null;
+          }
+
+          notifyToUI('Token pushed to Tokens Studio', { error: false });
```
six7, line 229: Can we add some metric tracking here? the `track()` function sends an event to mixpanel, maybe one for each of those actions? So that we are aware of how many people use this?

six7 (commented):
Looks really good overall, love it! Some comments and I noticed some whitespace issues (not sure if those are actual issues or if they just appear wrong for me)

six7 (commented):
Looks really good overall, love it! Some comments and I noticed some whitespace issues (not sure if those are actual issues or if they just appear wrong for me)

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
georgebuciuman: As discussed, we will keep this in here for now and will move it to a new property `urn` on `studio.tokens` namespace in the future. Id property gets removed from the namespace and we are not able to use it

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -226,6 +227,7 @@ export class TokensStudioTokenStorage extends RemoteTokenStorage<TokensStudioSav
             return null;
           }
 
+          track('Create token in Tokens Studio', { token: data });
```
six7, line 230: Is `data` an object? i think we can remove the second argument and just track the event without any extra data 🤔

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -260,6 +262,7 @@ export class TokensStudioTokenStorage extends RemoteTokenStorage<TokensStudioSav
             return null;
           }
 
+          track('Edit token in Tokens Studio', { token: data });
```
six7, line 265: same here, lets just do the event itself

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -286,6 +289,7 @@ export class TokensStudioTokenStorage extends RemoteTokenStorage<TokensStudioSav
             return null;
           }
 
+          track('Delete token from Tokens Studio', { token: data });
```
six7, line 292: same here, lets just do the event itself

six7 (approved):
just minor comments for the mixpanel stuff, after that let's get it merged 🎉

PR #2536, opened 2024-03-13
Fix focus in morebutton  context menu

LukeFinch:
### Why does this PR exist?

Closes #0000


six7 (approved):
🎉

we can remove the changeset though :D (if users have never seen the old state they wont understand this had a change)

PR #2543, opened 2024-03-19
[792] Tokens Studio - CUD for token sets and duplicate token

georgebuciuman:
### Why does this PR exist?

Closes https://github.com/tokens-studio/studio-app/issues/792

Added support for updating data in Tokens Studio from the plugin:
- duplicate token in multiple token sets
- create token set
- rename token set
- delete token set

### Testing this change

https://github.com/tokens-studio/figma-plugin/assets/19837930/24b37553-f547-4619-ae6d-204a8e904e8c

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/store/providers/tokens-studio/tokensStudio.tsx
+++ b/packages/tokens-studio-for-figma/src/app/store/providers/tokens-studio/tokensStudio.tsx
@@ -24,22 +24,30 @@ import { RemoteTokenStorageMetadata } from '@/storage/RemoteTokenStorage';
 type TokensStudioCredentials = Extract<StorageTypeCredentials, { provider: StorageProviderType.TOKENS_STUDIO }>;
 type TokensStudioFormValues = Extract<StorageTypeFormValues<false>, { provider: StorageProviderType.TOKENS_STUDIO }>;
 
-export type TokensStudioAction = 'CREATE_TOKEN' | 'EDIT_TOKEN' | 'DELETE_TOKEN';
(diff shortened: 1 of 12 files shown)


packages/tokens-studio-for-figma/src/app/store/providers/tokens-studio/tokensStudio.tsx
```diff
@@ -24,22 +24,30 @@ import { RemoteTokenStorageMetadata } from '@/storage/RemoteTokenStorage';
...
+export type TokensStudioAction =
+  | 'CREATE_TOKEN'
+  | 'EDIT_TOKEN'
+  | 'DELETE_TOKEN'
+  | 'CREATE_TOKEN_SET'
+  | 'UPDATE_TOKEN_SET'
+  | 'DELETE_TOKEN_SET';
 
 interface PushToTokensStudio {
   context: TokensStudioCredentials;
```
six7, line 36: shouldnt `projectUrn` be part of `context` already? its `context.secret` as far as i know? (the urn that users add when setting up the sync

six7 (approved):
Looks good overall! Left just a minor comment, i might be wrong - if i am, feel free to merge in

packages/tokens-studio-for-figma/src/app/store/providers/tokens-studio/tokensStudio.tsx
georgebuciuman: You're absolutely right. I completely forgot about having it in the context. It's in the `context.id` and I'm using that one now

PR #2546, opened 2024-03-19
fix release 139 issues

six7:
### Why does this PR exist?

Fixes several 2.0 branch related issues

Fixes DTCG not being available for local storage sync
Fixes dropdown for importing styles and variables being missing
Fixes create themes button not doing anything (removed it, feels like we're better off without it here)
Adjusts apply selector size

### Testing this change

should be all good. famous last words.

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/Inspector.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/Inspector.tsx
@@ -36,8 +36,10 @@ function Inspector() {
   ), [tokens, usedTokenSet]);
 
   const handleSetInspectView = React.useCallback((view: 'multi' | 'debug') => {
-    track('setInspectView', { view });
-    setInspectView(view);
+    if (view) {
+      track('setInspectView', { view });
+      setInspectView(view);
+    }
   }, []);
 
   function renderInspectView() {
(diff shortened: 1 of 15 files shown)


packages/tokens-studio-for-figma/src/app/components/Inspector.tsx
```diff
@@ -36,8 +36,10 @@ function Inspector() {
   ), [tokens, usedTokenSet]);
 
   const handleSetInspectView = React.useCallback((view: 'multi' | 'debug') => {
-    track('setInspectView', { view });
-    setInspectView(view);
+    if (view) {
```
six7, line 39: this is needed in radix to prevent an empty toggle switch

packages/tokens-studio-for-figma/src/app/components/SyncSettings.tsx
```diff
@@ -132,18 +132,16 @@ const SyncSettings = () => {
...
-                <StorageItem
-                  key={item?.internalId || `${item.provider}-${item.id}`}
-                  onEdit={handleEditClick(item)}
-                  item={item}
-                />
-              ))}
-            </Stack>
-          )}
+          <Stack direction="column" gap={2} width="full" align="start">
+            <LocalStorageItem />
```
six7, line 136: change to always show local storage item, no matter if sync providers exist

LukeFinch (approved):

PR #2560, opened 2024-03-28
2411 allow users to select whether to create styles or variables or both while exporting to figma

### LINKED ISSUE #2411: Allow users to select whether to create Styles or Variables or both while exporting to Figma
In #2406 we introduced a new option `Export to Figma` under the `Styles & Variables` 

Image

We want to allow users to choose whether they want to create variables or styles or both from the `Options` button

Image

- On a new Figma file the `Options` screen should open up first. 
- On subsequent launch, clicking on the `Options` button will open up the below screen

Image

- The options under `What variables should be created and updated?`
  - Select all : Creates color, number/dimension, text/string, booelan tokens as variables 
  - Color : only creates color tokens as variables
  - Number/Dimension: only creates dimension, number, spacing, sizing, border radius, border width as dimension variables
  - Text/String: only creates text tokens as text/string variables
  - Boolean : only creates boolean tokens as boolean variables
- The options under `What styles should be created and updated?`
  - Select all : Creates color, typography and  shadow tokens as color, typography and effects styles
  - Color: creates only color tokens as styles
  - Typography: creates only typography tokens as styles
  - Effects : creates only shadow tokens as styles

- On clicking confirm the screen below will be shown and the settings will be remembered for the FIgma file

Image

- On clicking `Export to Figma`, variables or styles or both will be created depending on the selected checkboxes in the `Options` screen

robinhoodie0823:
### Why does this PR exist?
The current implementation only create styles based on selected sets.

Closes #2411 

Provide a new feature that allow users to create styles based on selected themes.

### DIFF
--- a/.changeset/tough-guests-tease.md
+++ b/.changeset/tough-guests-tease.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Allow users to create styles together when creating variables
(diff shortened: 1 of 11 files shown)


esthercheran (approved):
Functionality is working

LukeFinch (approved):

PR #2561, opened 2024-03-29
Feat/819 add themes and tokensetorder

georgebuciuman:
### Why does this PR exist?

This is a partial implementation for https://github.com/tokens-studio/studio-app/issues/819

This PR adds support for:
- retrieving themes from Tokens Studio to the plugin
- retrieving token set order from Tokens Studio to the plugin and update Tokens Studio when updates in the plugin are happening

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/MoreButton/MoreButton.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/MoreButton/MoreButton.tsx
@@ -80,7 +80,7 @@ export const MoreButton: React.FC<React.PropsWithChildren<React.PropsWithChildre
       parent: activeTokenSet,
       path: token.name,
       type: token.type,
-      sourceId: token.$extensions?.id,
+      sourceId: token.$extensions?.['studio.tokens']?.urn,
     });
   }, [activeTokenSet, deleteSingleToken, token.$extensions, token.name, token.type]);
 
--- a/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
@@ -4,13 +4,16 @@ import {
   CreateTokenMutation,
   UpdateTokenMutation,
   DeleteTokenMutation,
-  TokenSetsQuery,
   RawToken,
   Raw_Token_border,
   Raw_Token_boxShadow,
   CreateTokenSetMutation,
   UpdateTokenSetMutation,
   DeleteTokenSetMutation,
+  ProjectQuery,
+  UpdateTokenSetOrderMutation,
+  ThemeGroup,
+  TokenSet,
 } from '@tokens-studio/sdk';
 import { deepmerge } from 'deepmerge-ts';
 import * as Sentry from '@sentry/react';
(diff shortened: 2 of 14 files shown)


packages/tokens-studio-for-figma/src/app/components/MoreButton/MoreButton.tsx
```diff
@@ -80,7 +80,7 @@ export const MoreButton: React.FC<React.PropsWithChildren<React.PropsWithChildre
       parent: activeTokenSet,
       path: token.name,
       type: token.type,
-      sourceId: token.$extensions?.id,
+      sourceId: token.$extensions?.['studio.tokens']?.urn,
```
six7, line 83: i think this is a mistake, we dont have any check if we're in tokens studio provider - for all other providers this shouldnt set it as it did before

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -124,7 +134,43 @@ async function getTokens(urn: string): Promise<TokensData> {
       },
       { tokens: {}, tokenSets: {} },
     );
-    return returnData;
+
+    const tokenSetOrder = tokenSets
+      .sort((a, b) => (+(a.orderIndex || 0) > +(b.orderIndex || 0) ? 1 : -1))
+      .reduce((acc: string[], tokenSet) => {
+        if (!tokenSet.name) return acc;
```
six7, line 141: can there be a case of a token set not having a name?

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -124,7 +134,43 @@ async function getTokens(urn: string): Promise<TokensData> {
...
+          .map((theme) => {
+            const selectedTokenSets = JSON.parse(JSON.parse(theme?.selectedTokenSets || '')) || [];
+
+            return {
+              id: theme?.urn as string,
+              name: theme?.name as string,
+              group,
+              selectedTokenSets,
+              $figmaStyleReferences: JSON.parse(JSON.parse(theme?.figmaStyleReferences || '{}')),
+              $figmaVariableReferences: JSON.parse(JSON.parse(theme?.figmaVariableReferences || '{}')),
```
six7, line 165: i think it could make sense to wrap the whole part here in a try / catch, given json parse could throw an error if we get faulty data back from the api.

six7 (changes requested):
Just a few small comments, overall nice 🎉

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
georgebuciuman: this entire function `getProjectData` is wrapped in try/catch

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
georgebuciuman: No. But BE type for `name` is `string | null | undefined`. It's the same for all the values. I don't know why but that is how BE sets the types

packages/tokens-studio-for-figma/src/app/components/MoreButton/MoreButton.tsx
georgebuciuman: It is used for delete. We need to send the urn to Tokens Studio to be able to delete the token. I could also not send the urn from the button but get it in `tokenState` based on the name and parent. But that would mean one more loop through all the tokens

six7 (approved):

Issue #2250, opened 2023-09-18
Ability to Reference Variables in Styles when Created via the Plugin


keeganedwin:
**Is your feature request related to a problem? Please describe.**
Figma now supports the ability to reference variables in styles. Users would possibly want to be able to do this in the future but we currently don't support this.

six7:
![Image](https://six7-fix-polish-prx.review-lab.github.com/tokens-studio/figma-plugin/assets/4548309/61953952-4c88-4360-ab1c-bc6c11b3ae8f)

six7:
Will be shipped in 2.0

PR #2562, opened 2024-03-29
Allow color styles to contain variables

six7:
### Why does this PR exist?

Closes https://github.com/tokens-studio/figma-plugin/issues/2250

Introduces ability for styles to link to variables, if the reference is a pure reference (just `{tokenname}`).

CleanShot 2024-03-29 at 23 32 12@2x

Applies raw value if the reference does not exist as a variable

CleanShot 2024-03-29 at 23 33 08@2x

Remembers setting if user doesnt want variables in their styles

https://github.com/tokens-studio/figma-plugin/assets/4548309/8e9c2de3-1d86-4b43-94a6-f857f3406f9a

### Testing this change

- Create a token set that has a color token. Create a theme with that, and create variables so the variable is attached
- Create another token set that references that token, for this set create styles.
- Note how the created style will contain a reference to the token


LukeFinch:
I think we need to be more intentional with how this applies to user's exports. I feel like this should be an option - and not just an assumed default. The steps to achieve this result aren't particularly intuitive. 

With the upcoming release, a user can export styles and variables at the same time. I'd expect this result to occur if I selected to export both, with an additional "Use variables in styles" option.

six7:
> With the upcoming release, a user can export styles and variables at the same time. I'd expect this result to occur if I selected to export both, with an additional "Use variables in styles" option.

I like that idea, will include it 👍

LukeFinch (approved):
Awesome!

PR #2567, opened 2024-04-03
Feat/819 dynamic set

georgebuciuman:
### Why does this PR exist?

Closes part of https://github.com/tokens-studio/studio-app/issues/819

I limited edits for dynamic sets using `editProhibitedSelector`. I agree this might not be the best approach. If this is considered to add more problems instead of resolving it then I will find a better approach and create a separate selector just for this case.


packages/tokens-studio-for-figma/src/selectors/editProhibitedSelector.ts
```diff
@@ -1,4 +1,7 @@
 import { createSelector } from 'reselect';
 import { tokenStateSelector } from './tokenStateSelector';
 
-export const editProhibitedSelector = createSelector(tokenStateSelector, (state) => state.editProhibited);
+export const editProhibitedSelector = createSelector(
+  tokenStateSelector,
+  ({ editProhibited, activeTokenSet, tokenSetMetadata }) => editProhibited || tokenSetMetadata[activeTokenSet]?.isDynamic,
```
six7, line 6: does this work? like if i have 2 token sets and i navigate between them, only the dynamic one is disabled?

i feel like we shouldnt mix those two concepts though.. wondering if instead of that we can create a new selector that is purely about the token set, and then when we decide if the token set is editable we take both into account?

georgebuciuman: It worked but with some weird issues. I wasn't sure about this solution either, that's why I also wrote it in the PR description.

I created selectors for this and updated every place where we needed to block edits

six7: Amazing, thanks for the extra work on this!

six7 (approved):

Issue #2619, opened 2024-04-11
[2.0]: Footer missing top border on Settings page (or all pages)


six7:
BOTTOM PLUGIN FOOTER
missing a "top border" or other visual treatment that separates it from the contents of the plugin page. When the plugin is resized it "melts" into the rest of the plugin content including actions and other elements (see small settings page screenshot for example)

Footer seems to not have a border, and we only show one on the Tokens page given it gets a border bottom. Let's change it to have a borderTop on the footer so it has one, always.

![Image](https://github.com/tokens-studio/figma-plugin/assets/4548309/1b478bb0-14df-48bb-8d08-f7cfdee3f622)

PR #2627, opened 2024-04-11
#2619 [2.0] footer missing top border on settings page or all pages

LukeFinch:
### Why does this PR exist?

Closes #2629 

Adds a border to the footer :)
image


six7 (changes requested):
This will likely cause a double border when on the `Tokens` page, which already had a border?

LukeFinch:
> This will likely cause a double border when on the `Tokens` page, which already had a border?

Nice catch - I fixed it :)

six7 (approved):
🎉

PR #2633, opened 2024-04-12
feat: [819] Add support for themes and groups operations

georgebuciuman:
### Why does this PR exist?

Closes https://github.com/tokens-studio/studio-app/issues/819

Adds support for operations on themes and groups:
- create theme group and theme
- update theme groups and themes
- delete theme groups and themes
- add variable ids and styles ids to themes
- update variable ids and styles ids for themes

### DIFF
--- a/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
@@ -313,7 +313,6 @@ export class TokensStudioTokenStorage extends RemoteTokenStorage<TokensStudioSav
       }
       case 'EDIT_TOKEN': {
         const tokenId = data.$extensions?.['studio.tokens']?.urn;
-        console.log('EDIT_TOKEN', data);
 
         if (!tokenId) {
           throw new Error('Invalid data');
--- a/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
@@ -14,6 +14,9 @@ import {
   UpdateTokenSetOrderMutation,
   ThemeGroup,
   TokenSet,
+  UpdateThemeGroupMutation,
+  CreateThemeGroupMutation,
+  DeleteThemeGroupMutation,
 } from '@tokens-studio/sdk';
 import { deepmerge } from 'deepmerge-ts';
 import * as Sentry from '@sentry/react';
@@ -37,6 +40,9 @@ import {
   UPDATE_TOKEN_SET_MUTATION,
   DELETE_TOKEN_SET_MUTATION,
   UPDATE_TOKEN_SET_ORDER_MUTATION,
+  UPDATE_THEME_GROUP_MUTATION,
+  CREATE_THEME_GROUP_MUTATION,
+  DELETE_THEME_GROUP_MUTATION,
 } from './tokensStudio/graphql';
 import { track } from '@/utils/analytics';
 import { ThemeObjectsList } from '@/types';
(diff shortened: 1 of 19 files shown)


packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -306,6 +313,7 @@ export class TokensStudioTokenStorage extends RemoteTokenStorage<TokensStudioSav
       }
       case 'EDIT_TOKEN': {
         const tokenId = data.$extensions?.['studio.tokens']?.urn;
+        console.log('EDIT_TOKEN', data);
```
six7, line 316: log

packages/tokens-studio-for-figma/src/storage/tokensStudio/updateThemeGroupsInTokensStudio.ts
```diff
@@ -0,0 +1,242 @@
...
+      }
+
+      return theme;
+    });
+
+    shouldUpdateThemes = true;
+  }
+
+  if (themeToCreate) {
+    const createdThemeGroup = await pushToTokensStudio({
```
six7, line 184: should we wrap this in a try / catch? what if our api is down?

packages/tokens-studio-for-figma/src/storage/tokensStudio/updateThemeGroupsInTokensStudio.ts
```diff
@@ -0,0 +1,242 @@
...
+import { ThemeObject, ThemeObjectsList } from '@/types';
+
+interface UpdateThemeGroupsInTokensStudioPayload {
+  prevState: RematchRootState<RootModel, Record<string, never>>;
+  rootState: RematchRootState<RootModel, Record<string, never>>;
+  dispatch: RematchDispatch<RootModel>;
+  action: any;
+}
+
+const getThemeGroupsToUpdate = (themes: ThemeObjectsList, groupIdsMap: Record<string, string>) => {
```
six7, line 15: lets move this out into its own file, this file is already quite huge - i think it could benefit from a bit of separation

packages/tokens-studio-for-figma/src/storage/tokensStudio/updateThemeGroupsInTokensStudio.ts
```diff
@@ -0,0 +1,242 @@
...
+      acc[theme.group] = theme.groupId;
+    }
+    return acc;
+  }, {} as Record<string, string>);
+
+  let themeToCreate: ThemeObject | null = null;
+  let themeGroupsToUpdate: Record<string, ThemeObjectsList> = {};
+  let themeGroupsToDelete: string[] = [];
+
+  switch (action.type) {
```
six7, line 51: can we somehow turn each action type into its own file? that would help with legibility and maintainability instead of having the file itself be quite large

six7 (changes requested):
just that one large file that could use some separation, rest is minor comments 👍

packages/tokens-studio-for-figma/src/storage/tokensStudio/updateThemeGroupsInTokensStudio.ts
georgebuciuman: This is already wrapper in a try/catch for each case inside `TokensStudioTokenStorage`. Do you think another wrapper is needed to tread the error here?

six7: No i guess we're fine then. We'd still get specific error messaging right? As in if the backend throws an error we can show that ? Instead of it being a generic error

georgebuciuman: We get specific errors from BE but I wouldn't show that to the users as we get errors that are not very readable. Right now we only show the errors in the console and don't show the success notification to the users when there is an error. Do you think we should always show the error as a notification to the users?
Errors are something like this: `Error creating theme group in Tokens Studio`, `Error updating token set in Tokens Studio`...

six7: That's fine 👍

six7 (approved):

PR #2670, opened 2024-04-19
2645 20 rename variables non functional 1

### LINKED ISSUE #2645: [2.0]: Rename variables non-functional
Trying to rename a token and checking the `Rename variable` option, does not lead to us renaming that variable

To reproduce:
- Create a token and create variables
- Rename that token in the plugin
- Notice how nothing happens

robinhoodie0823:
### Why does this PR exist?
Current plugin does not update existing variables and styles and does not remove variables and styles without connection.

Closes #2645 

Added sync functions to sync with all updates.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/62e1d675-0d85-4d1f-84a8-44faaa1baa4f


packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
```diff
@@ -324,11 +324,34 @@ export default function useTokens() {
...
   }, [dispatch.tokenState]);
 
+  const syncStyles = useCallback(async () => {
+    track('syncStyles', {
+      renameExistingStylesAndVariables: settings.renameExistingStylesAndVariables,
+      removeStylesAndVariablesWithoutConnection: settings.removeStylesAndVariablesWithoutConnection
+    });
+
+    const syncStyleResult = await wrapTransaction({ name: 'syncStyles' }, async () => AsyncMessageChannel.ReactInstance.message({
+      type: AsyncMessageTypes.SYNC_STYLES,
```
six7, line 334: i just looked into this function. from the code you have provided here it seems like it's running before the function to create styles.

however, if you go look up what SYNC_STYLES actually does, it creates styles as well! https://github.com/tokens-studio/figma-plugin/blob/2645-20-rename-variables-non-functional-1/packages/tokens-studio-for-figma/src/plugin/syncStyles.ts
this is inefficient. we do not want to have two places where we create styles. let's change it so that SYNC_STYLES is really only responsible for deleting and renaming styles that were created.

same for SYNC_VARIABLES: https://github.com/tokens-studio/figma-plugin/blob/2645-20-rename-variables-non-functional-1/packages/tokens-studio-for-figma/src/plugin/syncVariables.ts

we need to be very careful with existing variables. we should follow an order of
- look up existing variables
- rename if needed
- delete those that we will no longer need

AFTER this part
- run create and update variable values function

---

however, currently we seem to be running the create / update part in two entirely different areas of the plugin

six7 (changes requested):
left a comment on when and how to call the syncVariables / syncStyles stuff

packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
robinhoodie0823: If we have to handle `renaming` and `removing` tokens in sync functions, in case of an updated token (with same name), how can we have this case?

packages/tokens-studio-for-figma/src/plugin/syncStyles.ts
```diff
@@ -97,16 +97,20 @@ export default async function syncStyles(tokens: Record<string, AnyTokenList>, o
     if (styleSet[style.name] && !compareStyleValueWithTokenValue(style, styleSet[style.name], settings.baseFontSize)) {
       if (style.type === 'PAINT' && styleSet[style.name].type === TokenTypes.COLOR) {
         setColorValuesOnTarget(style, styleSet[style.name].name);
+        renamedTokenNames.push(styleSet[style.name].name);
```
six7, line 100: it would be better to perform this on line 67 of this file: https://github.com/tokens-studio/figma-plugin/pull/2670/files#diff-11232dee5def4abae12048db6b327184f8a1dad57746bd2eb2df7535d12d164aR67 where you perform the actual rename. makes more sense.

packages/tokens-studio-for-figma/src/plugin/syncVariables.ts
```diff
@@ -26,23 +19,18 @@ export default async function syncVariables(tokens: Record<string, AnyTokenList>
       if (variable) {
         connectedVariablesMap[variableId] = variable;
         if (options.renameVariable && variable.name !== path) {
+          renamedTokenNames.push(variable.name.split('/').join('.'));
```
six7, line 22: is there a reason to push the variable name here? why not push the `tokenName` that you get from the function props further above?

packages/tokens-studio-for-figma/src/plugin/syncVariables.ts
```diff
@@ -26,23 +19,18 @@ export default async function syncVariables(tokens: Record<string, AnyTokenList>
...
           variable.name = path;
```
six7, line 23: so do we perform 2 renames now? i'd say you can remove line 23, 1 rename must be enough?

six7 (changes requested):
left a few comments 👍

packages/tokens-studio-for-figma/src/plugin/syncVariables.ts
robinhoodie0823: The `tokenName` is updated token name which will be replace the original variable name.
The variable name represent the original token name.

packages/tokens-studio-for-figma/src/plugin/syncVariables.ts
robinhoodie0823: Only 1 renaming functionality is here. I think

six7 (approved):

PR #2740, opened 2024-05-08
2723 20 typography styles do not create with variable references if created from sets

### LINKED ISSUE #2723: 2.0: Typography styles do not create with variable references if created from Sets
When you create Typography styles based on Token sets using the Export to Figma dialog using these settings from below.

NOTE: The logic for this likely was only implemented when creating variables from modes, so something in the sets logic might be missing that we have on the other one.

Image

and using Token sets:

Image

https://github.com/tokens-studio/figma-plugin/assets/4548309/104ac906-12fa-489a-9c6f-0cc2c595d40b

None of the styles have variables attached.

robinhoodie0823:
### Why does this PR exist?
Current implementation did not apply typography styles with correct figma variable referenes.

Closes #2723 

Updated to store all figma variable references.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/4110b930-cc84-4902-9f13-ada52bcfbac7


packages/tokens-studio-for-figma/src/plugin/asyncMessageHandlers/getThemeReferences.ts
```diff
@@ -47,9 +47,7 @@ export async function getThemeReferences(prefixStylesWithThemeName?: boolean) {
   const localVariables = await figma.variables.getLocalVariablesAsync();
 
   localVariables.forEach((variable) => {
-    if (!figmaVariableReferences.has(variable.name)) {
-      figmaVariableReferences.set(variable.name, variable.key);
-    }
+    figmaVariableReferences.set(variable.name, variable.key);
```
cuserox, line 50: Would this cause any performance issues, as we're setting **all** `figmaVariableReferences`, whether they exist or not?

Perhaps there's a similar check that can be done to only set those that are necessary, not just comparing the name? For example, comparing the value (i.e. `variable.key`)?

six7 (approved):
👍  thanks for the update on the check

PR #2741, opened 2024-05-08
2722 free users should land in export token set tab when going to styles variables dropdown

### LINKED ISSUE #2722: Free users should land in `Export token set` tab when going to Styles & Variables dropdown
When I go to `Export to Figma` and then confirm the options if I am a free user I should land in the `Token sets` export tab, instead of the Themes (Pro) export tab, so that I can take action right away.

We already have some logic, perhaps some parts of that logic are wrong:

```
<Tabs defaultValue={isProUser ? 'useThemes' : 'useSets'}>
```

To reproduce:
- Remove license key
- Restart plugin
- Go to Export to Figma, confirm the options

You should, by original design, land in the `Token sets` tab, not the Themes (Pro) tab

Relevant thread:

https://hyma-team.slack.com/archives/C06RHB9RT2B/p1714657525482879

robinhoodie0823:
### Why does this PR exist?
Current implementation sets `Themes` tab as default both for free and pro users,

Closes #2722  

Set default tab as `Token sets` in case of free users.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/1293c2b7-b614-409a-8225-bc1b268d0a1c


packages/tokens-studio-for-figma/src/app/components/ManageStylesAndVariables/ManageStylesAndVariables.tsx
```diff
@@ -107,7 +111,7 @@ export default function ManageStylesAndVariables({ showModal, setShowModal }: {
   )}
         stickyFooter
       >
-        <Tabs defaultValue={isProUser ? 'useThemes' : 'useSets'}>
+        <Tabs defaultValue={activeTab}>
```
cuserox, line 114: I believe that this is the only necessary change to make it work. 

As we're setting the `activeTab` in the [initial state value, above](https://github.com/tokens-studio/figma-plugin/pull/2741/files#diff-119dd416293e5194ba6ecc2fef52242c3bac71925d949791e7030392c234d100R28), this will be correct at the time it renders, depending on the user's pro status.

The [useEffect to set the activeTab](https://github.com/tokens-studio/figma-plugin/pull/2741/files#diff-119dd416293e5194ba6ecc2fef52242c3bac71925d949791e7030392c234d100R84-R87) shouldn't be needed.

Video where I tested the suggested changes locally 🔨 

https://github.com/tokens-studio/figma-plugin/assets/114073780/f8aeeec6-99cd-4eb3-8eb3-5b13e06f2bf4

cuserox (changes requested):
[Small comment](https://github.com/tokens-studio/figma-plugin/pull/2741/files#r1593939355) to fix it with fewer changes 🤝

packages/tokens-studio-for-figma/src/app/components/ManageStylesAndVariables/ManageStylesAndVariables.tsx
robinhoodie0823: Removed.

Issue #2153, opened 2023-08-10
Allow for Truncation of Token Names in Detach Modal


keeganedwin:
**Describe the bug**
Currently in the manage themes section where you can detach styles & variables, long style names aren't truncated properly, causing the detach icon to be pushed off screen. 

**To Reproduce**
1. Create styles/attach styles
2. Find for long style names
3. See error

**Expected behavior**
Allow for the style names to be truncated properly to have the detach icon aligned. 

**Screenshots or Screencasts**
![image (10)](https://github.com/tokens-studio/figma-plugin/assets/93133012/eeab1630-68b6-455a-a309-3291ea1e6719)

PR #2761, opened 2024-05-16
fix: display full styling / variable info on theme management

cuserox:
### Why does this PR exist?

Closes [#2153](https://github.com/tokens-studio/figma-plugin/issues/2153#issue-1844490330)

Managing "Styles & Variables" within the theme editing modal would not display the full necessary info to attach or detach these. We now ensure the items don't overflow its parent container and add a tooltip on hover to provide all the details

### Testing this change

* Create a style or variables with long names and / or value
* Click on the 'Theme' dropdown ---> 'Manage Themes'
* Edit any theme ---> 'Styles & Variables' tab
* Note how the name, value and detach | attach option are well contained + tooltip appears on hover


LukeFinch (approved):

macintoshhelper (approved):
LGTM

PR #2762, opened 2024-05-16
Fix token value autosize

### LINKED ISSUE #2721: Token value field should grow in height
If users try to enter something like a long linear-gradient value and the value goes quite wide, the text input currently is not big enough to make sense of anything. Let's find a way where the input grows in height when users enter more.

We need to watch out for the prepending and appending button to the left so that those are still displayed correctly (likely vertical alignment top)

https://github.com/tokens-studio/figma-plugin/assets/4548309/886d07c0-e4be-467c-8c34-570cb37429e8

macintoshhelper:
### Why does this PR exist?

Fixes #2721 

> If users try to enter something like a long linear-gradient value and the value goes quite wide, the text input currently is not big enough to make sense of anything. Let's find a way where the input grows in height when users enter more.

MentionsInput height CSS is changed to auto; `MentionsInput` already has the `autosize` prop passed to it.

> We need to watch out for the prepending and appending button to the left so that those are still displayed correctly (likely vertical alignment top)

Handled by adding `height: auto` CSS to `<StyledPrefix ... />`. The prefix is automatically vertically aligned to the center.

### Testing this change

1. Open `Tokens` Tab
2. Scroll down to `Typography`
3. Create a Token with a very long name
4. Click `+` (Add New Token)
5. Under the Typography fields, try entering very long values – verify that the input `textarea` expands to multiline
6. Change to `Reference Mode`
7. Click the dropdown button, or mention the very long Typography token by typing `{` to bring up the mention dropdown
8. Verify that the input `textarea` expands to multiline

### Additional Notes (if any)

| Before | After |
|--------|-------|
| Before – Tokens / Typography / New Token | After – Tokens / Typography / New Token |
| Before – Tokens / Typography / New Token | After – Tokens / Typography / New Token |
| Before – Tokens / Typography / New Token | After – Tokens / Typography / New Token |
| Before – Tokens / Typography / New Token | After – Tokens / Typography / New Token |
| Before – Tokens / Typography / New Token | After – Tokens / Typography / New Token |
| Before – Tokens / Typography / New Token | After – Tokens / Typography / New Token |

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/DownshiftInput/DownshiftInput.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/DownshiftInput/DownshiftInput.tsx
@@ -192,8 +192,10 @@ export const DownshiftInput: React.FunctionComponent<React.PropsWithChildren<Rea
             {error ? <ErrorValidation>{error}</ErrorValidation> : null}
           </Stack>
           <Box css={{ display: 'flex', position: 'relative', width: '100%' }} className="input">
-            {!!inlineLabel && !prefix && <Tooltip label={name}><StyledPrefix isText>{label}</StyledPrefix></Tooltip>}
-            {!!prefix && <StyledPrefix>{prefix}</StyledPrefix>}
+            {!!inlineLabel && !prefix && (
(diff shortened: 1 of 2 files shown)


six7 (approved):
Nice!

mikekamminga:
Looks good 🚀

Issue #2618, opened 2024-04-11
[2.0]: Theme switching not working once variables are created


six7:
From https://hyma-team.slack.com/archives/C06RHB9RT2B/p1712769919381749

```
Once I have created variables from themes and exported to Figma, the themes feature in the plugin does not communicate with Figma, even for design elements on the canvas that only have design tokens applied.
This means the theme feature is non-functional after variables are created and is the highest priority item for teams who use tokens studio for more than just creating variables
Even when variables are connected, I want to use "apply to selection/page" to apply my tokens
I can see the plugin is scanning nodes but it does nothing
I can't even use token sets to theme the designs that have only tokens applied once variables are created
I want to be able to change themes (using the plugin) on design elements that have tokens directly applied, even if I have variables created in the same file.
This is especially important with non-local files feature.
I have variables linked to themes in file A
I have no intention of creating variables in file B
I can't use the theme switching in file B, its linking to the variables in file A
Existing issues in gitub related to theming in the thread
```

six7:
Closing this as it will be closed by #2213

PR #2781, opened 2024-05-20
2.0: let users choose if variables+styles should be applied

six7:
### Why does this PR exist?

Closes https://github.com/tokens-studio/figma-plugin/issues/2618

Gives users a choice if they want to apply styles+variables or raw values. If raw values is chosen, we will always apply raw values when applying to nodes. This effectively allows users to theme-switch even though they've created variables and styles. Before as we would always apply styles - we would not be able to react to theme changes unless the user had set up different variable modes.

CleanShot 2024-05-26 at 11 40 20@2x

### Testing this change

Play around with enabling/disabling those settings and try applying various tokens to layers that have styles / variables attached.


packages/tokens-studio-for-figma/src/app/components/ApplySelector.tsx
```diff
@@ -69,6 +78,8 @@ export default function ApplySelector() {
 
         <DropdownMenu.Portal>
           <DropdownMenu.Content side="top">
+            {/* TODO: Use DropdownMenu.Label - first add that to `ds` */}
```
LukeFinch, line 81: https://github.com/tokens-studio/ds/issues/197

packages/tokens-studio-for-figma/src/plugin/unbindVariableFromTarget.ts
```diff
@@ -1,11 +1,12 @@
 import { clone } from '@figma-plugin/helpers';
 
 export async function unbindVariableFromTarget(target: BaseNode | PaintStyle, key: 'paints' | 'fills' | 'strokes', newPaint: Paint) {
```
LukeFinch, line 3: With this now, it's possible for users to completely detach a system from styles and variables - super awesome!

LukeFinch (approved):
Only quickly tried this with a manual test - would be good to get a thorough manual QA check. 


Code is great - nice work!

PR #2789, opened 2024-05-24
Fix/no space after mentions

LukeFinch:
### Why does this PR exist?

When you use auto-suggestions from rc mentions, it would always add a space.
Turns out it's super simple to change the character we split suggestions with.

Adds `split=""` to the mentions component 

### Testing this change

Create a token which references another token.
Use mentions by typing `{` and selecting a suggestion

Notice there is no space after selection the mention now.


six7 (approved):
❤️  thank you!!

PR #2807, opened 2024-05-30
Deduplicate local tokens

### LINKED ISSUE #2779: When parsing tokens stored on the local document, first de-duplicate tokens
We had a bug that came with https://github.com/tokens-studio/figma-plugin/issues/2778 that basically means we store the same token name multiple times. Basically we had an array like this:

```
  [
    {
        "value": "4",
        "type": "sizing",
        "name": "foo.bar"
    },
    {
        "name": "foo.bar",
        "value": "14",
        "type": "sizing"
    }
]
```

This is a problem! Because we will have the same token exist multiple times.

To reproduce:

-  Create a token with a name `foo.bar` and another token with a name `otherfoo.bar`
- Right click the group `foo` and click `Rename`.
- Enter `otherfoo`.

Now if you go to `packages/tokens-studio-for-figma/src/plugin/node.ts` in `getTokenData` and add a console log to output the stored token values on line 118: `console.log('Getting tokens', values, themes, activeTheme, version, tokenFormat);` you will see that there's actually 2 tokens with the same name.

So, when we perform `getTokenData` in `packages/tokens-studio-for-figma/src/plugin/node.ts`, after parsing the tokens array from the document, let's de-duplicate that array and prefer the *last* one appearing (because likely users had been editing tokens and that's the token they've seen)

This is another reason to rather use a map-based token storage everywhere, but for this issue, let's just make sure that we fix existing user's data that's on our users documents by de-duplicating before we pass it on. This should make sure that users won't have faulty data stored.

https://github.com/tokens-studio/figma-plugin/issues/2778 will prohibit this from appearing in the first place but we should fix our stored data with this issue.

macintoshhelper:
### Why does this PR exist?

Fixes #2779  

- Adds a modal in the footer

Need to decide how to display the modal to the user on startup (will need more extensive testing around this since many users may experience it).

### Additional Notes (if any)

Will need a review of the text copy/locales surrounding this.

image

Highlighted with yellow circle

image


packages/tokens-studio-for-figma/src/app/components/DuplicateResolver/ResolveDuplicateTokenGroup.tsx
```diff
@@ -0,0 +1,73 @@
...
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'space-between',
+        marginBottom: '$3',
+      }}
+      key={`${groupKey}`}
+    >
+      <Stack direction="row" css={{ marginBottom: '$4' }}>
+        <Heading size="small" css={{ minWidth: '$9' }}>Token: </Heading>
+        <Heading size="small">{groupKey}</Heading>
```
six7, line 51:
```suggestion
        <Heading size="small"><Box as="span" css={{ minWidth: '$9' }}>Token: </Box>{groupKey}</Heading>
```

packages/tokens-studio-for-figma/src/app/components/DuplicateResolver/ResolveDuplicateTokenGroup.tsx
```diff
@@ -0,0 +1,73 @@
...
+      </Stack>
+      <Stack direction="row">
+        <Heading size="small" css={{ minWidth: '$9' }}>Value: </Heading>
```
six7, line 54:
```suggestion
        <Heading size="small" css={{ minWidth: '$9' }}>Value</Heading>
```

packages/tokens-studio-for-figma/src/app/components/DuplicateResolver/ResolveDuplicateTokenGroup.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <RadioGroup
+          onValueChange={onRadioClick}
+          value={checkedToken}
+          css={{ gap: '$3' }}
```
six7, line 58:
```suggestion
```

with the upgrade in the `@tokens-studio/ui` package to 0.6.9 this should not be needed anymore, if it is feel free to change back

packages/tokens-studio-for-figma/src/app/components/Footer.tsx
```diff
@@ -41,6 +44,20 @@ export default function Footer() {
   const { t } = useTranslation(['footer', 'licence']);
   const activeTheme = useSelector(activeThemeSelector);
   const { hasChanges } = useChangedState();
+  const tokens = useSelector(tokensSelector);
+  const [showResolveDuplicateTokensModal, setShowResolveDuplicateTokensModal] = React.useState<boolean>(false);
+
+  const hasDuplicates = useMemo(() => Object.keys(tokens).some((setName) => {
```
six7, line 50: something is off with this condition / memo, i *always* get that triangle in the bottom even on fresh files.
CleanShot 2024-07-01 at 08 54 01@2x

packages/tokens-studio-for-figma/src/app/components/Footer.tsx
```diff
@@ -41,6 +44,20 @@ export default function Footer() {
...
+    const currentSetTokens = tokens[setName];
+    return currentSetTokens.some((token) => {
+      const allTokensWithName = currentSetTokens.filter((a) => a.name === token.name);
```
six7, line 53: is this performant? a filter on a filter?

like why not just use that as the condition?

packages/tokens-studio-for-figma/src/app/components/Footer.tsx
```diff
@@ -140,6 +165,16 @@ export default function Footer() {
...
         </Box>
         <Stack direction="row" gap={1}>
+          {hasDuplicates && (
+            <IconButton
+              onClick={handleResolveDuplicateOpen}
+              icon={<WarningTriangleSolid />}
+              variant="invisible"
+              size="small"
+              tooltip="Duplicate Warning"
+              target="_blank"
```
six7, line 175:
```suggestion
```

this is not a link, no need for target blank

packages/tokens-studio-for-figma/src/app/components/modals/ResolveDuplicateTokensModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+          <Stack direction="column" key={`duplicateTokens-${setName}`}>
+            <Text bold css={{ fontSize: '$large', marginBottom: '$5' }}>{setName}</Text>
+            {Object.entries(allTokens).map(([tokenName, duplicates]) => (
+              <ResolveDuplicateTokenGroup
+                group={[tokenName, duplicates]}
+                setName={setName}
+                onRadioClick={onRadioClick}
+                selectedTokens={selectedTokens}
+                // tokens={tokens}
+                // resolvedTokens={resolvedTokens}
```
six7, line 124: remove

six7 (changes requested):
functionality wise i'm not sure if this should work already. i click resolve duplicates but nothing happens

can you add an e2e test so we verify on a test level that it does indeed work and we're safe for future changes?

https://github.com/tokens-studio/figma-plugin/assets/4548309/e1eadb65-1cfd-481a-89a0-c9ea352819c2

six7 (approved):
LGTM!

PR #2815, opened 2024-06-05
2804 20 gradients that have references to variables inside fail to applycreate

### LINKED ISSUE #2804: 2.0: Gradients that have references to variables inside fail to apply/create
If you have the following tokens:

```
{
  "gradient": {
    "value": "linear-gradient(212deg, {root} 0%, {dark} 100%)",
    "type": "color"
  },
  "root": {
    "value": "#00ff00",
    "type": "color"
  },
  "dark": {
    "value": "#0000ff",
    "type": "color"
  }
}
```

And you create variables and then try to apply that gradient, or create styles out of it, we currently fail.

It seems like the logic we're using to pass through references doesn't work as intended for gradients.

Getting the following error

Image

## To do
- [ ] If the variables exist that are being used here, it should use those variables inside the gradient
- [ ] If the variables do not exist, use the raw values
- [ ] Style creation should work as well, with both variables and non-variables
- [ ] Information on how to work with variables inside gradients: https://www.figma.com/plugin-docs/updates/2024/04/24/version-1-update-92/ (Paint.Colorstop can now be bound to variables)

robinhoodie0823:
### Why does this PR exist?
Current plugin dose not apply gradient token which have reference tokens.

Closes #2804  

Updated the value of gradient token when passing to function to apply token on node.
Updated `convertStringToFigmaGradient` function to handle several kinds of gradient tokens.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/1b55835a-c9d9-4638-bf40-6918c815ce3f


packages/tokens-studio-for-figma/src/plugin/setColorValuesOnTarget.ts
```diff
@@ -19,7 +19,7 @@ export default async function setColorValuesOnTarget({
     const resolvedToken = defaultTokenValueRetriever.get(token);
     if (typeof resolvedToken === 'undefined' && !givenValue) return;
     const { description } = resolvedToken || {};
-    const resolvedValue = defaultTokenValueRetriever.get(token)?.rawValue || givenValue;
```
cuserox, line 22: 👀 This error is also happening with **gradients that don't reference variables**.

📝 🔨 For example: I have `my-gradient: linear-gradient(#e66465, #9198e5)` and see another error like so:
```
...in set_fills: Expected "fills.[0].gradientTransform.0.0" to have type number but got NaN instead...
```

When passed to [convertStringToFigmaGradient](https://github.com/tokens-studio/figma-plugin/blob/2804-20-gradients-that-have-references-to-variables-inside-fail-to-applycreate/packages/tokens-studio-for-figma/src/plugin/setColorValuesOnTarget.ts#L35), with the value `linear-gradient(#e66465, #9198e5)`, it returns the following object:
```
gradientStops: { color: { a: 1, b: 0.8980392156862745, g: 0.596078431372549, r: 0.5686274509803921 }, position: NaN, ... },
gradientTransform: {  [NaN, NaN, NaN], [NaN, NaN, NaN] }
```
This `NaN` comes from the fact my `linear-gradient(#e66465, #9198e5)` doesn't have degrees, therefore [trying to convert it to degrees with **convertDegreeToNumber**](https://github.com/tokens-studio/figma-plugin/blob/2804-20-gradients-that-have-references-to-variables-inside-fail-to-applycreate/packages/tokens-studio-for-figma/src/plugin/figmaTransforms/gradients.ts#L21-L29) returns `NaN`, because we only have hex numbers.

Similarly, if I use `linear-gradient(0.25turn, #e66465, #9198e5)` or `linear-gradient(to top, #e66465, #9198e5)`, it will also fail ❌ 

🥁 Buuuut using `linear-gradient(45deg, #e66465, #9198e5)` - it still doesn't work 🤔 This is because we still get: `position: NaN` (all other values are now numbers)

✅ Finally, using this as a gradient works: `linear-gradient(45deg, #e66465 0%, #9198e5 100%)`

cuserox: Therefore, I believe the full fix for this would be to ensure that the [convertStringToFigmaGradient](https://github.com/tokens-studio/figma-plugin/blob/2804-20-gradients-that-have-references-to-variables-inside-fail-to-applycreate/packages/tokens-studio-for-figma/src/plugin/figmaTransforms/gradients.ts#L20) method has the correct fallbacks for when the string manipulation takes place, to avoid `NaN` results:
* If no angle degrees are given, default to `180deg` (not sure if that's the default?)
* If `to top` is given, translate to `0deg`
 
Anything else that could break it?! Suggestion 😇 Let me know if it makes sense and if you can reproduce my example.

cuserox (changes requested):

packages/tokens-studio-for-figma/src/plugin/setColorValuesOnTarget.ts
robinhoodie0823: @six7 I think we need to investigate `convertStringToFigmaGradient` function to handle several kinds of linear gradients.

"gradient": {
    "value": "linear-gradient(212deg, {root} 0%, {dark} 100%)",
    "type": "color"
  },
  "gradient1": {
    "value": "linear-gradient(212deg, #000000 0%, #ffffff 100%)",
    "type": "color"
  },
  "gradient2": {
    "value": "linear-gradient(#000000, #ffffff)",
    "type": "color"
  },
  "gradient3": {
    "value": "linear-gradient(#ffffff)",
    "type": "color"
  },

Only `gradient` works fine.

six7 (approved):

PR #2816, opened 2024-06-05
Fix empty modify on JSON save or edit

### LINKED ISSUE #2805: 2.0: Empty modify being saved to JSON
Some users reported seeing empty `modify` values in their JSON. This seems to happen randomly during duplication, or other scenarios. My guess is this is related to the [duplicate group issue](https://github.com/tokens-studio/figma-plugin/issues/2779) that we had.

In any way, we should clear the `modify` just like we clear it if the `studio.tokens` or `$extensions` field is empty.

To reproduce

- Go to JSON view
- Paste this JSON
```
{
 "blue": {
    "value": "#ff0000",
    "type": "color",
    "$extensions": {
      "studio.tokens": {}
    }
  },
  "red": {
    "value": "#ff0000",
    "type": "color",
    "$extensions": {
      "studio.tokens": {
        "modify": {}
      }
    }
  },
 "valid": {
    "value": "#ff0000",
    "type": "color",
    "$extensions": {
      "studio.tokens": {
        "modify": {
          "type": "darken",
          "space": "srgb",
          "value": "0.5"
        }
      }
    }
  }
}
```
- Notice that the "red" one keeps the empty modify.
- Notice that the "blue" one correctly removes the empty extensions / studio.extensions field

Expected:
In this example only the "valid" token which has a non-empty modify piece is being kept, the rest should be removed, so the expected output would be

```
{
 "blue": {
    "value": "#ff0000",
    "type": "color"
  },
  "red": {
    "value": "#ff0000",
    "type": "color"
  },
 "valid": {
    "value": "#ff0000",
    "type": "color",
    "$extensions": {
      "studio.tokens": {
        "modify": {
          "type": "darken",
          "space": "srgb",
          "value": "0.5"
        }
      }
    }
  }
}
```

macintoshhelper:
### Why does this PR exist?

Closes #2805  

- Adds a `validateStudioTokensExtensions` util function to strip empty `studio.tokens` extensions such as `modify`.
- Fix JSON not updating on first paste/edit by removing a `useEffect` dependency on `getStringTokens`.

| Before | After |
|-----:|---------------|
| image | image |
| image | image |

### Testing this change

To reproduce

1. Go to JSON view
2. Paste this JSON:

```json
{
 "blue": {
    "value": "#ff0000",
    "type": "color",
    "$extensions": {
      "studio.tokens": {}
    }
  },
  "red": {
    "value": "#ff0000",
    "type": "color",
    "$extensions": {
      "studio.tokens": {
        "modify": {}
      }
    }
  },
 "valid": {
    "value": "#ff0000",
    "type": "color",
    "$extensions": {
      "studio.tokens": {
        "modify": {
          "type": "darken",
          "space": "srgb",
          "value": "0.5"
        }
      }
    }
  }
}
```

**For existing tokens with empty `modify`:**

1. Right click and click `Edit Token`, then save.
2. The empty `modify` should be stripped.

### Additional Notes (if any)

A migration step would be nice to automatically strip empty extension objects when the plugin is opened/upgraded, but this would probably be better left for a future migration system with strong testing/error catching.


packages/tokens-studio-for-figma/src/app/components/Tokens.tsx
```diff
@@ -179,7 +179,7 @@ function Tokens({ isActive }: { isActive: boolean }) {
     // because of specific logic requirements
     setError(null);
     dispatch.tokenState.setStringTokens(getStringTokens());
-  }, [tokens, activeTokenSet, tokenFormat, tokenType, dispatch.tokenState, getStringTokens]);
+  }, [tokens, activeTokenSet, tokenFormat, tokenType, dispatch.tokenState]); // getStringTokens removed to fix bug around first paste/edit (useEffect was being triggered)
```
macintoshhelper, line 182: From my testing, on the first edit of `stringTokens`, `getStringTokens` was changing, triggering this `useEffect` to run with a race condition, wiping the initial change/paste.

This should fix it, but additional testing would be nice :)

packages/tokens-studio-for-figma/src/utils/validateStudioTokensExtensions.ts
```diff
@@ -0,0 +1,19 @@
+import { SingleToken } from '@/types/tokens';
+
+type StudioTokensExtension = NonNullable<SingleToken['$extensions']>['studio.tokens'];
+
+export default function validateStudioTokensExtensions(payload: SingleToken): StudioTokensExtension {
+  let studioTokensExtension;
+  if (typeof payload?.$extensions?.['studio.tokens'] !== 'undefined') {
+    studioTokensExtension = Object.keys(payload?.$extensions?.['studio.tokens']).reduce((accExt, k) => {
+      const extension = (payload?.$extensions?.['studio.tokens'] || {})[k];
+      if (extension && Object.values(extension).length > 0) {
```
macintoshhelper, line 10: This should look through every value of `studio.tokens`, and if an object with no values is found (i.e. `modify: {}`), it should filter it out (avoid adding it to the accumulator object).

packages/tokens-studio-for-figma/src/utils/validateStudioTokensExtensions.ts
cuserox, line 1: This is a great candidate to write some unit tests 🤓

cuserox:
✅ Tested by copy pasting the JSON, and it works!

🤔 This might not be in scope for the fix, but now that there's the `validateTokensStudioExtension` helper, it could be applied to other cases where these empty `modify` objects appear. I tried replicating [the issue being faced by this user](https://github.com/tokens-studio/figma-plugin/issues/2805#issuecomment-2151535239)(for example, the **border-radius** video)

With your branch locally:
* Go into renaming a token, which opens the modal and rename the token
* See the modal that asks to remap the tokens that use it
* Save it

😫 This also adds an empty modify object

My hunch is that the remapping action is adding it somewhere, maybe? `REMAP_TOKENS`

Screenshot 2024-06-06 at 16 48 46

https://github.com/tokens-studio/figma-plugin/assets/114073780/7681da89-cfb1-4089-853d-94c5690d0a55

six7 (approved):

PR #2817, opened 2024-06-06
2609 shorten error message when styles are not created due to ignore first part

### LINKED ISSUE #2609: Shorten error message when styles are not created due to ignore first part
**Describe the bug**
Error messages get truncated on longer messages

**Expected behavior**
We should make sure to shorten the error message for this case though, I'd suggest

`Some styles were ignored due to Ignore first part of token name setting`

image

robinhoodie0823:
### Why does this PR exist?
Current plugin has longer error msg content which get truncated.

Closes #2609 

Shorten error message content.


packages/tokens-studio-for-figma/src/plugin/updateStyles.ts
```diff
@@ -58,7 +58,7 @@ export default async function updateStyles(
     })] : []),
   ]).then((results) => Object.assign({}, ...results));
   if (styleTokens.length < tokens.length && shouldCreate) {
-    notifyUI('Some styles were not created due to your settings. Make sure Ignore first part of token name doesn\'t conflict', { error: true });
+    notifyUI('Some styles were ignored due to Ignore first part of token name setting\'t conflict', { error: true });
```
cuserox, line 61:
```suggestion
    notifyUI('Some styles were ignored due to "Ignore first part of token name" export setting', { error: true });
```
I know it's [slightly different than suggested](https://github.com/tokens-studio/figma-plugin/issues/2609), but it makes it clearer, no?

robinhoodie0823: I think so @cuserox 
@six7 Can you please share your opinion?

cuserox (approved):

Issue #2777, opened 2024-05-20
2.0: Editing JSON directly seems to cause some issues with W3C DTCG format


six7:
To reproduce:

- Have local document storage on
- Switch your token format to W3C DTCG
- Try to edit the JSON by adding a token manually in the JSON

Initial guess: Something seems to be wrong when stringifying tokens or parsing tokens.. We should definitely *NOT* try to do anything until that user hits save, but it seems like we are.

https://github.com/tokens-studio/figma-plugin/assets/4548309/a9cef11c-62e2-4afb-a0aa-4b383cbd0541

PR #2824, opened 2024-06-07
fix: prevent errors from appearing while editing JSON

cuserox:
### Why does this PR exist?

Closes [#2777](https://github.com/tokens-studio/figma-plugin/issues/2777#issue-2305858396) 

Removes the constant error appearing when editing JSON.

* Simply sets the JSON editor's value when it changes, without validation (`Tokens.tsx`)
* Moves the validation into `TokensBottomBar.tsx`, which is triggered when user clicks the **Save JSON** CTA

### Testing this change
* Open the plugin
* Go into the JSON tab
* Type in a new token, manually and click on **Save JSON**
   * If valid: it will persist
   * If not valid: it will show an error toast 

| Before | After |
|-----|-----|
| <video height="300" src="https://github.com/tokens-studio/figma-plugin/assets/114073780/b4d31946-eaa3-4d7e-bf33-3b49641eeefc"></video> | <video height="300" src="https://github.com/tokens-studio/figma-plugin/assets/114073780/cf766e4f-e47c-43de-b50a-50e752e68ada"></video> |

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/Tokens.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/Tokens.tsx
@@ -17,8 +17,6 @@ import JSONEditor from './JSONEditor';
 import Box from './Box';
 import IconListing from '@/icons/listing.svg';
 import useTokens from '../store/useTokens';
-import parseTokenValues from '@/utils/parseTokenValues';
-import parseJson from '@/utils/parseJson';
 import AttentionIcon from '@/icons/attention.svg';
 import { TokensContext } from '@/context';
 import {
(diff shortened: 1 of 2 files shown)


macintoshhelper (approved):
Seems to fix the issue of editing JSON in the `W3C` format.

Showing a valid `Save JSON` button while still editing the JSON, then showing an error only after clicking `Save JSON` feels like an antipattern a bit to me, but probably this is the best solution for now.

Longer term I think having line numbers on the side of the JSON editor together with a real-time error component above the editor, or inline like in VSCode would be nice.

six7 (approved):

PR #2826, opened 2024-06-07
2643 20 dtcg icon does not update when switching requires relaunch

### LINKED ISSUE #2643: 2.0: DTCG icon does not update when switching, requires relaunch
RE - DTCG (new) - when switching branches, the icon indicators don't accurately show the format
the only way to get the icons to be correct is to relaunch the plugin
(format appears to be functioning as expected, its just the icons)

https://hyma-team.slack.com/archives/C06RHB9RT2B/p1713048168200149

robinhoodie0823:
### Why does this PR exist?
Current plugin does not change the dtcg icon when switching branches which have different token format.

Closes #2643  

Added a new state function to update token format when switching branches.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/d11e774c-fb44-4f39-9963-b5b183839d78


six7 (commented):
Why not just call setTokenFormat when pulling?

robinhoodie0823:
> Why not just call setTokenFormat when pulling?

Because the `setTokenFormat` function needs TokenFormatOptions as parameter.

six7:
> > Why not just call setTokenFormat when pulling?
> 
> Because the `setTokenFormat` function needs TokenFormatOptions as parameter.

But so does the `setPulledTokenFormat`? its calling setTokenFormat in state - cant we directly call it and not introduce this new function? It feels like somewhere that setTokenFormat call is missing where it should be there?

(I'm just hesitant to add new state calls for when we might just be missing the right one)

six7 (approved):

PR #2835, opened 2024-06-10
2820 lineheight variables not binding and applying with styles

### LINKED ISSUE #2820: LineHeight Variables Not Binding and Applying With Styles
**Describe the bug**
If you have line-height tokens setup using `rem` values, create styles with variable references and apply the token, the line-height variable isn't applied. However, if you apply the line-height token as a particle in the typography type then create styles with variables references and apply the token, the variable is applied as well.

**To Reproduce**
Reproduction steps available in the test file link below.

**Expected behavior**
The line-height's token should also have its variable applied even if applied as a particle token and not just part of the typography token. 

https://github.com/tokens-studio/figma-plugin/assets/93133012/b6f10d4f-2fc4-4988-a7bb-9cb8b0336396

**Figma file (optional)**
[Test section](https://www.figma.com/design/zsk08ddoA82nm1GU4ye9MK/Untitled?node-id=2-11&t=clGHrYqCvAKLz0m2-4)

robinhoodie0823:
### Why does this PR exist?
Current plugin missed the logic to apply variables of typography property tokens like lineHeight, letterSpacing. It just applies token value.

Closes #2820  

Added a logic to `applyTypographyTokenOnNode` and `tryApplyTypographyCompositeVariables` functions to apply variables in case of typography property tokens.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/c317f9b4-18c3-4576-aa1b-cd730c0e04d7


six7 (changes requested):
Now the plugin is no longer able to apply typography tokens that do not have a variable attached. Try to *not* create a variable and apply, you'll notice it wont work.

six7 (approved):

PR #2836, opened 2024-06-11
2830 20 font weight tokens using numbers like 300 are not created properly

### LINKED ISSUE #2830: 2.0: Font weight tokens using numbers like 300 are not created properly
To reproduce:

Use this JSON:

```
{
  "normal": {
    "value": 300,
    "type": "fontWeights"
  }
}
```

try to create variables based on this.

Expected: `normal` is created as a numerical variable with 300 as its value
Actual result: `normal` is created with `String value` as its value.

Important: This was working fine a release candidate ago - it seems like there was a regression, the logic for handling numerical font weights should already be there

robinhoodie0823:
### Why does this PR exist?
Current plugin has problem for creating `fontWeight` tokens as variables in case of its value is a numeric value.

Closes #2830  

Figure out the problem and convert numeric value into string value for the function `convertFontWeightToFigma`.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/83a3cbd0-8fef-444d-a4c4-5a5edfb25219


six7 (changes requested):
This now creates a string variable for a font weight like `300`. Expected: It creates a number variable.

this diff should fix it: 
CleanShot 2024-06-11 at 09 50 45@2x

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
```diff
@@ -66,7 +66,7 @@ export default async function setValuesOnVariable(
               // Given we cannot determine the combined family of a variable, we cannot use fallback weights from our estimates.
               // This is not an issue because users can set numerical font weights with variables, so we opt-out of the guesswork and just apply the numerical weight.
             } else if (token.type === TokenTypes.FONT_WEIGHTS && Array.isArray(token.value)) {
-              setStringValuesOnVariable(variable, mode, String(token.value[0]));
+              setStringValuesOnVariable(variable, mode, Number(token.value[0]));
```
six7, line 69: this wont work, not every font weight token is number.

try it with 2 font weight tokens:
one is `Bold`
another is `500`

Expected:
`Bold` is created as a string variable
`500` is created as a number variable

we had this working already, something must've broken this

six7 (changes requested):

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
six7: Ignore this comment, seems to work! 👍

six7 (approved):

PR #2840, opened 2024-06-11
2838 20 no variables created note does not reflect actual created count

robinhoodie0823:
### Why does this PR exist?
Current plugin only calculates the update variables which have references when creating variables.

Closes #2839  

Updated the calculating logic to calculate the newly created variables when creating variables.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/32144463-ac1f-477a-8b16-c81ebb59e8e4


packages/tokens-studio-for-figma/src/plugin/createLocalVariablesInPlugin.ts
```diff
@@ -46,6 +50,7 @@ export default async function createLocalVariablesInPlugin(tokens: Record<string
       const allVariableObj = await updateVariables({
         collection, mode: modeId, theme, tokens, settings,
       });
+      figmaVariablesAfterCreate += allVariableObj.removedVariables.length;
```
six7, line 53: im not sure i understand. we add removed variables here? wouldnt we want to capture "removed variables" as a dedicated item to notify the user `5 variables removed` ?

six7 (commented):
just a question on the removedVariables part, not sure i understand why we add that and not note that to the user as a specific thing

packages/tokens-studio-for-figma/src/plugin/createLocalVariablesInPlugin.ts
robinhoodie0823: The reason why I added the amount of removed tokens is here to let the users know the exact number of created variables.
If user has one token and created as variables. After that, he removed that token and created a new one. In this case, if I do not include the removed variables amount here, then plugin will notify like this `No variables were created`.

six7: got it!

six7 (approved):

PR #2842, opened 2024-06-11
Fix/ min&max width&height on instances

### LINKED ISSUE #2426: Unable To Apply Max-width Tokens On Component Instances
**Describe the bug**
When applying the ﻿max-width token directly to an instance, it does not get applied/override. The ﻿max-width override however works as expected when applied to the master component.

> This is quite impactful for us because it means not only are we  unable to see proper max-widths on the instances with theme/set overrides, but if a designer accidentally types in a max-width value on the instance, it sticks i.e. the tokenized value from the master never resets it back to what it should be.

This also seems similar to #2270 

**To Reproduce**
1. Select main component
2. Check `Test` set
3. See change
4. Uncheck `Test` set
5. Select component instance
6. Check `Test` set
7. See error

**Expected behavior**
Max/min-width tokens should also override on instances.

https://github.com/tokens-studio/figma-plugin/assets/93133012/68d3ff55-3683-425c-8d84-ac1e0fc87094

**Figma file (optional)**
[Test board](https://www.figma.com/file/egeTaCchJqdTFfoJJK8vIZ/Max-Width-issue-on-instance?type=design&node-id=11-71&mode=design&t=xGdM0dZLMVs0hJun-4)

LukeFinch:
### Why does this PR exist?

Closes #2426 

Changes the logic to apply min and max-width on instances

### Testing this change

Create a component with autolayout
Create an instance of the component
Apply min/max width/height sizing through the plugin

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/applySizingValuesOnNode.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/applySizingValuesOnNode.ts
@@ -3,7 +3,6 @@ import { NodeTokenRefMap } from '@/types/NodeTokenRefMap';
 import { isPrimitiveValue } from '@/utils/is';
 import { tryApplyVariableId } from '@/utils/tryApplyVariableId';
 import { transformValue } from './helpers';
-import { isPartOfInstance } from '@/utils/is/isPartOfInstance';
 import { isAutoLayout } from '@/utils/isAutoLayout';
 
 export async function applySizingValuesOnNode(
--- a/packages/tokens-studio-for-figma/src/plugin/applySizingValuesOnNode.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/applySizingValuesOnNode.ts
@@ -4,6 +4,7 @@ import { isPrimitiveValue } from '@/utils/is';
 import { tryApplyVariableId } from '@/utils/tryApplyVariableId';
 import { transformValue } from './helpers';
 import { isAutoLayout } from '@/utils/isAutoLayout';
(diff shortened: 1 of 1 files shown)


robinhoodie0823 (approved):

packages/tokens-studio-for-figma/src/plugin/applySizingValuesOnNode.ts
```diff
@@ -54,8 +53,6 @@ export async function applySizingValuesOnNode(
   if (
     node.type !== 'DOCUMENT'
     && node.type !== 'PAGE'
-    && node.type !== 'INSTANCE'
-    && !isPartOfInstance(node.id)
```
six7, line 58: I think this part here needs to stay in - we cant apply min max width tokens on instance children, this would throw an error

CleanShot 2024-06-12 at 10 07 18@2x

six7 (changes requested):

packages/tokens-studio-for-figma/src/plugin/applySizingValuesOnNode.ts
cuserox: If this stays, then this is not the fully fledged fix? Can't see any other changes 😅

six7: It basically just resolves the case descrbied in the issue: if the layer itself is an instance, we want to allow setting sizing related tokens (as this is possibly). However, if its *part of an instance* this wouldnt work because figma doesnt allow it.

six7 (approved):

PR #2850, opened 2024-06-12
Fix/ when switching to input mode from a referenced Border token, all…

### LINKED ISSUE #2823: Overriding border token doesn’t reflect the values in the token
**Describe the bug**
When trying to override an existing border token with "input mode," the token does not accept the new values. As a result, we cannot update the border token as intended.

**To Reproduce**
1. Navigate to the plugin 
2. create a border token with respected values ex: `border.sm`
3. now create another border token ex: `border.card` referring to the border.sm 
4. now edit the `border.card`, switch to input mode, and try to change the values and see the error.

**Expected behavior**
When trying to override an existing border token with "input mode," the token should accept the new values. As a result, we should be able to update the border token as intended.

https://github.com/tokens-studio/figma-plugin/assets/140593995/bfa8c6a7-92b8-4361-a70f-6aa49dea5a81

**Figma file (optional)**
[Test File
](https://www.figma.com/design/1KfJpbGnUYs41XzUsuRpG1/Untitled?node-id=1-5&t=ozDDmUNUNKw3XhS2-1)

akshay-gupta7:
Fixes #2823 

- When a user will switch to input mode from a border token which already points to a reference, the fields color, width and stroke style shall be pre-populated with the referenced values and the user will be able to choose values to override them

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/BorderTokenForm.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/BorderTokenForm.tsx
@@ -4,7 +4,7 @@ import get from 'just-safe-get';
 import { TokensIcon, LinkBreak2Icon } from '@radix-ui/react-icons';
 import { useTranslation } from 'react-i18next';
 import { IconButton, Heading } from '@tokens-studio/ui';
-import { EditTokenObject } from '@/types/tokens';
+import { EditTokenObject, SingleBorderToken } from '@/types/tokens';
 import { TokenTypes } from '@/constants/TokenTypes';
 import { ResolveTokenValuesResult } from '@/utils/tokenHelpers';
 import Stack from './Stack';
@@ -28,6 +28,7 @@ export default function BorderTokenForm({
   handleBorderValueDownShiftInputChange,
   handleBorderAliasValueChange,
   handleDownShiftInputChange,
+  setBorderValue,
   onSubmit,
 }: {
   internalEditToken: Extract<EditTokenObject, { type: TokenTypes.BORDER }>;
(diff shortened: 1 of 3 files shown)


packages/tokens-studio-for-figma/src/app/components/EditTokenForm.tsx
```diff
@@ -265,14 +277,29 @@ function EditTokenForm({ resolvedTokens }: Props) {
...
+    (newBorderValue: SingleBorderToken['value']) => {
+      if (internalEditToken?.type === TokenTypes.BORDER && typeof newBorderValue === 'object') {
+        setInternalEditToken({
+          ...internalEditToken,
+          value: { ...newBorderValue },
+        });
+      }
+    },
+    [internalEditToken],
+  );
```
akshay-gupta7, line 302: This is the relevant part of the change in this file, I ran prettier command before committing which made lots of formatting changes

packages/tokens-studio-for-figma/src/app/components/BorderTokenForm.tsx
```diff
@@ -36,61 +37,65 @@ export default function BorderTokenForm({
...
+            icon={<LinkBreak2Icon />}
+          />
+        )}
       </Stack>
-      {(mode === 'input' && internalEditToken.schema.schemas.value.type === 'object') ? (
+      {mode === 'input' && internalEditToken.schema.schemas.value.type === 'object' ? (
         <Stack gap={2} direction="column">
           {Object.entries(internalEditToken.schema.schemas.value.properties ?? {}).map(([key], keyIndex) => (
-            <>
+            <React.Fragment key={`border-input-fragment-${seed(keyIndex)}`}>
```
akshay-gupta7, line 98: There was a react child unique key error, hence I replaced the blank fragment with a React Fragment(with a key)

packages/tokens-studio-for-figma/src/app/components/BorderTokenForm.tsx
```diff
@@ -36,61 +37,65 @@ export default function BorderTokenForm({
...
   const handleMode = React.useCallback(() => {
-    const changeMode = (mode === 'input') ? 'alias' : 'input';
-    setMode(changeMode);
+    if (mode === 'alias' && typeof internalEditToken.value === 'string') {
+      setBorderValue(selectedToken?.rawValue ?? {});
+    }
+    setMode(mode === 'input' ? 'alias' : 'input');
     setAlias('');
-  }, [mode]);
+  }, [mode, selectedToken, internalEditToken, setBorderValue]);
```
akshay-gupta7, line 73: The pertinent change in this file. Lots of prettier formatting in this file as well

PR #2854, opened 2024-06-14
2822 gitlab sync is not allowing users to push tokens

### LINKED ISSUE #2822: Gitlab Sync is Not Allowing Users To Push Tokens
**Describe the bug**
A bit of a challenge to reproduce this bug, but users using the Gitlab sync claim that every time they push tokens to the repo it doesn't actually push them; blue indicators on both `push` and `pull` icons. Or even if it does push, when they close and reopen the plugin it reverts back to a previous version. This is a major blocker for them so we need to try and get to the bottom of it asap. 

It also seems to be related to the following tickets:
1. #2320 
2. #1636 
3. #2652 

**To Reproduce**
1. Enable gitlab sync
4. Create/edit tokens
5. Push tokens
6. See error - I've both blue indicators **displayed** on both rc7 and the 1.38.9 the first time I tested the issue. Subsequent times not so much luck.

https://github.com/tokens-studio/figma-plugin/assets/93133012/0c888f2e-6656-4c3d-976b-0127d49c79b9

robinhoodie0823:
### Why does this PR exist?
Current plugin has no feature to let the users know if there is an error during pushing tokens to remote provider.

Closes #2822  

Added a feature, so show message to users if there is an error during pushing tokens to remote providers.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/6bb3742b-9708-41bc-bb1f-73d735fcdaa9


packages/tokens-studio-for-figma/src/app/store/providers/gitlab/gitlab.tsx
```diff
@@ -101,7 +101,7 @@ export function useGitLab() {
           themes,
           metadata: {},
         };
-      } catch (e) {
+      } catch (e: any) {
```
cuserox, line 104:
```suggestion
      } catch (e: Error) {
```

This would save checking `e instanceof Error` below, no?

packages/tokens-studio-for-figma/src/app/store/remoteTokens.tsx
```diff
@@ -325,34 +326,38 @@ export default function useRemoteTokens() {
...
         }
         case StorageProviderType.TOKENS_STUDIO: {
-          await pushTokensToTokensStudio(context);
+          pushResult = await pushTokensToTokensStudio(context);
           break;
         }
         default:
           throw new Error('Not implemented');
       }
+      if (pushResult.status && String(pushResult.status).includes('failure')) {
```
cuserox, line 358: Why does the status need to be cast to a `string`? Is it not always `status: 'failure'` or `status: 'success'`? Then it should work.

```suggestion
      if (pushResult.status && pushResult.status === 'failure')) {
```

packages/tokens-studio-for-figma/src/app/store/remoteTokens.tsx
```diff
@@ -325,34 +326,38 @@ export default function useRemoteTokens() {
...
+        notifyToUI(String(pushResult.errorMessage), { error: true });
```
cuserox, line 359: Similar question, in what cases are the `errorMessage`s not a string? We should be handling these to being strings before throwing.

packages/tokens-studio-for-figma/src/constants/ErrorMessages.ts
cuserox, line 1: These messages should be translated and offer an extensive list. Created a ticket: https://github.com/tokens-studio/figma-plugin/issues/2856

cuserox (approved):

packages/tokens-studio-for-figma/src/app/store/providers/gitlab/gitlab.tsx
robinhoodie0823: No. The error type `Gitbeaker` module returns is not Error.
So it need to be flexible.

packages/tokens-studio-for-figma/src/app/store/remoteTokens.tsx
robinhoodie0823: The error that `Gitbeaker` returns like this "403 Forbidden - ...". Also contains `"`.
So that is why I used `includes` function to check the error message.

packages/tokens-studio-for-figma/src/app/store/remoteTokens.tsx
robinhoodie0823: Updated.

packages/tokens-studio-for-figma/src/app/store/remoteTokens.tsx
robinhoodie0823: Updated

PR #2855, opened 2024-06-14
Fix base font size display in settings

### LINKED ISSUE #2845: No UI Indication of The Base Font-size Not Actually Being 1rem
**Describe the bug**
The recording shows a token applied as a base font-size that is not being displayed unless you click the “change” button. There’s no indication in the UI that the base font size is not actually 1rem.

**To Reproduce**
1. Create a `base.fontSize` token in the font size type with a rem value
2. Reference this token as the Base font size in the `Settings` tab
3. Save

**Expected behavior**
base font-size should be displayed once referenced in the `Settings` tab

https://github.com/tokens-studio/figma-plugin/assets/93133012/78895d1d-2c44-4bdf-9e1d-0d073d340eb1

akshay-gupta7:
- When a user changes the base font size from settings, either directly or via a font size token, the rem value will now be displayed in the settings with the current base font size(Default being 1 rem)

Closes #2845

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
@@ -41,14 +41,16 @@ const RemConfiguration = () => {
     const resolvedAliasBaseFontSize = getAliasValue(aliasBaseFontSize, resolvedTokens);
 
     if (typeof resolvedAliasBaseFontSize === 'string' || typeof resolvedAliasBaseFontSize === 'number') {
-      const remValue =
+      const resolvedAliasBaseFontSizeValue =
         typeof resolvedAliasBaseFontSize === 'number'
-          ? resolvedAliasBaseFontSize / 16
-          : parseFloat(resolvedAliasBaseFontSize) / 16;
-      const formattedRemValue = isNaN(remValue) ? 1 : Number(remValue.toFixed(2));
-      return `${t('baseFont')} (${formattedRemValue} rem)`;
+          ? resolvedAliasBaseFontSize
+          : parseFloat(resolvedAliasBaseFontSize);
+      const formattedpxValue = isNaN(resolvedAliasBaseFontSizeValue)
+        ? 16
+        : Number(resolvedAliasBaseFontSizeValue.toFixed(2));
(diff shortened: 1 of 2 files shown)


packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
```diff
@@ -31,15 +27,30 @@ const RemConfiguration = () => {
...
+  const displayBaseFontValue = React.useMemo(() => {
+    const resolvedAliasBaseFontSize = getAliasValue(aliasBaseFontSize, resolvedTokens);
+
+    if (typeof resolvedAliasBaseFontSize === 'string' || typeof resolvedAliasBaseFontSize === 'number') {
+      const remValue =
+        typeof resolvedAliasBaseFontSize === 'number'
+          ? resolvedAliasBaseFontSize / 16
+          : parseFloat(resolvedAliasBaseFontSize) / 16;
+      const formattedRemValue = isNaN(remValue) ? 1 : Number(remValue.toFixed(2));
+      return `${t('baseFont')} (${formattedRemValue} rem)`;
```
six7, line 49: the calculation you have here is for outputting `rem`.

i'd say lets output `px` instead, because that field really lets the user define the value of `1rem`. so `1rem` doesnt ever change, its just the value it has.

for example

`1rem = 16px` (default)
`1rem = 20px` (base font size token = 20px)

so instead of displaying `(1.5rem)` here, lets show `(16px)` or `(20px)` here and in the outer settings page

CleanShot 2024-06-15 at 11 32 41@2x

packages/tokens-studio-for-figma/src/i18n/lang/en/settings.json
```diff
@@ -6,7 +6,7 @@
   "confirm": "Confirm",
   "storeTokenId": "Store IDs on storage",
   "storeTokenIdExplanation": "Automatically generate unique IDs for your tokens for tracking.",
-  "baseFont": "Base font size (1rem)",
+  "baseFont": "Base font size",
```
six7, line 9: can you also make these changes in the other languages?

akshay-gupta7: Other languages already don't include the 1rem

six7 (approved):

PR #2868, opened 2024-06-18
2742 errors caused with importing variables depending on selected options 1

### LINKED ISSUE #2742: Errors Caused With Importing Variables Depending On Selected Options
**Describe the bug**
So currently when a user has number variables setup and imports them with the `Convert Numbers to Dimensions` option, number variables are converted to tokens in the plugin under the Dimension type. 

**Errors Caused:**
1. If you import number variables with `Use REM for Dimension Values`, you get tokens under the Number type with `[object object]` as values. One assumption is that the plugin is trying to use REM values for Dimension tokens that don't exist as they're imported as Number tokens.

2. If you import number variables with both options selected `Convert Numbers to Dimensions` and `Use REM for Dimension Values`, this results in converting number variables to tokens under the Dimension type with `NaNrem` as values. 

**To reproduce:**
Please check test file attached below

**Expected behavior**
With this option selected, values should be imported with REM values

Recording shows token values with [object object] when the `Use REM for Dimension Values` option is selected on importing. 
https://github.com/tokens-studio/figma-plugin/assets/93133012/4115b30e-79fb-4efa-ae18-ae46f21c7f56

**Figma file (optional)**
[Test file](https://www.figma.com/file/4OGLxYtquLN9ZO094TKXoP/Untitled?type=design&node-id=1-35&mode=design&t=BcFWV8G9jz8Ol3pE-4)

robinhoodie0823:
### Why does this PR exist?
Current plugin has errors in importing variables based on dimension options.

Closes #2742 

Updated logic to give first priority to Rem option.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/a1dcf59f-379e-4fb8-b3a5-a772e8c6905d


packages/tokens-studio-for-figma/src/plugin/pullVariables.test.ts
```diff
@@ -84,6 +115,352 @@ describe('pullStyles', () => {
...
+          parent: 'Collection 1/Custom',
+          type: 'number',
+          value: '16'
+        }
+      ]
+    });
+  });
+
+  it('pulls variables with Convert numbers to dimensions option', async () => {
+    // @ts-ignore next-line
```
six7, line 180: instead of just ignoring, can you add a note *why* we're ignoring - we should strive to not ignore anything but if we have to we should know why.

packages/tokens-studio-for-figma/src/plugin/pullVariables.test.ts
```diff
@@ -84,6 +115,352 @@ describe('pullStyles', () => {
...
+          parent: 'Collection 1/Custom',
+          type: 'dimension',
+          value: '1rem'
+        }
+      ]
+    });
+  });
+
+  it('pulls variables with dimensions in rem if both options are selected', async () => {
+    // @ts-ignore next-line
```
six7, line 372: same here

six7 (changes requested):
I still get `NaNrem`. Can you make sure that the value we're importing is a number?

CleanShot 2024-06-19 at 09 05 17@2x

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -94,14 +94,16 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
...
-          } else if (options.useDimensions && options.useRem) {
-            tokenValue = `${Number(tokenValue) / baseRem}rem`;
-          } else {
-            tokenValue = `${tokenValue}px`;
+          } else if (typeof value === 'number') {
+            if (options.useRem) {
+              tokenValue = `${Number(tokenValue) / baseRem}rem`;
+            } else if (options.useDimensions) {
+              tokenValue = `${tokenValue}px`;
+            }
```
cuserox, line 102: Should there be a fallback value for `tokenValue` if neither of the options are true, but `typeof value === 'number'` is true?

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -94,14 +94,16 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
...
           }
           const modeName = collection?.modes.find((m) => m.modeId === mode)?.name;
 
-          if (options.useDimensions) {
+          if (options.useDimensions || options.useRem) {
             dimensions.push({
               name: variableName,
               value: tokenValue as string,
```
cuserox, line 109: Does `type: TokenTypes.DIMENSION` still hold true when it's REM?

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
robinhoodie0823: This is for checking to prevent `NaNRem` issue

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
six7: Confirming this - when we have a unit attached it's going to be a `dimension` token (technically it could be a sizing or spacing one too, but we dont know that from the number. And a number token would come *without* a unit, a dimension token *with* a unit, so we can go with that one

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -94,10 +94,12 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
           if (typeof value === 'object' && 'type' in value && value.type === 'VARIABLE_ALIAS') {
             const alias = figma.variables.getVariableById(value.id);
             tokenValue = `{${alias?.name.replace(/\//g, '.')}}`;
-          } else if (options.useRem) {
-            tokenValue = `${Number(tokenValue) / baseRem}rem`;
-          } else if (options.useDimensions) {
-            tokenValue = `${tokenValue}px`;
+          } else if (typeof value === 'number') {
+            if (options.useRem) {
+              tokenValue = `${Number(tokenValue) / baseRem}rem`;
```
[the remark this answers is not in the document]
six7: parseFloat(baseRem( would do the trick

six7 (changes requested):

six7 (approved):

Issue #2876, opened 2024-06-19
Create Variables options: Create as variables / styles for Color settings do not seem to persist


six7:
Whenever you restart the plugin we seem to reset the state of the Create as Varaibles (Color) / Create as Styles (Color) option to have *both* active.

We should remember that choice, something seems to be missing about the way how we save these settings.

Also, the default should never be that both are active - let's just active `Variables: Color` if no value is set.

PR #2877, opened 2024-06-19
fix: variables styles color remember

six7:
Fixes: https://github.com/tokens-studio/figma-plugin/issues/2876

This sets the default for those options when no value is provided correctly, as in if nothing is stored locally, we just set variablesColor to true, not both.


macintoshhelper (approved):
LGTM

PR #2886, opened 2024-06-19
feat: Enhance base font size display in settings tab

### LINKED ISSUE #2875: Show base font size token even outside
Currently we only show the used base font size token when you're in the Change modal of the base font size token. Currently we show the resolved value outside, but not what value that token or value has.

Users mentioned this is too hidden, they need to see whats stored there given this can differ between projects.

Let's show the token from the outside, but keep the rest of the behavior the same.

- If no value is set, just show default of 16px
- If value is set, show value and show resolved value (e.g. {token-base} 24px)
- If value is set but is not a reference, just show the value (do a string check if it starts with {)
- If value is set to a reference but we failed to resolve, show it in a broken state and show the default value that we fall back to (16)

Image

akshay-gupta7:
- Now base font size's reference will be displayed directly outside in the settings tab
- when the referenced token is deleted or invalid, it shall display the broken link in red
- Resolves #2875 
Screenshot 2024-06-20 at 00 36 13
Screenshot 2024-06-20 at 00 36 44

In dark mode:

Screenshot 2024-06-20 at 11 46 09

Screenshot 2024-06-20 at 11 46 41

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
@@ -101,24 +101,24 @@ const RemConfiguration = () => {
                 display: 'flex',
                 alignItems: 'center',
                 backgroundColor: isBrokenLink ? '$dangerBg' : '$accentBg',
-                padding: '2px',
+                padding: '$2',
               }}
             >
-              <Label css={{ fontSize: '12px', color: isBrokenLink ? '$dangerFg' : '$accentDefault' }}>
+              <Label css={{ fontSize: '$small', color: isBrokenLink ? '$dangerFg' : '$accentDefault' }}>
                 {aliasBaseFontSize}
               </Label>
               {isBrokenLink && (
                 <IconBrokenLink
-                  style={{ color: '#e03e1a', width: '24px', height: '24px', marginLeft: '3px', fontWeight: 'bold' }}
+                  style={{ color: 'var(--colors-dangerFg)', width: 'var(--sizes-6)', height: 'var(--sizes-6)', marginLeft: '3px' }}
                 />
               )}
             </Box>
           )}
           <Label
             css={{
-              fontSize: '12px',
+              fontSize: '$small',
               color: '$fgSubtle',
-              marginLeft: aliasBaseFontSize.startsWith('{') ? '8px' : '0px',
+              marginLeft: aliasBaseFontSize.startsWith('{') ? '$3' : '',
             }}
           >
             {displayBaseFontValue}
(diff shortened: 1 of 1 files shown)


packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
```diff
@@ -77,12 +86,44 @@ const RemConfiguration = () => {
...
+                padding: '4px',
+              }}
+            >
+              <Label css={{ fontSize: '12px', color: isBrokenLink ? '$dangerFg' : '$accentDefault' }}>
+                {aliasBaseFontSize}
+              </Label>
+              {isBrokenLink && (
+                <IconBrokenLink
+                  style={{ color: '#e03e1a', width: '24px', height: '24px', marginLeft: '3px', fontWeight: 'bold' }}
+                />
```
akshay-gupta7, line 113: I used color as a fixed value here, for some reason the place $dangerFg wasn't resolving here

six7: The $ notation for stitches will only work when stitches is used for example with the `styled()` function or when used on our internal ds components like `css={{color: $dangerFg}}`

You can always use the css variable too in this case `style={{color: 'var(--colors-dangerFg)'}}` (not sure on the exact color name you might want to double check

packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
```diff
@@ -77,12 +86,44 @@ const RemConfiguration = () => {
...
+              )}
+            </Box>
+          )}
+          <Label
+            css={{
+              fontSize: '12px',
```
six7, line 119:
```suggestion
              fontSize: '$small',
```

packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
```diff
@@ -37,20 +39,27 @@ const RemConfiguration = () => {
...
     if (typeof resolvedAliasBaseFontSize === 'string' || typeof resolvedAliasBaseFontSize === 'number') {
       const resolvedAliasBaseFontSizeValue =
         typeof resolvedAliasBaseFontSize === 'number'
           ? resolvedAliasBaseFontSize
           : parseFloat(resolvedAliasBaseFontSize);
-      const formattedpxValue = isNaN(resolvedAliasBaseFontSizeValue)
-        ? 16
-        : Number(resolvedAliasBaseFontSizeValue.toFixed(2));
-      return `${t('baseFont')} (1rem = ${formattedpxValue}px)`;
+      if (isNaN(resolvedAliasBaseFontSizeValue)) {
```
six7, line 53: Does this work when the baseFontSize token is a token with a unit attached? Such as a dimension token with 24px as a value? It should!

akshay-gupta7: Yes! It works when you type in px as well

packages/tokens-studio-for-figma/src/app/components/RemConfiguration.tsx
akshay-gupta7: Aah, alright, got it!!

six7 (approved):

PR #2910, opened 2024-06-24
2894 20 gradients inside variable creation should only use variables if setting is on

### LINKED ISSUE #2894: 2.0: Gradients inside variable creation should only use variables if setting is on
We have a setting here when creating variables that lets the user choose if variables should be used.

Right now gradients that use variables do not respect this setting and *always* create with variables.

Look how we do that for tryApplyVariableId - we have a check in there that detects that setting

Image

Image

robinhoodie0823:
### Why does this PR exist?
Current plugin does not respect export options in case of creating styles of linear gradient color tokens.

Closes #2894  

Updated the creating styles for linear gradient tokens logic to respect the export setting.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/e29322a6-1145-4489-8adf-d1470dbdb2a4


packages/tokens-studio-for-figma/src/plugin/setColorValuesOnTarget.ts
```diff
@@ -1,3 +1,4 @@
+import { ApplyVariablesStylesOrRawValues } from '@/constants/ApplyVariablesStyleOrder';
```
six7, line 1: this line is unused - your linter should throw an error here?

six7 (approved):
just the import

Issue #2906, opened 2024-06-22
2.0: Non local variables issues


six7:
Non local variables sometimes dont pick up references, when the theme groups that contain the references arent active.

Slack: https://tokens-studio.slack.com/archives/C073SSG7DC7/p1718989346999029

PR #2911, opened 2024-06-24
2.0: Fix variable references when references aren't coming from any active themes

six7:
Closes https://github.com/tokens-studio/figma-plugin/issues/2906

When we're creating variables, we currently use themes that are active to inform what variables should be used as references. However, if those theme groups aren't active, we won't know what variable references to use for tokens and will fall back to raw hex values inside variables. Especially in multi-file setup users sometimes only enable one theme group, meaning they won't have the necessary references available.

Changes the logic to add another set of themes to the merge function, so that the references from the current theme group have preference, but after that we'll take into account other themes. 

Note that this is kind of flawed as we will pick the first available variable reference that we find. However, I'd say it's OK enough, given:
- We should educate users that they should keep tokens only to one theme group
- Eventually we should refactor themes so that users can directly specify what theme group a variable reference should be coming from

**Before:**
CleanShot 2024-06-24 at 08 32 53@2x

**After:** 
CleanShot 2024-06-24 at 08 32 33@2x

### Testing this change

- Create multiple theme groups
- One theme group should reference token from another
- Create a multi-file setup (you need to push to git to do that)
- Create and Publish variables from the first theme group in file A
- Go to file B, activate the library from file A via Figma
- Create variables from the second theme group (that's using tokens from the first one) in file B

With this change, the references correctly map. Before, they used hex values.

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
@@ -1,10 +1,23 @@
 import { ThemeObject } from '@/types';
 
-export async function mergeVariableReferencesWithLocalVariables(themes: ThemeObject[] = []): Promise<Map<string, string>> {
+// Gather references that we should use. Merge current theme references with the ones from all themes as well as local variables
+// Note that this is a bit naive. As the active theme does not have the themes set as active that it should use as a reference, it will not be able to find the correct reference.
+// This is why we also pass in allThemes and merge them together. This is naive because it might be that the reference required is not the "first to detect".
+// We need to refactor this probably. Themes should be able to strictly specify which other theme the references are coming from, like we do for token sets.
+export async function mergeVariableReferencesWithLocalVariables(themes: ThemeObject[] = [], allThemes: ThemeObject[] = []): Promise<Map<string, string>> {
   const localVariables = await figma.variables.getLocalVariablesAsync();
 
   const variables = new Map();
   themes.forEach((theme) => {
+    if (!theme.$figmaVariableReferences) return;
+    Object.entries(theme.$figmaVariableReferences).forEach(([tokenName, variableId]) => {
+      // For each variable reference we add one to the global Map - ONLY if it is not already set.
+      // Meaning, users will run into problems if they have a token defined in multiple theme groups
+      if (variables.has(tokenName)) return;
+      variables.set(tokenName, variableId);
+    });
+  });
+  allThemes.forEach((theme) => {
     if (!theme.$figmaVariableReferences) return;
     Object.entries(theme.$figmaVariableReferences).forEach(([tokenName, variableId]) => {
       // For each variable reference we add one to the global Map - ONLY if it is not already set.
--- a/packages/tokens-studio-for-figma/src/plugin/createLocalVariablesInPlugin.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/createLocalVariablesInPlugin.ts
@@ -61,12 +61,14 @@ export default async function createLocalVariablesInPlugin(tokens: Record<string
       }
       updatedVariableCollections.push(collection);
     }));
-    const existingVariables = await mergeVariableReferencesWithLocalVariables(selectedThemeObjects);
+    // Gather references that we should use. Merge current theme references with the ones from all themes as well as local variables
+    const existingVariables = await mergeVariableReferencesWithLocalVariables(selectedThemeObjects, themeInfo.themes);
 
+    // Update variables to use references instead of raw values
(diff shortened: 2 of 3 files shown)


packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
```diff
@@ -13,6 +17,15 @@ export async function mergeVariableReferencesWithLocalVariables(themes: ThemeObj
       variables.set(tokenName, variableId);
     });
   });
+  allThemes.forEach((theme) => {
+    if (!theme.$figmaVariableReferences) return;
+    Object.entries(theme.$figmaVariableReferences).forEach(([tokenName, variableId]) => {
+      // For each variable reference we add one to the global Map - ONLY if it is not already set.
+      // Meaning, users will run into problems if they have a token defined in multiple theme groups
```
cuserox, line 24: Is this "problem" documented somewhere, and how to mitigate if so?

packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
```diff
@@ -1,6 +1,10 @@
 import { ThemeObject } from '@/types';
 
-export async function mergeVariableReferencesWithLocalVariables(themes: ThemeObject[] = []): Promise<Map<string, string>> {
+// Gather references that we should use. Merge current theme references with the ones from all themes as well as local variables
+// Note that this is a bit naive. As the active theme does not have the themes set as active that it should use as a reference, it will not be able to find the correct reference.
+// This is why we also pass in allThemes and merge them together. This is naive because it might be that the reference required is not the "first to detect".
+// We need to refactor this probably. Themes should be able to strictly specify which other theme the references are coming from, like we do for token sets.
```
cuserox, line 6: Could we create a follow-up to document this logic? It'll help simplify when the time comes.

packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
```diff
@@ -1,6 +1,10 @@
 import { ThemeObject } from '@/types';
 
-export async function mergeVariableReferencesWithLocalVariables(themes: ThemeObject[] = []): Promise<Map<string, string>> {
```
cuserox, line 3: This is also used in [createLocalVariablesWithoutModesInPlugin](https://github.com/tokens-studio/figma-plugin/blob/release-139/packages/tokens-studio-for-figma/src/plugin/createLocalVariablesWithoutModesInPlugin.ts#L91) - did it work, and will it stll work with this change?

packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
```diff
@@ -13,6 +17,15 @@ export async function mergeVariableReferencesWithLocalVariables(themes: ThemeObj
       variables.set(tokenName, variableId);
     });
   });
+  allThemes.forEach((theme) => {
```
cuserox, line 20: Is the fallback in case `allThemes` is empty `[]` going to cause any issues?

cuserox (approved):
Approving with a few caveats. The questions are more for my understanding of whether this change might have consequences on other functionalities of the plugin.

packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
six7: We don't so far, but it would likely be ideal if we would surface it in the `Manage themes` dialog when a token is part of multiple theme groups or is directed at multiple variables, which would then lead to problems like this.

Let's talk about this in our variable sync to figure out a good path forward

packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
six7: Yes! Because of the empty array as default value it won't have an effect. As for did it work: No, creating without modes only takes references of the current document into account, so that variant never had this problem

packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
six7: It shouldn't, also that should never be a case - when all themes are empty we'd not even get to this step. In `packages/tokens-studio-for-figma/src/plugin/createLocalVariablesInPlugin.ts` on line 42 we check if selectedThemes is of length > 0 and only then proceed

packages/tokens-studio-for-figma/src/plugin/mergeVariableReferences.ts
six7: Let's use Thursday's session to start documenting this 👍

Issue #2905, opened 2024-06-22
Bitbucket sync: Create branch doesnt take source branch into account


six7:
Slack: https://tokens-studio.slack.com/archives/C073SSG7DC7/p1718989456482829

Turns out when we create a branch from bitbucket the source branch is not properly used as a reference, but the first one from the list.

PR #2912, opened 2024-06-24
fix bitbucket create branch from having wrong origin

six7:
Fixes https://github.com/tokens-studio/figma-plugin/issues/2905

Users reported bitbucket sync behaves weirdly when creating a new branch. It takes a different source branch than specified.

Looked into the code and we naively took the first branch that came back from the list of branches. Changed it so that we use the specified branch name.

### DIFF
--- a/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
@@ -82,7 +82,7 @@ export class BitbucketTokenStorage extends GitTokenStorage {
       const sourceBranchName = source || this.branch;
       const sourceBranch = originBranch.data.values.find((branchValue: Schema.Branch) => branchValue.name === sourceBranchName);
 
-      if (!originBranch.data.values || !sourceBranch || !sourceBranch.target) {
+      if (!originBranch.data || !originBranch.data.values || !sourceBranch || !sourceBranch.target || !sourceBranch.target.hash) {
         throw new Error('Could not retrieve origin branch');
       }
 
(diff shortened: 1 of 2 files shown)


packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
```diff
@@ -91,7 +91,7 @@ export class BitbucketTokenStorage extends GitTokenStorage {
         _body: {
           name: branch, // branch name
           target: {
-            hash: originBranch.data.values[0].target.hash, // hash of the commit the new branch should point to
+            hash: sourceBranch.target.hash, // hash of the commit the new branch should point to
```
cuserox, line 94: No need to check if there's a `.hash` [above](https://github.com/tokens-studio/figma-plugin/pull/2912/files#diff-dd0df3f2e27a2f6d82e04cd2bbbd4fb4bee00209499c674144bbc8dd2e3f036eR85)?

packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
```diff
@@ -74,15 +74,15 @@ export class BitbucketTokenStorage extends GitTokenStorage {
...
-      const newRef = `refs/${branch}`;
-
       const originBranch = await this.bitbucketClient.repositories.listRefs({
         workspace: this.owner,
         repo_slug: this.repository,
       });
 
-      if (!originBranch.data.values || !originBranch.data.values[0] || !originBranch.data.values[0].target) {
+      const sourceBranchName = source || this.branch;
+      const sourceBranch = originBranch.data.values.find((branchValue: Schema.Branch) => branchValue.name === sourceBranchName);
```
cuserox, line 83: Would this check for [originBranch.data.values](https://github.com/tokens-studio/figma-plugin/pull/2912/files#diff-dd0df3f2e27a2f6d82e04cd2bbbd4fb4bee00209499c674144bbc8dd2e3f036eR85) not be before using `.find(...)` on it? It will fail if `undefined`, or maybe it's an empty array?!

cuserox (approved):
A couple of ❓ around safety fallbacks. Looks good otherwise!

robinhoodie0823 (approved):

PR #2913, opened 2024-06-24
2.0 Polish modal size

### LINKED ISSUE #2909: 2.0: Modals are too big
Some modals are definitely too big in 2.0 - we should adjust this.

Seems like modal size="large" doesnt really have any maximum in mind (ideally we solve this at the DS level and provide patterns and default sizes of modals, see https://primer.style/components/dialog#size

Image

six7:
Closes #2909

Modals in 2.0 are overall too big - they span the full window size.

Introduce a max width for our modals so that they don't go as wide.

### Before

image

### After

CleanShot 2024-06-24 at 08 37 09@2x

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/Modal/Modal.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/Modal/Modal.tsx
@@ -49,7 +49,7 @@ const StyledDialogContent = styled(Dialog.Content, {
     size: {
       large: {
         width: 'calc(100vw - $7)',
-        maxWidth: 'calc(100vw -$7)',
+        maxWidth: '480px',
         padding: 0,
         boxShadow: '$contextMenu',
         borderColor: '$borderSubtle',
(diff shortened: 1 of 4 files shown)


packages/tokens-studio-for-figma/src/app/components/Modal/Modal.tsx
```diff
@@ -49,7 +49,7 @@ const StyledDialogContent = styled(Dialog.Content, {
     size: {
       large: {
         width: 'calc(100vw - $7)',
-        maxWidth: 'calc(100vw -$7)',
+        maxWidth: '480px',
```
cuserox, line 52: Could this number live in [core.ts](https://github.com/tokens-studio/figma-plugin/blob/release-139/packages/tokens-studio-for-figma/src/tokens/core.ts#L14)? And not sure if it'll work (styling dependant) but perhaps a `vw` or `%` could be "safer" to fluidly allow the modal to grow when needed!

cuserox (approved):

packages/tokens-studio-for-figma/src/app/components/Modal/Modal.tsx
six7: I'll create an issue to have that number come from our tokens itself in the ds repo for each `size`. On vw: I think we'd want a px based value as we want to ensure maximum character length and somewhat of a "safe" max width for these - i've created an issue for the `max size` to come from the ds: https://github.com/tokens-studio/ds/issues/221

PR #2914, opened 2024-06-24
2.0: Fix order of keys when pushed to remote

### LINKED ISSUE #2907: 2.0: Order of keys inconsistent when pushed to remote
Its important that we do not change or modify the order of keys that users specified in their JSON. If `type` comes before `value`, we need to respect that.

2.0 changed that at least for Bitbucket or other git sync providers, it moves the `type` key first, always.

Let's respect that instead.

six7:
Closes #2907

With 2.0 we seem to have introduced a regression when it comes to the order of token keys. We made it so that `type` always comes first, then `value`, and so on. This is bad because we need to respect users key ordering.

With this PR we're now respecting the order and keeping it the same by not destructuring the object.

### Before

CleanShot 2024-06-24 at 08 39 54@2x

As you can see in this screenshot we reordered keys, even though the user didnt specify that.

### After

CleanShot 2024-06-24 at 08 40 55@2x

Here's me after the change just adding a modifier - note how it preserves the order of `value` first.

### DIFF
--- a/packages/tokens-studio-for-figma/src/utils/convertTokensToObject.ts
+++ b/packages/tokens-studio-for-figma/src/utils/convertTokensToObject.ts
@@ -10,24 +10,31 @@ export default function convertTokensToObject(tokens: Record<string, AnyTokenLis
     const tokenGroupObj: AnyTokenSet<false> = {};
     val.forEach((token) => {
       const tokenWithType = appendTypeToToken(token);
-      const { name, ...tokenWithoutName } = removeTokenId(tokenWithType, !storeTokenIdInJsonEditor);
+      const tokenWithoutId = removeTokenId(tokenWithType, !storeTokenIdInJsonEditor);
+
+      // Remove the name key to not include it in the output
+      const { name, ...tokenWithoutName } = tokenWithoutId;
+
+      // Directly work with tokenWithoutName to preserve order
       if (tokenWithoutName.inheritTypeLevel) {
-        const {
-          inheritTypeLevel, ...tokenWithoutTypeAndValue
-        } = tokenWithoutName;
-        // set type of group level
-        tokenWithoutTypeAndValue[TokenFormat.tokenValueKey] = tokenWithoutName.value;
-        tokenWithoutTypeAndValue[TokenFormat.tokenDescriptionKey] = tokenWithoutName.description;
-        set(tokenGroupObj, getGroupTypeName(token.name, inheritTypeLevel), tokenWithoutName.type);
-        set(tokenGroupObj, token.name, tokenWithoutTypeAndValue, { merge: true });
+        // If inheritTypeLevel exists, handle it specifically
+        const { inheritTypeLevel } = tokenWithoutName;
+        delete tokenWithoutName.inheritTypeLevel; // Remove it to avoid altering the order
+
+        // Set type of group level without altering the order of tokenWithoutName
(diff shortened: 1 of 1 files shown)


packages/tokens-studio-for-figma/src/utils/convertTokensToObject.ts
```diff
@@ -10,24 +10,31 @@ export default function convertTokensToObject(tokens: Record<string, AnyTokenLis
...
-        const {
-          type, value, description, ...tokenWithoutTypeAndValue
-        } = tokenWithoutName;
-        tokenWithoutTypeAndValue[TokenFormat.tokenTypeKey] = type;
-        tokenWithoutTypeAndValue[TokenFormat.tokenValueKey] = value;
-        tokenWithoutTypeAndValue[TokenFormat.tokenDescriptionKey] = description;
-        set(tokenGroupObj, token.name, tokenWithoutTypeAndValue, { merge: true });
+        // For tokens without inheritTypeLevel, directly add type, value, and description to preserve order
+        tokenWithoutName[TokenFormat.tokenTypeKey] = tokenWithoutName.type;
+        tokenWithoutName[TokenFormat.tokenValueKey] = tokenWithoutName.value;
```
akshay-gupta7, line 34: would it be better to move these two lines outside the if-else considering they are being set in both the conditions

akshay-gupta7 (changes requested):

packages/tokens-studio-for-figma/src/utils/convertTokensToObject.ts
```diff
@@ -10,24 +10,31 @@ export default function convertTokensToObject(tokens: Record<string, AnyTokenLis
...
-          inheritTypeLevel, ...tokenWithoutTypeAndValue
-        } = tokenWithoutName;
-        // set type of group level
-        tokenWithoutTypeAndValue[TokenFormat.tokenValueKey] = tokenWithoutName.value;
-        tokenWithoutTypeAndValue[TokenFormat.tokenDescriptionKey] = tokenWithoutName.description;
-        set(tokenGroupObj, getGroupTypeName(token.name, inheritTypeLevel), tokenWithoutName.type);
-        set(tokenGroupObj, token.name, tokenWithoutTypeAndValue, { merge: true });
+        // If inheritTypeLevel exists, handle it specifically
+        const { inheritTypeLevel } = tokenWithoutName;
+        delete tokenWithoutName.inheritTypeLevel; // Remove it to avoid altering the order
```
cuserox, line 22: Could this `delete` be replaced by a `filter` / `reduce` (or something that doesn't mutate the original object) as it's already looping through it?

six7: fixed in 92732eff8c016a59ce17a89d14eb085e3abcc1c8

Overall: We should be able to get rid of this file - i checked an much of this logic is repeated in stringifyTokens which already does this - I'll add to DX as things to simplify

packages/tokens-studio-for-figma/src/utils/convertTokensToObject.ts
six7: In the state of the PR that you reviewed, yes! howver with 92732eff8c016a59ce17a89d14eb085e3abcc1c8 no more - as the logic now properly diverges

akshay-gupta7: got it!

akshay-gupta7 (approved):

PR #2918, opened 2024-06-25
2915 20 token set tree reorder issues

### LINKED ISSUE #2915: 2.0: Token set tree reorder issues
We've received reports from @SamIam4Hyma (so far only from Sam), that our tree is more broken in 2.0 than it is in production. We should look into if those issues are fixable, or if they require a full-on replacement of our tree (eg with react-arborist as we're using that in Studio). If the problems are not fixable, let's *not* do the full-on replacement just yet, but treat this like a spike of what's required to do so and write down a game plan (because the tree is fairly complex).

The issues happen when you have a deeply nested structure and can be easily reproduced.

- Create a few nested token sets:
  - `global/foo/bar`
  - `global/foo/baz`
  - `primitives/foo/bar`
  - `foo/bar/baz`
  - `global`
  - `primitives`
  - `foo`

Now reorder those sets a few times, like so: 

https://github.com/tokens-studio/figma-plugin/assets/4548309/632921e9-689a-4d07-8842-050e7d359d49

The issue seems to have to do with token sets having the same name as token folders. Which is something that can definitely happen, as you can have a `tokens` folder as well as `tokens.json` file.

It seems like the token set reorder doesnt work as expected (it's expected that it only works on 1 level, but in this case it makes it appear like some token sets are now children of `primitives` when in fact we only changed the ordering.

--- 

Notes for engineering:
- Let's timebox it to 1-2 days. If the issue is not fixable, we'll replace the component in a follow-up during devex.

Entry point for the set tree:
`packages/tokens-studio-for-figma/src/app/components/TokenSetSelector.tsx` (which renders the overall tree and modals)
which renders `packages/tokens-studio-for-figma/src/app/components/TokenSetTree.tsx` (which is responsible for the custom reorder logic we implemented)
which renders a `ReorderGroup` as well as `packages/tokens-studio-for-figma/src/app/components/TokenSetTree/TokenSetTreeContent.tsx` which is responsible for rendering the items

robinhoodie0823:
### Why does this PR exist?
Current plugin is broken when we have multiple nested token set tree structure (when try to collapse/expand, and drag/drop token sets).

Closes #2915  

Added an unique identifier to prevent multi-select and collapse/expand problem. And updated the `tokenSetListToTree` function to prevent merging when drag/drop same name token set (leaf).

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/c1002808-02af-406f-9c06-bf1756ef2bc2


six7 (approved):
🎉

Issue #2902, opened 2024-06-21
Prevent extra brackets`{ }` and whitespaces being added on token name searches (mentions)


cuserox:
**Describe the bug**
Tweaking the extra space when token search is used (https://github.com/tokens-studio/figma-plugin/pull/2789) the way that search results are populated has been slightly broken!

**To Reproduce**
1. Edit a token to reference an alias
2. Select an alias
3. Edit the alias manually by searching
4. Click on any result, see the bracket `{}` added randomly on the input 

**Expected behavior**
No extra bracket nor spaces should be added when a token name is clicked on after searching.

**Screenshots or Screencasts**
https://github.com/tokens-studio/figma-plugin/assets/114073780/335dd67b-6bf2-484f-b528-03733e1a3d50

six7:
@cuserox I have a fix for this available in `six7/fix-mentions`. It turns out a recent change we did to mentions was causing this. Removing this again makes everything work, except that the extra space at the end of the input is added. It's the `split=""` that was the issue.

Perhaps we could adjust this so after applying it strips out the extra space?

For example if you have a token like `{fontsize-small}rem` and you replace fontsize-small with something else, it adds an extra space. But it no longer adds the } after, which is a great improvement but the space is a bit of an issue (which we attempted to solve with that split="" but the sideeffects are worse with that.

cuserox:
> @cuserox I have a fix for this available in `six7/fix-mentions`. It turns out a recent change we did to mentions was causing this. Removing this again makes everything work, except that the extra space at the end of the input is added. It's the `split=""` that was the issue.
> 
> Perhaps we could adjust this so after applying it strips out the extra space?
> 
> For example if you have a token like `{fontsize-small}rem` and you replace fontsize-small with something else, it adds an extra space. But it no longer adds the } after, which is a great improvement but the space is a bit of an issue (which we attempted to solve with that split="" but the sideeffects are worse with that.

@six7 makes sense. All in all, this fix should be for no side effects to occur when a user searches tokens 🔨

cuserox:
@six7 @macintoshhelper @robinhoodie0823 

I have created a [Draft MR here](https://github.com/tokens-studio/figma-plugin/pull/2927), merely because it needs more testing. I believe it works well with my fix, but I wouldn't want to merge it and cause more trouble for V2 (and I'm off this Friday & Monday) 😇 

Therefore if it's a pressing fix that needs to go out, please feel free to tweak, test & merge if needed!

PR #2927, opened 2024-06-27
fix: correctly format suggestions by mentions when submitted

cuserox:
### Why does this PR exist?

Closes [#2902](https://github.com/tokens-studio/figma-plugin/issues/2902)

[A previous fix to prevent trailing whitespace being added](https://github.com/tokens-studio/figma-plugin/pull/2789), had made the [Mentions](https://www.npmjs.com/package/rc-mentions) component misbehave. Inserting extra curly brackets at random spots when searching.

* Removes the `trim` prop in favour of trimming the raw value

### Testing this change

| Before | After |
| ----- | ----- |
| https://github.com/tokens-studio/figma-plugin/assets/114073780/c1bde474-834c-4b70-b6a7-171d04ebd27c | https://github.com/tokens-studio/figma-plugin/assets/114073780/893fae55-d55c-4add-ace7-56ddb4ceced1 |

### Additional Notes (if any)

✨ 👀 Extra improvement: when typing in a search, ensure that if there is a full match, it is highlighted. Perhaps adding a custom `filterOption` prop to the [Mentions](https://www.npmjs.com/package/rc-mentions) component?

At the moment, even if there is a full match (i.e. `colors.blue.500`), the first one on the list is highlighted and submitted when pressing **Enter**:
Screenshot 2024-06-27 at 17 53 57

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/DownshiftInput/MentionInput.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/DownshiftInput/MentionInput.tsx
@@ -80,7 +80,7 @@ export default function MentionsInput({
   }, [initialName, resolvedTokens, referenceTokenTypes, type]);
 
   const handleMentionInputChange = React.useCallback((newValue: string) => {
-    handleChange(name, newValue.replace(/}(?=\s)[^{}]*}/gi, '}'));
+    handleChange(name, newValue.replace(/}(?=\s)[^{}]*}/gi, '}').trim());
   }, [handleChange, name]);
 
   const handleInputBlur = React.useCallback(() => {


packages/tokens-studio-for-figma/src/app/components/DownshiftInput/DownshiftInput.tsx
```diff
@@ -1,30 +1,46 @@
 import React, { useCallback, useMemo } from 'react';
 import Downshift from 'downshift';
 import { useSelector, useDispatch } from 'react-redux';
+
+// Components
```
six7, line 5: is this something that we want to be doing? doesnt feel like something we can automate?

six7 (changes requested):
Tested out the implementation and it has 2 major flaws:
- Cannot type `space` anymore
- The logic automatically adds a space between two token references. This isnt great because users might rely on their values being joined together.

https://github.com/tokens-studio/figma-plugin/assets/4548309/40a30047-4068-42f7-97bf-ee557579cce3

packages/tokens-studio-for-figma/src/app/components/DownshiftInput/DownshiftInput.tsx
cuserox: Could be automated, happy to revert the changes & set as a task for DX

PR #2929, opened 2024-06-28
fixes mentions input re-apply

six7:
Precursor to https://github.com/tokens-studio/figma-plugin/pull/2923 that fixes the most important pieces: Selecting mentions


six7:
Merging in without approval given we'll have that change as part of the larger PR too

Issue #2951, opened 2024-07-02
Color swatch not displaying gradients on search suggestions


cuserox:
**Describe the bug**
When a gradient value has a semicolon `;` it won't get applied to the style in the DOM

**To Reproduce**
1. Create a color gradient
2. Search for it via value or alias
3. See the color swatch is empty

**Expected behavior**
Color swatch should show gradient

**Screenshots or Screencasts**

<img src="https://github.com/tokens-studio/figma-plugin/assets/114073780/bb28ba52-117a-4740-a522-d56c646f19e0" width=200 />
<img src="https://github.com/tokens-studio/figma-plugin/assets/114073780/65386084-d85d-4e22-95a5-7ce102fb1f89" width=200 />

**JSON (optional)**
```
  "just-a-gradient": {
    "$type": "color",
    "$value": "linear-gradient(to right,#dc573c,#247efc);"
  }
```

PR #2952, opened 2024-07-02
fix: color swatches to show colors with gradients

cuserox:
### Why does this PR exist?

Closes [#2951](https://github.com/tokens-studio/figma-plugin/issues/2951#issue-2385828332)

Colors with semicolons were not applying to the color swatches due to an error when the DOM CSS styling was attempting to be applied

* Adds a small utility that strips a semicolon from a string
* Uses the utility for correct styling in search and mentions

### Testing this change
1. Create a color token as a gradient (i.e. `linear-gradient(to right,#dc573c,#247efc);`)
2. Create or edit another color token - search by name or alias
3. See the gradient showing on the swatch

| Before | After |
| ----- | ----- |
| Screenshot 2024-07-02 at 12 04 03 | Screenshot 2024-07-02 at 12 17 05 |
| Screenshot 2024-07-02 at 12 04 14 | Screenshot 2024-07-02 at 12 17 13 |

### DIFF
--- a/packages/tokens-studio-for-figma/src/utils/color/getColorSwatchStyle.ts
+++ b/packages/tokens-studio-for-figma/src/utils/color/getColorSwatchStyle.ts
@@ -1,5 +1,5 @@
 export default function getColorSwatchStyle(tokenValue: string) {
   return tokenValue.includes(';')
     ? { background: tokenValue.replace(/;/g, '') }
-    : { backgroundColor: tokenValue };
+    : { background: tokenValue };
 }
--- a/packages/tokens-studio-for-figma/src/utils/color/getColorSwatchStyle.ts
+++ b/packages/tokens-studio-for-figma/src/utils/color/getColorSwatchStyle.ts
@@ -1,0 +1,5 @@
+export default function getColorSwatchStyle(tokenValue: string) {
+  return tokenValue.includes(';')
+    ? { background: tokenValue.replace(/;/g, '') }
+    : { backgroundColor: tokenValue };
+}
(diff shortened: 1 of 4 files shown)


packages/tokens-studio-for-figma/src/utils/color/getColorSwatchStyle.ts
```diff
@@ -0,0 +1,5 @@
+export default function getColorSwatchStyle(tokenValue: string) {
+  return tokenValue.includes(';')
+    ? { background: tokenValue.replace(/;/g, '') }
```
macintoshhelper, line 3: Should a semicolon be required for `linear-gradient`s? I tested this with a `linear-gradient` without a semicolon and it displays correctly in `InspectorResolvedToken` as that is using `background`, but not in the `DownshiftInput`.

Might be safer to just use `background` and not `backgroundColor`, since there is no validation mandating the use of a semicolon for `linear-gradient`

image

image

cuserox: > Should a semicolon be required for linear-gradients
Not necessarily, but as we are used to writing CSS, it might slip?! As it did with me 🙃 

It didn't work in the DownshiftInput nor search because it was attempting to set a `style` [here](https://github.com/tokens-studio/figma-plugin/blob/fix/2951-color-swatch-gradients-fix/packages/tokens-studio-for-figma/src/app/components/DownshiftInput/DownshiftInput.tsx#L298) and it was causing the DOM to error due to that. In the tokens list, [it already sets it as a CSS variable](https://github.com/tokens-studio/figma-plugin/blob/fix/2951-color-swatch-gradients-fix/packages/tokens-studio-for-figma/src/app/components/TokenButton/TokenButtonContent.tsx#L57)so it works well.

Updated to use `background` 😄

robinhoodie0823 (approved):

PR #2954, opened 2024-07-02
Fix stuck load-bar when token sets are empty whilst creating styles

### LINKED ISSUE #2896: 2.0: Style creation can become stuck if theme only has `references` and no active token sets
When creating styles we end up in a endless loading state when sets used in the themes are not active but disabled or reference only.

To reproduce:
- Create new file
- Create a token
- Create a theme group called `Theme / Default` or anything
- Set the token set to `disabled` or `reference only`
- Proceed to create styles (it's enough to leave one of the styles checkboxes active)

Expected: Do not get stuck on style creation
Actual: Style creation spinner goes on forever

Bonus: Can we adjust the error message if the reason for it was not that no themes were active but that inside there weren't any sets active? `No styles created. Make sure some sets are active`

Image

akshay-gupta7:
Resolves #2896 
Screenshot 2024-07-02 at 22 59 14

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
+++ b/packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
@@ -68,24 +68,37 @@ export default function useTokens() {
   const store = useStore<RootState>();
   const tokensContext = useContext(TokensContext);
   const shouldConfirm = useMemo(() => updateMode === UpdateMode.DOCUMENT, [updateMode]);
-  const VALID_TOKEN_TYPES = [TokenTypes.DIMENSION, TokenTypes.BORDER_RADIUS, TokenTypes.BORDER, TokenTypes.BORDER_WIDTH, TokenTypes.SPACING];
+  const VALID_TOKEN_TYPES = [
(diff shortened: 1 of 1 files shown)


cuserox (approved):
Nice!

packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
```diff
@@ -398,10 +393,15 @@ export default function useTokens() {
...
-      notifyToUI('No styles created. Make sure themes are active.', { error: true });
+      notifyToUI('No styles created. Make sure some sets are active.', { error: true });
       return;
     }
 
+    dispatch.uiState.startJob({
+      name: BackgroundJobs.UI_CREATE_STYLES,
+      isInfinite: true,
+    });
+
```
akshay-gupta7: @robinhoodie0823 , this is the only change I made in useTokens.ts, I switched the order of the check and the start job, and made a change in the text displayed.
The other commit is just a linting update, you can get get those changes by running lint command

PR #2955, opened 2024-07-02
2931 20 defining type at the group level is not being persisted in json

### LINKED ISSUE #2931: 2.0: Defining type at the group level is not being persisted in JSON
We seem to have caused a regression with defining type at the group level. It seems we no longer save it in the JSON.

To reproduce:
- Paste the following JSON in the plugin:

```
{
  "scale": {
    "type": "spacing",
    "small": {
      "value": "4px",
    },
    "medium": {
      "value": "8px",
    },
    "large": {
      "value": "12px",
      "type": "sizing"
    }
  }
}
```

Expected: We keep JSON as is
Actual: We strip away the `type` from the group level and insert a blank key.

This might have something to do with how we stringify the JSON back, and the `inheritTypeLevel` property that we set internally.

Note: This does *not* happen if I am on the W3C DTCG format, maybe that can be helpful in understanding what went wrong here.

https://github.com/tokens-studio/figma-plugin/assets/4548309/b66ddfa4-f364-4f90-81b8-90e3f4606745

robinhoodie0823:
### Why does this PR exist?
Current plugin has problem in case of group name level token types especially in DTCG format tokens.

Closes #2931  

Updated the logic to use correct value of `inheritTypeLevel` in case of DTCG format token.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/e5a3a8bd-77ef-4fa5-9762-1d5a9b0b8a00


packages/tokens-studio-for-figma/src/utils/stringifyTokens.ts
```diff
@@ -28,14 +28,15 @@ export default function stringifyTokens(
       } = tokenWithoutName;
       const tokenInJSON: TokenInJSON = tokenWithoutInheritTypeLevel;
       // set type of group level
-      set(tokenObj, getGroupTypeName(token.name, inheritTypeLevel), tokenInJSON.type);
+      set(tokenObj, getGroupTypeName(token.name, TokenFormat.format === TokenFormatOptions.DTCG ? inheritTypeLevel : inheritTypeLevel + 1), tokenInJSON.type);
```
six7, line 31: can you explain why this is needed? why is the level different between the two formats?

six7 (commented):
there is an issue with this PR.

if you remove the `type` from a token, you end up with `inheritTypeLevel` exposed. we dont want that.


https://github.com/tokens-studio/figma-plugin/assets/4548309/dade95d0-120c-484b-9cbd-067433b41b16

six7 (approved):

PR #2965, opened 2024-07-05
Deduplicate resolve modal fixes

macintoshhelper:
### Why does this PR exist?

#2779 

- Move resolve duplicate warning button
- Fix duplicate resolve modal copy

### Testing this change

- Switch to production version
- Create a group with a conflicting/overlapping token name, but unique group name
- Duplicate this new group to an existing group name, with an overlapping token
- Switch back to this version
- Check the icon button UI and the modal UI

### Additional Notes (if any)

image

image

image


six7 (approved):
LGTM! 🎉

Issue #2908, opened 2024-06-22
2.0: Adjust button styling to Figma's again


six7:
With 2.0 we changed button styling to come from our internal DS. However we deviated from Figma with that. Let's move back here, make primary buttons blue, and secondary buttons outlined.

![Image](https://github.com/tokens-studio/figma-plugin/assets/4548309/940450fd-75e1-4a41-9fb1-90aceec2ad8e)

![Image](https://github.com/tokens-studio/figma-plugin/assets/4548309/5fc95966-3747-46eb-8c63-013883614753)

Issue #2895, opened 2024-06-21
Light color theme is too dark


six7:
The bg of the light color theme is now too dark and doesnt play well with some of our DS components.

can we rather use a lighter bg?

![Image](https://github.com/tokens-studio/figma-plugin/assets/4548309/d339fdb6-aafd-4d31-8360-f63d7954d26c)

PR #2967, opened 2024-07-06
Upgrade UI

six7:
Fixes https://github.com/tokens-studio/figma-plugin/issues/2895
Fixes https://github.com/tokens-studio/figma-plugin/issues/2908

Fixes those issues by upgrading to the latest `tokens` package version


six7:
Merging in without approval given its a package upgrade

PR #2968, opened 2024-07-06
revert new token set button sticky

six7:
makes token set button no longer sticky

image


six7:
Merging because of 1-liner

PR #2982, opened 2024-07-10
2976 variable for a strokes width is not applied for the border token

### LINKED ISSUE #2976: Variable for a Stroke's Width is Not Applied for the Border Token
**Describe the bug**
FIgma has the option to apply number variables for the stroke's width, but the plugin currently does not apply these variables even if the border token is referencing them.

**To Reproduce**
1. Create a border token
2. Create variables for the atomic/particle tokens and the border token
3. Apply the token

**Expected behavior**
The token is also supposed to apply the number variable for the stroke, as defined in the token. Just like the plugin applies the color variable defined in the token.

[Video](https://github.com/tokens-studio/figma-plugin/assets/93133012/40c82013-2258-4a53-8cae-dd22fb410060)

robinhoodie0823:
### Why does this PR exist?
Current plugin does not support reference a existing border width token if user try to apply border token which has borderWidth token as a reference token.

Closes #2976  

Added a `borderWidth` property and pass it to `setValuesOnNode` function to let plugin knows which token it needs to reference on.

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/6f0fbbe2-f371-4a38-a4d0-8fa3be5ab214


cuserox:
Tests for the win!

packages/tokens-studio-for-figma/src/plugin/node.test.ts
```diff
@@ -277,6 +277,58 @@ describe('destructureTokenForAlias', () => {
       expect(destructureTokenForAlias(tokens, value)).toEqual(applyTokens[index]);
     });
   });
+
+  it('return extract border width from border token', () => {
```
six7, line 281:
```suggestion
  it('should retur border width from border token', () => {
```

six7 (approved):

PR #2986, opened 2024-07-10
2984 pull from GitHub indicator remains on when pull was successful

### LINKED ISSUE #2984: Pull from Github indicator remains ON when pull was successful
**Describe the bug**
After pulled from Github, did not get any confirmation that the pull was successful. The indicator persisted.
Even the pull was successful and themes and tokens were visible in the themes config (so the pull was successful).
Pulled several more times to see if the indicator would clear, it did not.
Restarting the plugin cleared the indicator. Originally from [here](https://hyma-team.slack.com/archives/C06RHB9RT2B/p1720554231904009).

**Expected**
The pull indicator should be off after pull was successful.

robinhoodie0823:
### Why does this PR exist?
Current plugin checks if there is updates on remote storage (like github) every 60 seconds. So even I pull tokens from github, their is pull token indicator on and it will be disappeared after another 60 seconds.

Closes #2984  

Updated logic to change `hasRemoteChange` variables just after pull tokens from remote storage.


packages/tokens-studio-for-figma/src/app/components/Footer.tsx
```diff
@@ -73,7 +73,11 @@ export default function Footer() {
   }, [checkRemoteChange]);
 
   const onPushButtonClicked = React.useCallback(() => pushTokens(), [pushTokens]);
-  const onPullButtonClicked = React.useCallback(() => pullTokens({ usedTokenSet, activeTheme }), [pullTokens, usedTokenSet, activeTheme]);
+  const onPullButtonClicked = React.useCallback(() => {
+    pullTokens({ usedTokenSet, activeTheme }).then(() => {
+      setHasRemoteChange(false);
```
six7, line 78: i think we should rather fix the bug inside `pullTokens` directly instead of applying a band-aid.

something inside `pullTokens` causes the blue dot to set it to `hasRemoteChanges`: true.. instead of it seeing the changes as the same

six7 (changes requested):

cuserox:
When this PR gets merged (https://github.com/tokens-studio/figma-plugin/pull/3035) I'd say this should get re-tested. Calculating the difference between local and remote changes was causing lots of issues!

six7:
Closing as #3048 fixes this

Issue #2987, opened 2024-07-10
Theme switching does not work if multiple collections have the same token name as variables


six7:
When you create a setup like this:

global set:
```
{
  "fg": {
    "default": {
      "value": "#000000",
      "type": "color"
    }
  }
}
```

global2 set:
```
{
  "fg": {
    "default": {
      "value": "#ffffff",
      "type": "color"
    }
  }
}
```

and you create two themes like this

![Image](https://github.com/user-attachments/assets/649ad469-de27-47a9-b2b3-e13a98a94b01)

![Image](https://github.com/user-attachments/assets/9b7609ce-13a9-4fef-99ab-a635158d6b64)

![Image](https://github.com/user-attachments/assets/68402cea-11b3-4a7f-b302-94df06e0b4f7)

and you create variables based on that, meaning that both collections have the same token names..

theme switching wont work, as it will not replace the variables that are attached

https://github.com/user-attachments/assets/6b1d4795-e384-42ac-b85d-2697c776cb30

PR #2988, opened 2024-07-10
fix local variable references normalization

six7:
### Why does this PR exist?
Fixes https://github.com/tokens-studio/figma-plugin/issues/2987

Fixes the name normalization as it was performed too late and not before we actually made the comparison.

Also added a test to ensure this doesnt happen

### Testing this change

Create a setup as described or take this .fig and try to theme switch before and after
[theme-switch-bug.fig.zip](https://github.com/user-attachments/files/16168421/theme-switch-bug.fig.zip)

#### Before

https://github.com/user-attachments/assets/49d6650e-d6e1-4863-b405-a115719784df

#### After

https://github.com/user-attachments/assets/6efe54e6-ba64-4b01-92c0-7fda87b5db4f


cuserox (approved):
Nice one for adding the tests as well 😄

PR #2990, opened 2024-07-11
2978 box shadows that reference a color token that is connected to a variable will have their spread as 0 always

### LINKED ISSUE #2978: Box shadows that reference a color token that is connected to a variable will have their `spread` as 0 always
User reported in 

Found issue with tokenized/variable_aliased shadows.
- if i use raw color value on layered shadow, spread work as intended. (pic.1)
- if i use alias/variable color, then spread value always 0 (pic.2).

To reproduce:
Use the following JSON:
```
{
  "shadow": {
    "value": "#00ff00",
    "type": "color"
  },
  "default": {
    "value": [
      {
        "x": "4px",
        "y": "2px",
        "blur": "4px",
        "spread": "8px",
        "color": "{shadow}",
        "type": "innerShadow"
      }
    ],
    "type": "boxShadow"
  }
}
```

- Create variables so that the `shadow` token exists as a variable (E.g. using token sets)
- Apply the shadow

Expected result: Spread is `8`
Actual result: Spread is `0`

Image

It will work as expected if the color does *not* exist as a variable, though.

robinhoodie0823:
### Why does this PR exist?
Current plugin resets `spread` property of shadow token if the shadow token has reference token and variable id of the reference token.

Closes #2978  

The root cause is the figma function `setBoundVariableForEffect` resets `spread` property. You can find out the discussion [here](https://forum.figma.com/t/setboundvariableforeffect-bug/59788).

### Additional Notes (if any)

https://github.com/tokens-studio/figma-plugin/assets/103296157/cd807adf-d76a-4bbd-989d-46ae61e32be0


packages/tokens-studio-for-figma/src/plugin/setEffectValuesOnTarget.ts
```diff
@@ -82,25 +86,31 @@ export default async function setEffectValuesOnTarget(
 ) {
   try {
     const resolvedToken = defaultTokenValueRetriever.get(token);
+    console.log('resolvedToken in setEffectValues: ', resolvedToken);
```
six7, line 89: remove logs

six7 (changes requested):
Except for logs good to merge!

packages/tokens-studio-for-figma/src/plugin/setEffectValuesOnTarget.ts
```diff
@@ -64,7 +64,11 @@ async function tryApplyCompositeVariable({
       if (val.toString().startsWith('{') && val.toString().endsWith('}') && shouldCreateStylesWithVariables) {
         const variableToApply = await defaultTokenValueRetriever.getVariableReference(val.slice(1, -1));
         if (variableToApply) {
-          effect = figma.variables.setBoundVariableForEffect(effect, transformShadowKeyToFigmaVariable(key), variableToApply);
+          const updatedEffect = figma.variables.setBoundVariableForEffect(effect, transformShadowKeyToFigmaVariable(key), variableToApply);
+          effect = {
+            ...effect,
+            boundVariables: updatedEffect.boundVariables
```
cuserox, line 70: Nice fix! 👏  Would you mind placing the discussion link here, with a small explanation as to why it's needed?

In case Figma fix it in the future and we no longer need this override 😇

cuserox (approved):

Issue #3007, opened 2024-07-17
Plugin Misorders Set when Creating New Nested Set 


keeganedwin:
**Describe the bug**
When creating a new set in a nested set, the plugin does not place the newly created set in the correct position/order. In the video, when I create a new set in `light/primitives` named `light/primitives/new/test`, the plugin places `new/test` under at the bottom of the tree.

**To Reproduce**
Steps to reproduce the behavior: Available in the test file

**Expected behavior**
The plugin should place the new sets in the correct position or relative to the position of the main set. 

**Screenshots or Screencasts**
[Video](https://github.com/user-attachments/assets/87adb78e-cb3e-4c6b-8659-e7db3b1d93ae)

**Figma file (optional)**
[Figma file](https://www.figma.com/design/EkiVsXJyKtLTi9oolTpzIm/Untitled?node-id=12-371&t=FcQUl8ei64V6ZDXF-4)

cuserox:
### Previous brainstorming 💡

Discussed with @keeganedwin that this issue was reported by a user in the community, where a duplicated set was not placed in the correctly nested area.

Simplified current behaviour vs expected behaviour:
<img src="https://github.com/user-attachments/assets/aaac884d-2055-47ee-8b52-60aa8273e127" width=300 />

### Next steps 💭 

Adapt the behaviour of the helper that calculates **where** to insert a newly created set. Namely: [findParentIndexToInsertSet.ts](https://github.com/tokens-studio/figma-plugin/blob/main/packages/tokens-studio-for-figma/src/utils/tokenset/findParentIndexToInsertSet.ts).

* Currently: simply finds the sets that have the highest % of intersection on the paths.
* Ideally: it should calculate the intersection with _weights_ on each of paths level. Where base path (parent) has the highest weight, hence making sure that the new set is inserted in the correct area of the hiearchy.

Set the expected behaviour with complex examples and expand the set of unit tests. This way the helper can be updated as per the user's expectations = everyone will be on the same page.

The ones we currently have are test these simple scenarios:

**Test 1**
We want to create `last-set`

Result:
```
base
red-brand
yellow-brand
last-set
```

**Test 2**
We want to create `base/light`

Result:
```
base
base/light
red-brand
yellow-brand
```

**Test 3**
We want to create `red-brand/light/mobile`

Result:
```
base
red-brand
red-brand/light/mobile
yellow-brand
```

**Test 4**
We want to create `yellow-brand/light/mobile/ios`

Result:
[...]

PR #3008, opened 2024-07-18
fix: maintain set ordering on creation

cuserox:
### Why does this PR exist?

Closes [#3007](https://github.com/tokens-studio/figma-plugin/issues/3007#issue-2413515391) 

When creating a new set, even if nested, it was always placed at the end of list.

* Creates a `findParentIndexToInsertSet` helper, which takes in a list of sets and a set name, and returns the index where to insert the new set!
* Uses the helper in the `updateTokenSetsInState` method
* Create and update tests

### Testing this change
* Open the plugin and ensure you have some sets created, [set-ordering.zip](https://github.com/user-attachments/files/16286192/set-ordering.zip), for example:
```
- base
- yellow-brand
- red-brand
```
* Create a new nested set (1 layer) like `base/light` ---> Note how the set is **next** to `base`
* Create a new nested set (2 layers) like `yellow-brand/light/mobile`  ---> Note how the set is **next** to `yellow-brand`
* Create a new set like `final-set` ---> Note how it goes to the end of the list

https://github.com/user-attachments/assets/66359cd5-f7a5-4f9e-8ba6-c6b2a0a78915

### DIFF
--- a/packages/tokens-studio-for-figma/src/utils/tokenset/__tests__/findParentIndexToInsertSet.test.ts
+++ b/packages/tokens-studio-for-figma/src/utils/tokenset/__tests__/findParentIndexToInsertSet.test.ts
@@ -1,0 +1,34 @@
+import type { AnyTokenList } from '@/types/tokens';
+import { findParentIndexToInsertSet } from '../findParentIndexToInsertSet';
+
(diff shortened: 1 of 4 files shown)


six7 (approved):
🎉  Awesome!

Issue #2997, opened 2024-07-16
Pulling from BitBucket overwrites $metadate/$themes with empty data


rbosker:
**Describe the bug**
In the sync provider BitBucket you can push everything fine, but when you pull changes with themes, these themes are overwritten on pull with empty data and no themes are present anymore.

**To Reproduce**
1. Create a tokenset with themes and sync those to BitBucket
2. Check BitBucket to see that the themes data is available on the remote.
3. Pull the tokens again > you will get a message that configuration has changed.
4. Check themes and see that there are no themes.

**Screenshots or Screencasts**
![image](https://github.com/user-attachments/assets/61f71b3f-8db1-48a7-9304-1afa24383ab5)

**Note**
Beta RC9 > this happens with tokens created in the Beta as well as v1.38.9

rbosker:
hey @cuserox , could you take a look at this issue? Let me know if you have questions about it

akshay-gupta7:
The issue was only occurring when tokens were being pulled from single tokens file sync.
Is addressed in the following Pull Request:

PR #3013, opened 2024-07-22
Fix bitbucket sync themes when pulling from a single json file

### LINKED ISSUE #2977: Using Opacity in Tokens Creates Blank Variable [2]
**Describe the bug**
This is a recurring bug that has been reported previously in:
1. #2549 
2. #2015 

When you create a variable of a token that's using the alpha operation as a color modifier, it creates a blank #ffffff variable.

**To Reproduce**
1. Create a color token also using an opacity token as reference
2. Create variables
3. See error
4. Go back to the the token
5. Edit token
6. Click on Save
7. Error is resolved - variable is updated

**Expected behavior**
The user should not have to edit the token and then save it for it to update the variable to its correct value

[Recording](https://github.com/tokens-studio/figma-plugin/assets/93133012/77ab5c7c-440a-4f28-b6de-46266031225a)

**JSON (optional)**
Reach out to me for token-set

akshay-gupta7:
Resolves https://github.com/tokens-studio/figma-plugin/issues/2997

### DIFF
--- a/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
@@ -80,9 +80,17 @@ export class BitbucketTokenStorage extends GitTokenStorage {
       });
 
       const sourceBranchName = source || this.branch;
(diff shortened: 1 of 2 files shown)


six7 (approved):
🎉

six7:
@akshay-gupta7 tests are failing here

six7 (approved):

Issue #3017, opened 2024-07-23
Converting to W3C format causes two values in new and old format on push


rbosker:
**Describe the bug**
When you convert your tokens to W3C and push those tokens to a syncing provider, the json shows the tokens in the old and new format.

**To Reproduce**
1. Open tokenset
2. Convert tokens to W3C format
3. Push to syncing provider (tested with GitHub, GitLab & BitBucket).
4. See the tokens on the remote that have double values in the new and old format.
5. The JSON in the plugin shows the JSON correctly, but on the remote it's wrong.

**Note**
Reproducable in rc9 & rc10

https://github.com/user-attachments/assets/f93869cb-dd8e-483b-a458-8431dad0122f

rbosker:
@cuserox could you take a look at this issue? Seems to be an important issue, since it messes up the formating of tokens.

cuserox:
🤖 PR is being reviewed: https://github.com/tokens-studio/figma-plugin/pull/3022

🔨 Build of the fix: [3017-fix-double-format-keys.zip](https://github.com/user-attachments/files/16378735/3017-fix-double-format-keys.zip)

👀 @rbosker @keeganedwin @akshay-gupta7 Could one of you please test this works as expected with other providers? I only tested with GitHub.

rbosker:
tested this for BitBucket, GitLab and ADO > all is working as expected 🥳

cuserox:
🤔 Question: if the user was experiencing duplicate keys (legacy & DTCG) on the remote, this should sanitize and fix it, correct?

If so, it will be worth adding it to troubleshooting docs and making it visible to those encountering it.

six7:
Closing as PR is merged

PR #3022, opened 2024-07-23
fix: prevents duplicated format sensitive keys on tokens when pushing to remotes

cuserox:
### Why does this PR exist?

Closes [#3017](https://github.com/tokens-studio/figma-plugin/issues/3017#issue-2424807255)

Switching between legacy and W3C DTCG format would wrongly push duplicated keys on the remotes 
(i.e. `type: 'color'` and `$type: 'color'`) causing issues on tokens when using them.

* Creates a helper called `setTokenKey` which checks the format being used before adding | removing keys and values on the token
* Uses this helper in `convertTokensToObject`, which is the core method used in **RemoteTokenStorage** before pushing to remotes!
* Adapts and adds test coverage to these methods

### Testing this change

📖 🔨 Test file: [colors.json](https://github.com/user-attachments/files/16374078/colors.json)
☝️ Set up alongside a remote providers (this was tested with Github)

#### Non-inherited types

1. Note how these are in **legacy** format
2. Switch to **W3C DTCG** format, don't push yet
3. Go back to the Tokens tab and see the token & the updated JSON view (`$value`, `$description`, `$type`)
4. Push to your remote
5. See no duplicated keys, and the correct  **W3C DTCG** format ✅ 
6. 🔁 Do the same switching back to **legacy** format ✅ 

https://github.com/user-attachments/assets/900ee853-826d-443c-a686-a9c78ebd38b4

#### Inherited types

1. Remove the `"type": "color"` from any group
2. Note how it is shifted to the bottom of the group
3. Push to your remote
4. Repeat steps **2 - 6** as above 👀 

https://github.com/user-attachments/assets/ea37b8a7-639a-4fe7-911e-e0a95dc4c5d0

### Additional comments
⏳ Nice to have in the future: the `Diff` tab (before pushing) is empty when switching formats, as it doesn't compare **keys** (`type` vs `$type**). Should have a message that shows the user that the formatting has been updated, somehow? 
Screenshot 2024-07-23 at 19 34 22

### DIFF
--- a/packages/tokens-studio-for-figma/src/utils/convertTokensToObject.test.ts
+++ b/packages/tokens-studio-for-figma/src/utils/convertTokensToObject.test.ts
@@ -1,54 +1,205 @@
 import convertTokensToObject from './convertTokensToObject';
+import { AnyTokenList } from '@/types/tokens';
+import { TokenFormatOptions, TokenFormat } from '@/plugin/TokenFormatStoreClass';
+
 import input from './test/github/input.json';
 import output from './test/github/output.json';
 
-describe('convertTokensToObject', () => {
-  it('converts array-like tokens to nested object', () => {
-    const tokens = {
-      options: [
-        {
-          name: 'colors.red',
-          type: 'color',
-          value: '#ff0000',
-        },
-        {
-          name: 'colors.blue',
-          value: '#0000ff',
-        },
-      ],
-      theme: [
-        {
-          name: 'colors.primary',
-          value: '$colors.red',
-        },
-      ],
-    };
+const INPUT_TOKENS = {
+  base: [
+    {
+      name: 'colors.red',
+      type: 'color',
+      value: '#ff0000',
+    },
+    {
+      name: 'colors.blue',
+      type: 'color',
+      value: '#0000ff',
+    },
+  ],
+  extras: [
+    {
+      name: 'dimensions.scale',
+      type: 'dimension',
+      value: '2',
+      description: 'We use this to scale things',
+    },
+    {
+      name: 'dimensions.xs',
+      type: 'dimension',
+      value: '4',
+    },
+  ],
+} as Record<string, AnyTokenList>;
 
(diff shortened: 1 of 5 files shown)


packages/tokens-studio-for-figma/src/utils/convertTokensToObject.test.ts
```diff
@@ -1,54 +1,217 @@
...
-        },
-      },
-      theme: {
-        colors: {
-          primary: {
-            value: '$colors.red',
-            type: 'color',
-          },
-        },
+// TODO: check if the 'description' key always gets added to the output in production
```
six7, line 8: i dont think it does 🤔

packages/tokens-studio-for-figma/src/utils/convertTokensToObject.test.ts
```diff
@@ -1,54 +1,217 @@
...
+  ],
+} as Record<string, AnyTokenList>;
+
+const OUTPUT_LEGACY_FORMAT_OBJECT = {
+  base: {
+    colors: {
+      red: {
+        type: 'color',
+        value: '#ff0000',
+        description: undefined,
```
six7, line 74: that feels like something we might want to strip away during the conversion if its undefined

packages/tokens-studio-for-figma/src/utils/convertTokensToObject.test.ts
cuserox: You are right, will not add it if it doesn't exist 🙏

packages/tokens-studio-for-figma/src/utils/convertTokensToObject.test.ts
cuserox: Amended as per my [comment above](https://github.com/tokens-studio/figma-plugin/pull/3022#discussion_r1691133023). Thanks 😄

six7 (approved):
Really well done. Love the tests and the structure ❤️

Issue #3020, opened 2024-07-23
Renaming tokens and token groups results in the rename variable checkbox being positioned differently.


NuthanHyma:
**Describe the bug**
Renaming tokens and token groups results in the rename variable checkbox being positioned differently in the dialog.

**To Reproduce**
1. Renaming any token, note the position of the Rename variables check box, like in IMG.1
2. Rename any token group and see the Rename variables check box changes the position, ike in IMG.2

**Expected behavior**
 when renaming tokens and token groups, the rename variable checkbox should not be positioned differently.
**Screenshots or Screencasts**
IMG.1
<img width="800" alt="Screenshot 2024-07-23 at 7 02 35 PM" src="https://github.com/user-attachments/assets/de155fa3-56d0-4ce4-be45-e72f1f47af2f">

IMG.2
<img width="803" alt="Screenshot 2024-07-23 at 7 02 18 PM" src="https://github.com/user-attachments/assets/e43d7d50-ca94-42b3-848e-226a16275d7a">

Issue #3016, opened 2024-07-23
UI of Property Dropdown in Composition Tokens is Cut-off [rc9]


keeganedwin:
**Describe the bug**
The UI of the property dropdown is cut-off for composition tokens in the `rc9` build. 

**Screenshots or Screencasts**
![screenshot](https://github.com/user-attachments/assets/6e8058f2-e284-4337-ac55-e04b92bfddfa)

PR #3023, opened 2024-07-24
fix variable rename pos and composition form

six7:
Fixes https://github.com/tokens-studio/figma-plugin/issues/3020
Fixes https://github.com/tokens-studio/figma-plugin/issues/3016

Includes formatting changes as well (automatically on save)

CleanShot 2024-07-24 at 08 42 45@2x

also includes a fix for composition token forms (by making sure the longest string is readable)

image


akshay-gupta7 (approved):
🎉

PR #3024, opened 2024-07-24
Fix resolved tokens preview in typo composition

### LINKED ISSUE #3019: Typography/Composition tokens don't give a good preview in the plugin, when using the W3C format
**Describe the bug**
The preview in the plugin shows a list of values, but not the same (good) view when using the old format.
The screenshots below show the W3C and legacy preview, where the legacy preview looks a lot better (with the resolved values).

W3C format
image

Legacy format
image

akshay-gupta7:
Resolves #3019

### DIFF
--- a/packages/tokens-studio-for-figma/src/utils/is/isSingleBoxShadowToken.ts
+++ b/packages/tokens-studio-for-figma/src/utils/is/isSingleBoxShadowToken.ts
@@ -3,24 +3,27 @@ import { TokenFormat } from '@/plugin/TokenFormatStoreClass';
 import { SingleToken, SingleBoxShadowToken, SingleBoxShadowTokenInJSON } from '@/types/tokens';
 import { TokenInJSON, Tokens } from '../convertTokens';
 
+const tokenTypeKey = 'type';
+const tokenValueKey = 'value';
+
 export function isSingleBoxShadowToken(token: SingleToken | any): token is SingleBoxShadowToken {
   if (typeof token !== 'object') return false;
   return (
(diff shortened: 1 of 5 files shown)


packages/tokens-studio-for-figma/src/utils/is/isSingleBoxShadowToken.ts
```diff
@@ -1,16 +1,30 @@
...
-export function isSingleBoxShadowToken(token: SingleToken | any): token is SingleBoxShadowToken {
+function getTokenTypeKey(ignoreTokenFormat: boolean): string {
+  const key = TokenFormat.tokenTypeKey;
+  return ignoreTokenFormat && key.startsWith('$') ? key.slice(1) : key;
+}
+
+function getTokenValueKey(ignoreTokenFormat: boolean): string {
+  const key = TokenFormat.tokenValueKey;
+  return ignoreTokenFormat && key.startsWith('$') ? key.slice(1) : key;
+}
```
six7, line 14: You repeat this in all files - instead of duplicating logic, we should try to abstract the function.

Could you do that as a follow-up ? Approving this PR with the hope that this follow-up can be done 🙏

six7 (approved):
one comment that would be good to resolve as a followup

packages/tokens-studio-for-figma/src/utils/is/isSingleBoxShadowToken.ts
akshay-gupta7: Raised a PR to address this as a follow-up #3034

PR #3026, opened 2024-07-24
Fix resolve duplicate modal crash with object values

### LINKED ISSUE #3025: Plugin crash with non-string token values in resolve duplicate modal
**Describe the bug**

Plugin crashes when opening the duplicate resolve modal and there are duplicate tokens with typography/composition values.

**Expected behavior**

- Shared logic with `TokenTooltipContentValue`
- Later on this UI should be decoupled from being tooltip specific.

image

**Figma file (optional)**

**JSON (optional)**

macintoshhelper:
### Why does this PR exist?

Fixes #3025 

### Testing this change

- Create duplicate typography, composite and box shadow token values by switching to the production version and use `Duplicate Group`
- Open the Resolve Duplicates Modal and verify that the previews show correctly and the plugin doesn't crash

### Additional Notes (if any)

image

image


six7 (approved):
🎉

PR #3029, opened 2024-07-25
chore: add action to post link to built artifact

cuserox:
### Why does this PR exist?

Make it easier to share individual builds to test.

* Adds an action [after the artifact is stored](https://github.com/tokens-studio/figma-plugin/blob/release-139/.github/workflows/node.js.yml#L32) that posts a link to download it. Runs on every CI once a PR is created.
* Uses `actions/upload-artifact@v4` to have access to the `artifact-url` it outputs. Older versions will also be deprecated soon 👉 [read more here](https://github.com/actions/upload-artifact?tab=readme-ov-file#actionsupload-artifact)

### Testing this change

[See automated comment below](https://github.com/tokens-studio/figma-plugin/pull/3029#issuecomment-2250833012) 😄

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -31,12 +31,30 @@ jobs:
         run: npm run build
       # 6: Store artifact
       - name: Store Artifact
-        uses: actions/upload-artifact@v2
+        id: "store_artifact"
+        uses: actions/upload-artifact@v4
         with:
           name: package
           path: |
             packages/tokens-studio-for-figma/dist
             packages/tokens-studio-for-figma/manifest.json
+
+      # 7: Post link to artifact
+      - name: Find artifact comment
+        uses: peter-evans/find-comment@v3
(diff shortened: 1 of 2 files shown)


macintoshhelper:
Awesome! 🎉 

Would be nice to avoid creating a new comment for every commit (and avoid notification after the first comment). Could use something like this: https://github.com/marketplace/actions/create-or-update-comment 🙂

cuserox:
> Awesome! 🎉
> 
> Would be nice to avoid creating a new comment for every commit (and avoid notification after the first comment). Could use something like this: https://github.com/marketplace/actions/create-or-update-comment 🙂

Nice, I'll adapt it to post on the first run, then update it on subsequent runs 👀 New action discovered. Thanks!

macintoshhelper (approved):
LGTM! 🎉

six7:
🎉  This is _AWESOME_! Thank you!!

PR #3032, opened 2024-07-26
Fix composition tokens not applying raw values for colours

### LINKED ISSUE #3014: Composition Token Does Not Apply Raw Values [rc9]
**Describe the bug**
Currently testing this in `rc9`, but composition tokens with raw values do not apply, regardless of the property; this does work as expected in `38.9`

**To Reproduce**
1. Create a composition token with `borderColor: #FFF000`
2. Apply to selection
3. See it does not get applied

✅ If using `borderColor: {comp-yellow}` - where `comp-yellow: #FFF000` - it gets applied

**Expected behavior**
Composition tokens should also just be able to apply raw values just like in `38.9`, and not just work with referenced values. 

[Test Video](https://github.com/user-attachments/assets/6409e232-c137-4216-8e21-e7aa34154ed7)

macintoshhelper:
### Why does this PR exist?

Follow-up PR

Fixes #3014  

### Testing this change

- Try applying this composition token to a node and verify that it applies and doesn't error:

```
{
  "test": {
    "value": {
      "borderColor": "#FFF000",
      "sizing": "{medium} + 40",
      "fill": "#0fa",
      "borderRadiusTopLeft": "16 + 8",
      "backgroundBlur": "12px"
    },
    "type": "composition"
  }
}
```

### Additional Notes (if any)

image

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/setColorValuesOnTarget.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/setColorValuesOnTarget.ts
@@ -32,7 +32,7 @@ export default async function setColorValuesOnTarget({
       existingPaint = target.strokes[0] ?? null;
     }
 
-    if (resolvedValue.startsWith('linear-gradient')) {
+    if (resolvedValue.startsWith?.('linear-gradient')) {
       const fallbackValue = defaultTokenValueRetriever.get(token)?.value;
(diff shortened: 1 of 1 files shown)


macintoshhelper:
Test coverage seems to be failing because of one line difference, unrelated to this PR; I attempted to fix it with little luck, probably not worth sinking more time into getting it passing.

cuserox (approved):
Tried & tested. Good find & fix 😄

PR #3033, opened 2024-07-26
fix: push dialog toggles correctly

cuserox:
### Why does this PR exist?

Toggling tabs in the "Push to remote X" dialog would not set any active tab when an option is clicked more than once.

* Removes unused `FormatSelector` component
* Sets the active tab on the state correctly for `PushDialog`
* Saves the available tabs as an `enum` `PushDialog`
* Fixes other `ToggleGroup`s that were not setting the value correctly on change

### Testing this change
* Open the plugin 
* Make changes
* Push to a remote
* Click on any tabs repeatedly and see the view stays in place

| Before    | After |
| -------- | ------- |
| <video src="https://github.com/user-attachments/assets/f4a24d75-db45-4899-a5ce-d6ee70358084" width="300px" />  | <video src="https://github.com/user-attachments/assets/6b67a07f-ae96-4643-8448-cd50ad579c60" width="300px" /> |


macintoshhelper (approved):
LGTM

PR #3037, opened 2024-07-30
fix: revert and fix attach when creating styles

six7:
### Why does this PR exist?

Fixes the issue our users have been having with RC11 on https://tokens-studio.slack.com/archives/C073SSG7DC7/p1722027287469299   if they were exporting to Styles we would ignore certain token sets. This was due to the new behavior we introduced in RC11 that made it so that when you create styles we iterate over each theme and look at each own on its own.

Its obvious our users arent setup for this as they have themes that are using `disabled` token sets and not as would be correct set as source.

This PR reverts the behavior while keeping the behavior that the original issue there was trying to solve (attaching styles wasnt working anymore)

### Testing this change

Create a few themes and then create styles.


cuserox (approved):
Thanks for updating the tests for safety!

PR #3041, opened 2024-07-31
Fix border sides application and removal

### LINKED ISSUE #1829: Border token can't be applied to bottom, left and right separate
**Describe the bug**
When you apply a border to the left, bottom or right that doesn't work. If you apply border top, this will work.

Once a border top is applied, all others (left, right, bottom) also work (See video).

https://user-images.githubusercontent.com/109062656/234510975-ae28f13c-eb12-4a88-bf28-24225b18fbde.mov

macintoshhelper:
### Why does this PR exist?

Fixes #1829 

### Testing this change

- Apply a border to the top/left/right/bottom
- The change should appear
- Remove an individual border side value and the other border sides should be maintained
- Video: https://github.com/tokens-studio/figma-plugin/issues/1829


packages/tokens-studio-for-figma/src/plugin/removeValuesFromNode.ts
cuserox, line 1: Should `setValuesOnNode` have some similar checks?

It seems when a single side is chosen, and applied gets stuck afterwards. Cannot select anything. See video. 

https://github.com/user-attachments/assets/e055c5cf-9baa-4221-a3a4-7d5fd609e3c5

cuserox:
I believe there's another bug?
* Create a border token
* Set **all** borders ✅ 
* Unset by clicking on the token pill ✅ 
* Set **right** & **left** borders ✅ 
* Unset by clicking on the token pill ✅ 
* Set by clicking on the token pill ---> all borders get applied 🤔 

Do the token settings get forgotten, somehow?

https://github.com/user-attachments/assets/7da7f9fc-5122-47ea-9883-c998d424b059

macintoshhelper:
> Re unchecking `All` also resetting the other side borders

Seems that there is the same behaviour in production, so will create an issue for this as a future fix.

> Re unselecting a token on a node, then selecting/clicking it again, not remembering the previous side values

Same behaviour in production, should probably be a feature request or discussion around changing the behaviour of clicking a border token button to not be the same as clicking `All`

cuserox (approved):

PR #3048, opened 2024-08-06
fix: store last synced state from remote

### LINKED ISSUE #2984: Pull from Github indicator remains ON when pull was successful

cuserox:
### Why does this PR exist?

Closes #2984 

Ensures that every time remote changes are made, the provider data is saved in the local state as `lastSyncedState`. Treat this as a follow-up from the merged PR https://github.com/tokens-studio/figma-plugin/pull/3035

Dispatches data to set `setLastSyncedState` (the same as `setRemoteData`) so that `tokenState` is up to date with the freshest remote status.

### Testing this change
* Open the plugin
* Make token changes
* Push changes
* Continue to use the plugin and see no 'notification' bubble shows to push anything!

https://github.com/user-attachments/assets/6568d740-1039-4d59-b546-842ca90f1454

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/store/providers/ado/ado.tsx
+++ b/packages/tokens-studio-for-figma/src/app/store/providers/ado/ado.tsx
@@ -1,6 +1,7 @@
 import { useDispatch, useSelector } from 'react-redux';
 import React from 'react';
 import { LDProps } from 'launchdarkly-react-client-sdk/lib/withLDConsumer';
+import compact from 'just-compact';
 import { Dispatch } from '@/app/store';
 import useConfirm from '@/app/hooks/useConfirm';
 import usePushDialog from '@/app/hooks/usePushDialog';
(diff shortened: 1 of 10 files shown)


cuserox:
> The diff is showing all the correct info!
> 
> Only found one (minor) bug: when you activate a syncing provider > pull the tokens from remote > once pulled, you do see a blue dot that indicates that you can push. However, when you check the diff, there is nothing there. This is the case for all syncing providers except for GitHub & GitLab.
> 
>  Screen.Recording.2024-08-06.at.17.15.35.mov

Thanks for checking this @rbosker 🙏 I believe the issue you found is not fixed for BitBucket and ADO. Could you please check if you get the chance?

Attached a video of my local test with BitBucket 🙌 

https://github.com/user-attachments/assets/ee2723b9-109f-4dc0-9bb8-918c1e8a63c2

six7 (approved):
🎉

PR #3056, opened 2024-08-08
fix endless loader

### LINKED ISSUE #3054: Loader on "creating variables" Doesn't Cease Even After Variables Are Created
**Describe the bug**
We haven't heard many users report this issue, so far just two. Once you export variables, the loader on the top indicating the process of exporting `'Creating Variables'` doesn't cease even after variables have been successfully exported. I have left the plugin to run for a couple of hours and the loader was still present until I clicked on `'Cancel'`. 

**Expected behavior**
Loader should stop after variables have been exported. 

[Recording](https://github.com/user-attachments/assets/f4b727e4-3cf1-46ca-8694-0c56905332b4)

**Figma file (optional)**
Don't want to expose the user's tokens publicly. Please reach out to me.

six7:
### Why does this PR exist?

Closes #3054

Users were facing an endless `Creating variables…` loader that didnt go away

Adds a try { } catch { } block to the `updateVariableToReferences` function that safely proceeds creating variables if we encounter an error trying to set a variable. The reason for this was that sometimes we were passing in the wrong type for a variable that expected something else.

This PR also adds checks to make sure that we do not attempt to set the value if it is `NaN` and logs a warning message if this is the case. As well as adding a check to see if the value is still a raw value.

### Testing this change

Try to recreate with the token sets shared in https://hyma-team.slack.com/archives/C06RHB9RT2B/p1723030872522979

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/ManageStylesAndVariables/ExportSetsTab.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/ManageStylesAndVariables/ExportSetsTab.tsx
@@ -145,7 +145,6 @@ export default function ExportSetsTab({ selectedSets, setSelectedSets }: { selec
       <Modal
         size="fullscreen"
         full
-        compact
         isOpen={showChangeSets}
         close={handleCancelChangeSets}
         backArrow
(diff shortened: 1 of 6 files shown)


cuserox (approved):
Did a small change on the try / catch block, to throw the number error inside the block: [here](https://github.com/tokens-studio/figma-plugin/pull/3056/files#diff-70f7c15694c8c73801a0052093244ef4ec68083eef66e15b41b08b30fd5463df).

Other than that, tested and looks good!

six7:
> Did a small change on the try / catch block, to throw the number error inside the block: [here](https://github.com/tokens-studio/figma-plugin/pull/3056/files#diff-70f7c15694c8c73801a0052093244ef4ec68083eef66e15b41b08b30fd5463df).

thank you!! ❤️

PR #3072, opened 2024-08-15
Fix bitbucket folder sync pull

### LINKED ISSUE #3070: BitBucket: push/pull for sets using folders not working properly
**Describe the bug**
push/pull for sets using folders not working properly

**To Reproduce**
1. Create a folder of a token set
2. Push those to Bitbucket (changes will appear) > (blue dot appears, but probably related to the actual issue here)
3. After pulling the tokens, it deletes this folder, while it's still present on the remote.

**Acceptance**
- Properly push changes and remove the blue dot after the change is pushed
- On pull: keep the changes that are present on the remote and reflect those in the plugin

https://github.com/user-attachments/assets/42565883-6282-48ed-a166-f3c19efce6bc

**Note**
This occurs on multifile and single file sync.

akshay-gupta7:
### Why does this PR exist?

Resolves #3070 

In the 2.0 version of the plugin, the users are unable to fetch sets from Bitbucket remote storage, when the sets are in a folder structure. For example if a set exists as folder1/set1, it shall be unable to fetch it and even delete the existing folder sets.

- This PR scans through the response of the Bitbucket API and filters the directories as well(which was earlier being ignored), and then fetches the sets in json format from each of those folders.
- It then scans through all the json files(which was being done previously as well, but now includes the folder structure sets as well)

### Testing this change

- Create a set in a folder,
- Add some tokens in the set
- Push to Bitbucket
- Then pull from Bitbucket
- Expected: It should not wipe the folder sets and pull the sets in the folder

### DIFF
--- a/.changeset/mighty-cooks-punch.md
+++ b/.changeset/mighty-cooks-punch.md
@@ -2,4 +2,4 @@
 "@tokens-studio/figma-plugin": patch
 ---
 
-Fix issue with Bitbucket sync wiping sets when they are in a folder format
+Fixed an issue with Bitbucket sync that caused sets to be wiped when they are in a folder format
--- a/.changeset/mighty-cooks-punch.md
+++ b/.changeset/mighty-cooks-punch.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fix issue with Bitbucket sync wiping sets when they are in a folder format
--- a/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
@@ -217,7 +217,7 @@ export class BitbucketTokenStorage extends GitTokenStorage {
             data: parsed as AnyTokenSet<false>,
           };
         });
-      } else if(jsonFiles) {
+      } else if (jsonFiles) {
         const parsed = jsonFiles as GitSingleFileObject;
         return [
           {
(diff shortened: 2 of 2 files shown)


packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
```diff
@@ -232,9 +248,6 @@ export class BitbucketTokenStorage extends GitTokenStorage {
           })),
         ];
       }
-      return {
-        errorMessage: ErrorMessages.VALIDATION_ERROR,
-      };
```
six7, line 237: will this have an effect that we no longer return the errormessage?

.changeset/mighty-cooks-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fix issue with Bitbucket sync wiping sets when they are in a folder format
```
six7, line 5:
```suggestion
Fixed an issue with Bitbucket sync that caused sets to be wiped when they are in a folder format
```

six7 (approved):
super minor comments, otherwise good to merge!

packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
akshay-gupta7: the error message in the catch statement will now display, not a custom one, should I reinstate this one?

six7 (approved):

Issue #3085, opened 2024-08-19
Fix tests related to ADO


six7:
We recently commented out tests for ADO to pass release, but we should work on having them pass again.

PR #3096, opened 2024-08-21
fix: ADO single and multi file tests

cuserox:
### Why does this PR exist?

Closes [#3085](https://github.com/tokens-studio/figma-plugin/issues/3085)

Changes to the way that ADO works meant that the fetch mocks had to be updated to test the functionality correctly

### Testing this change

Run the tests (or see them in the pipeline) - see that `ADOTokenStorageTest.ts` ✅

### DIFF
--- a/.changeset/shaggy-sheep-know.md
+++ b/.changeset/shaggy-sheep-know.md
@@ -1,5 +1,0 @@
----
-"@tokens-studio/figma-plugin": patch
----
-
-Adapts unit tests to reflect ADO writing in single and multi file setups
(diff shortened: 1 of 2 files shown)


.changeset/shaggy-sheep-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Adapts unit tests to reflect ADO writing in single and multi file setups
```
macintoshhelper, line 5: Not sure if a changelog is necessary for non-user facing changes?

macintoshhelper (approved):
LGTM. One comment re changeset for tests, not sure if this would be (or whether it should be) surfaced for the `2.0.2` changelog

.changeset/shaggy-sheep-know.md
cuserox: Removing

PR #3107, opened 2024-08-30
Fix padding in export sets and add gap between sets

akshay-gupta7:
Small fix to a missing padding issue in Export Sets Tab

Before:
Screenshot 2024-08-30 at 18 53 51

After:
Screenshot 2024-08-30 at 18 51 40


cuserox (approved):
Making the modal into its `compact` variant will always have the padding on. Hopefully to stay this time 😄

Issue #2331, opened 2023-10-19
Support for Determining a Gradient's Anchor Points


keeganedwin:
**Is your feature request related to a problem? Please describe.**
We currently don't support the ability to determine the anchor points of gradient tokens (probably also Figma API related). 

One user's used case was that when he imported gradient styles into the plugin, the anchor points when applied through token were inaccurate compared to the way the style counterparts were setup and this caused for inconsistency across brands. 

We will need more information on this as we also weren't able to find much information of this being supported in CSS. That being said, the aforementioned user did share the following links emphasising on the support of anchor-points:

1. https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient
2. https://developer.mozilla.org/en-US/play

**Screen-recording:**
![2023-10-16 14 58 56](https://github.com/tokens-studio/figma-plugin/assets/93133012/a12f4e69-db8d-46e4-b58f-a1ab436798ff)

qimmiq:
we are using this method since quite some time. 
And it always worked, i don't know if Figma changed sth, or it is something else.

guedzzz:
I think you described the issue precisely. The only thing I can add to this is the fact that the anchor points percentages are imported by Tokens Studio, but when applied back to a shape they are ignored. So, my guess is that this is indeed a bug in TS. Thanks for opening this ticket.

qimmiq:
@keeganedwin so this is definitly a bug, because the sync/applying back works when the angle of the gradient is in 45° steps. it only brakes when other angles are used

keeganedwin:
@qimmiq @guedzzz, I came across an old ticket that was fixed a long time ago. It also seems to be related to importing gradient styles into TS which then changes the degree: #156

qimmiq:
@keeganedwin @guedzzz  this ticket describes the way figma -> tokenstudio ... 
what we are pointing out is the the other way. tokenstudio->figma. applying a color token to a figma shape breaks the appearance.

qimmiq:
@keeganedwin i would also rather label it as a bug, because the position points of the gradient
Figma CSS:
background: linear-gradient(153.43deg, #AD00FF **0%**, rgba(82, 255, 0, 0) **83.33%**)
copied to Token applied back to shape, copied Figma CSS
background: linear-gradient(153.43deg, #AD00FF **8.33%**, rgba(82, 255, 0, 0) **77.78%**)

and also creating styles from tokens result in empty fills. 
[...]

PR #3109, opened 2024-08-30
Fix custom linear gradient angles

macintoshhelper:
### Why does this PR exist?

Partial fix for #2331 

**Before**

image

**After**

image


packages/tokens-studio-for-figma/src/plugin/figmaTransforms/gradients.ts
```diff
@@ -63,39 +63,41 @@ export function convertStringToFigmaGradient(value: string) {
...
+  const sin = Math.sin(rad);
+  const cos = Math.cos(rad);
+
+  let scale = 1;
+
+  if (['RECTANGLE', 'FRAME'].includes(node?.type || '')) {
+    const normalisedCos = Math.cos(normalizedAngleRad);
+    scale = normalisedCos;
+  } else {
+    // FIXME: fallback, but might break paintStyleMatchesColorToken
```
cuserox, line 82: _Might_ break? Is there another safe fallback for this case, or a test to doublecheck?

packages/tokens-studio-for-figma/src/plugin/figmaTransforms/gradients.ts
```diff
@@ -63,39 +63,41 @@ export function convertStringToFigmaGradient(value: string) {
...
+    // adjust angle after 45 degrees to scale down correctly towards 90 degrees
+    normalizedAngleRad = Math.PI / 2 - normalizedAngleRad;
+  }
+
+  const sin = Math.sin(rad);
+  const cos = Math.cos(rad);
+
+  let scale = 1;
+
+  if (['RECTANGLE', 'FRAME'].includes(node?.type || '')) {
```
cuserox, line 78: This could be saved as a constant above, easier to find if there's other types that need to be added later?

packages/tokens-studio-for-figma/src/plugin/figmaTransforms/gradients.ts
```diff
@@ -63,39 +63,41 @@ export function convertStringToFigmaGradient(value: string) {
...
-    // by negating the center transform
-    new Matrix([
-      [1, 0, -0.5],
-      [0, 1, -0.5],
-      [0, 0, 1],
-    ]),
-  );
+  ]);
 
-  const gradientTransformMatrix = inverse(transformationMatrix).to2DArray();
```
cuserox, line 98: `inverse(...)` is not needed anymore?

packages/tokens-studio-for-figma/src/plugin/figmaTransforms/gradients.ts
macintoshhelper: Have removed this as it seems to be unnecessary. Vector linear gradient scales were breaking because of this check.

packages/tokens-studio-for-figma/src/plugin/figmaTransforms/gradients.ts
macintoshhelper: Node type check removed, so this is not relevant now. `paintStyleMatchesColorToken` might fail until the nodes are updated (apply to selection/page is run), but the previous gradient should be treated as broken/incorrect I think until updated.

six7:
@macintoshhelper linting seems to have failed here

macintoshhelper:
Fixed the ESLint errors

qimmiq:
Since September, tbh this bug is so frustrating ... also to see that there is fix and every release since then does not include that fix :(

akshay-gupta7:
> Since September, tbh this bug is so frustrating ... also to see that there is fix and every release since then does not include that fix :(

@qimmiq sorry about the miss in this one
I will take a look at this PR and will keep you posted on it's progress

akshay-gupta7 (approved):

Issue #3110, opened 2024-09-02
Issue creating styles when themes are not created _exactly_ as needed


six7:
Before 2.0.2 we looked at all theme configurations an created styles based on the combination of them all, which wasn't always creating expected results.

In 2.0.2 we changed that to only look at the active theme's selectedSets, but that meant users who didnt have their `references` created properly, got incorrectly created styles

So, if you have a theme config where you have

`Core`: `Core` set is enabled
`Mode/Light`: `Light` set is enabled (no sources)

and inside `Light` you have some tokens that reference items from `Core`

And you create using 2.0.2 or 2.0.3, your styles will not have the right values and will not be created (variables work fine)

![Image](https://github.com/user-attachments/assets/f321309a-b6af-4510-84dc-e93e207e7921)

single file export
```
{
  "core": {
    "colors": {
      "red": {
        "500": {
          "value": "#ff0000",
          "type": "color"
        }
      }
    }
  },
  "mode/light": {
    "accent": {
      "default": {
        "value": "{colors.red.500}",
        "type": "color"
      }
    }
  }
}
```

PR #3111, opened 2024-09-02
fix themes: include fallback logic

six7:
This addresses a flaw introduced in 2.0.2

Fixed https://github.com/tokens-studio/figma-plugin/issues/3110

Before 2.0.2 we looked at all theme configurations an created styles based on the combination of them all, which wasn't always creating expected results.

In 2.0.2 we changed that to only look at the active theme's selectedSets, but that meant users who didnt have their `references` created properly, got incorrectly created styles

This PR changes things to have the following logic:
- We combine all theme's selectedSets into an `overallConfig` which acts as the baseline, the fallback.
- For each theme, we rank the current theme's selectedSets higher, so they will always win
- When creating styles, we combine those
- In addition, we add *all* sets to the list, with a very low ranking. This effectively means, we run WAY less into issues where we are missing 

Also, it fixes the "Remove styles" problem which was always deleting styles for users when they're not part of the current theme. Now, we look at the overall created styles and then remove those that are not touched, instead of doing so for each theme which is always wrong.

In addition, I'm disabling token-transformer tests. We're not maintaining this anymore, and this changes core logic on the transformer, which doesnt make sense to keep testing now that we have Style Dictionary.


rbosker:
@six7 One issue I found: the loader stays indicating 'Create styles', but styles are created.

https://github.com/user-attachments/assets/5d304885-3834-4ac6-b0a1-29fe5fc0759d

akshay-gupta7 (approved):

PR #3114, opened 2024-09-03
fix import of variables speed

six7:
### Why does this PR exist?

Fixes speed of variable creation by ... perceived 100x
Fixes https://github.com/tokens-studio/figma-plugin/issues/3122

Properly import variables by not doing it one by one but rather in one swift swoop

### Testing this change

Import variables into tokens... also test if you can update existing ones properly as expected

### DIFF
--- a/.changeset/wet-ways-hear.md
+++ b/.changeset/wet-ways-hear.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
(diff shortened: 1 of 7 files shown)


rbosker:
tested and this improves the import speed a lot

macintoshhelper (approved):
LGTM; tested with an import, then updating variables and importing again. Editing existing tokens seems to work too.

Importing variables with the production plugin hangs the plugin/Figma for about ten seconds, import happens in under a second now 💯 🎉

PR #3116, opened 2024-09-03
Fix export sets

akshay-gupta7:
Follow Up PR to fix styling issue where styles are not being applied when sets are being exported as styles(with ignore first part of token names for styles selected)

# Description

Fixes the issue where styles are not being applied when sets are being exported as styles (with ignore first part of token names for styles selected)

## Type of change

* [ ] None of the above

# How to test this

- Create some themes with enabled sets in the plugin
- Export selected sets as styles
- Make sure to check the 'Ignore First Part of token name for Styles' option
- Choose the sets to export as styles and export to Figma
- Apply token from a the active theme where the set belongs
- The corresponding style should now be applied

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
@@ -333,7 +333,7 @@ describe('Can set values on node', () => {
         },
       },
     );
-    expect(setTextValuesOnTargetSpy).toHaveBeenCalled();
+    expect(setTextValuesOnTargetSpy).not.toHaveBeenCalled();
     expect(textNodeMock).toEqual({ ...textNodeMock });
   });
 
@@ -453,8 +453,8 @@ describe('Can set values on node', () => {
         },
       },
     );
-    expect(setEffectValuesOnTargetSpy).toHaveBeenCalled();
-    expect(solidNodeMock).toEqual({ ...solidNodeMock, effectStyleId: '' });
+    expect(setEffectValuesOnTargetSpy).not.toHaveBeenCalled();
+    expect(solidNodeMock).toEqual({ ...solidNodeMock, effectStyleId: '123' });
   });
 
   it('sets fillStyle if matching Style', async () => {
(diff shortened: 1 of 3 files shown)


packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
```diff
@@ -333,7 +333,7 @@ describe('Can set values on node', () => {
         },
       },
     );
-    expect(setTextValuesOnTargetSpy).toHaveBeenCalled();
+    expect(setTextValuesOnTargetSpy).not.toHaveBeenCalled();
```
six7, line 336: i dont understand why this would not be expected? if i read the text description `it('sets textstyle if matching Style is found and first part is ignored', async () => {` it should expect it?

packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
```diff
@@ -453,8 +453,8 @@ describe('Can set values on node', () => {
         },
       },
     );
-    expect(setEffectValuesOnTargetSpy).toHaveBeenCalled();
-    expect(solidNodeMock).toEqual({ ...solidNodeMock, effectStyleId: '' });
+    expect(setEffectValuesOnTargetSpy).not.toHaveBeenCalled();
```
six7, line 456: same here, i would expect it to be called because it was able to find it?

akshay-gupta7: the setEffectValuesonTarget is not supposed to be called since the theme prefixed style is found.
It would have been called as a fallback had the style not be found

packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
akshay-gupta7: the setTextValuesonTarget is not supposed to be called since the theme prefixed style is found.
It would have been called as a fallback had the style not be found
I will add the textStyleId expectation since it was found

six7 (approved):

PR #3123, opened 2024-09-05
Fixes remove styles logic

six7:
This changes the logic around removing styles and variables without connection to a token. What we did before was we looked at each theme individually, looked at the tokens created in that go, and then removed any styles that were untouched.

This was flawed as it did not take into account other themes in the same session.

Changed this to look at all themes created, combine their created style ids, and only then remove any that were untouched.

To test this,
- create a few tokens with variable connections spread out across themes (e.g. a Core, and a Mode/Light, Mode/Dark)
- export using `Remove styles...`
- Notice how it now no longer removes created styles right after creating them.
Before, it did - especially if you were only exporting one of your themes - all other styles were always removed. Note this only applies to Styles, not variables!

https://github.com/user-attachments/assets/9e9c72a7-40c7-43ab-bfdf-f2c257c4eb65

https://github.com/user-attachments/assets/935b5661-9db9-44e1-b208-c91e72dc3fc5


macintoshhelper (approved):
LGTM

PR #3124, opened 2024-09-05
Fix variable creation issue

six7:
This pull request solves 2 issues:

## Font weight issues
When users were using numerical font weights in their typography tokens, and they had a varaible reference to both the family and the numerical weight, we didnt properly create styles. This only occurred if the font family and weight wasnt covered by our font weight fallbacks (e.g. use `Comic Sans MS` with a weight of `600`)
This is due to Figma requiring us to `loadFontAsync` using the string version of the weight ("Bold"). However, all we know is the numerical weight ("600"). We can not realistically know the string version. So this PR changes this to rather load *all* font weights of the current font family.

## Color token issues with modifiers when creating color styles
If users were using a color token that had a pure reference such as `{colors.gray.500}` and a modifier, and the colors.gray.500 token existed as a variable, previously we would use a variable reference. We now check if the token is using a modifier, and if yes, we do not go an create a variable reference.
This issue ONLY occurred if the Variable exists as a reference and the user is trying to create color styles using those variable reference

https://github.com/user-attachments/assets/55b3dfdb-92ec-4144-8901-36cc739dcd7f

https://github.com/user-attachments/assets/66ef0c75-e144-4113-9c47-66c8186e6b91
[example.json](https://github.com/user-attachments/files/16904363/example.json)


akshay-gupta7 (approved):

macintoshhelper (approved):
LGTM

PR #3128, opened 2024-09-06
add filter to search global

six7:
Changes the token filter to also filter the list of sets on the left by:
- the set name
- if a set includes a token that matches the filter

if a set includes tokens, we show the count of tokens inside.

CleanShot 2025-04-26 at 00 19 06@2x
CleanShot 2025-04-26 at 00 19 15@2x


six7:
merging, low risk

PR #3137, opened 2024-09-14
fix zombie variables

six7:
### Why does this PR exist?

Closes https://github.com/tokens-studio/figma-plugin/issues/3140

This adds a check before we gather variables to use for references or for application, to see if the variable collection of those variables still exists.

### Testing this change

Open a file affected by this bug (hard to replicate). File in DM with `@rbosker` has this: https://hyma-team.slack.com/archives/D03HL8WE738/p1726231101338089 - https://www.figma.com/design/foDPZl0c28gzctr6ztvNfy/Components-Roberto-(Copy)?t=I2ASNxMojuiwTyTv-0

#### Before

CleanShot 2024-09-14 at 18 20 44@2x

#### After

CleanShot 2024-09-14 at 18 25 46@2x

### DIFF
--- a/.changeset/yellow-doors-fail.md
+++ b/.changeset/yellow-doors-fail.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fixed an issue with "zombie variables". Basically, even though a Figma file shows 0 variables, Figma's plugin API will sometimes tell us there's variables existing - probably ones that existed in the past but should be deleted - Figma seems to report those as existing still. This led to issues around applying and referencing variables where we'd point to those zombies. We now correctly check if the variable's collection still exist, and only then use those as references.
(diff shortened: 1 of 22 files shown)


akshay-gupta7 (approved):

macintoshhelper (approved):
LGTM.

Does any usage of `figma.variables.getLocalVariables()` need to be replaced with `(await getVariablesWithoutZombies())` going forward? Maybe a custom ESLint rule or something could be useful for this to prevent the bug from being reintroduced in the future.

six7:
> LGTM.
> 
> Does any usage of `figma.variables.getLocalVariables()` need to be replaced with `(await getVariablesWithoutZombies())` going forward? Maybe a custom ESLint rule or something could be useful for this to prevent the bug from being reintroduced in the future.

Yes! There's a eslint rule by figma provided here: https://www.figma.com/plugin-docs/migrating-to-dynamic-loading/

we should add that

PR #3138, opened 2024-09-14
fix override order

six7:
### Why does this PR exist?

Closes https://github.com/tokens-studio/figma-plugin/issues/3139

When we're merging token sets, we weren't ordering token sets as they should. Instead, we were relying on the ordering that existed in the themes. This caused any variable creation or merging of token sets to not respect overrides coming later.

This PR fixes this. In addition, this PR fixes a bug when exporting tokens via Sets - given they both were using that same function I fixed that along the way. Basically, if you exported sets that had math calculations inside, we didnt properly calculate those, as we were only passing in the current set tokens, and not the whole tokens when resolving.

### Before

image

CleanShot 2024-09-14 at 18 14 36@2x

### After

CleanShot 2024-09-14 at 17 59 38@2x

CleanShot 2024-09-14 at 18 15 02@2x

### Testing this change

Load the .JSON linked in the Issue (single file import). Try to create before and after.

### DIFF
--- a/.changeset/purple-chairs-learn.md
+++ b/.changeset/purple-chairs-learn.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fixed an issue introduced in 2.1.0 that sometimes caused the order of token sets to not be as expected, meaning sets that acted as overrides didn't correctly get calculated.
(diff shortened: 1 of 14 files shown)


macintoshhelper (approved):
LGTM

PR #3141, opened 2024-09-17
metrics update

six7:
Adds a few new metrics to track for mixpanel


macintoshhelper (approved):
LGTM. Haven't tested the full flow of a prod build and verifying that everything is going to Mixpanel, but trust that it works since there's no change to the Mixpanel track function or config.

Issue #3118, opened 2024-09-03
Typography Composite Token not being Overridden 


keeganedwin:
**Describe the bug**
In `38.9` users were able to override a single property of a typography composite token with the appropriate token type. For example, if a typography token is applied, a user can override just the `fontFamily` property with a `font-family` token. This is however no longer possible in `2.0` and proceeding versions.

**To Reproduce**
1. Create a typography token with references
2. Apply the token to a text
3. Apply a any token overriding one of the typography properties

**Expected behavior**

**Screenshots or Screencasts**

**Figma file (optional)**
Send us the link to a Figma file containing a reproducible example

**JSON (optional)**
If possible, attach the JSON that leads to this bug.

six7:
I think this is solved in 2.1.0 - can you confirm @keeganedwin ?

keeganedwin:
> I think this is solved in 2.1.0 - can you confirm @keeganedwin ?

Bug is still present

PR #3157, opened 2024-09-26
fix typography tokens resolution

akshay-gupta7:
Addresses #3118 

# Description

Fixes a bug occurring when a user applies a typography token on a text node, and then applies another token, tinkering with an individual typography property(like font size or font family, etc.), and the user clicks on 'Apply to selection', the typography token overrides the individual token property, which is not expected behaviour.

## Type of change

* [ ] None of the above

# How to test this

* Create a typography token with references
* Apply the token to a text
* Apply a any token overriding one of the typography properties(like font size, or font family)
* Then Apply to selection
* The individual token typography should be applied and work as expected

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
@@ -235,7 +235,7 @@ describe('Can set values on node', () => {
     expect(setTextValuesOnTargetSpy).not.toHaveBeenCalled();
   });
 
-  it('calls setTextValuesOnTarget if text node and composite typography tokens are given', async () => {
+  it('does not call setTextValuesOnTarget if text node and composite typography tokens are given', async () => {
     defaultTokenValueRetriever.initiate({
       tokens: [
         {
--- a/packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
@@ -265,7 +265,7 @@ describe('Can set values on node', () => {
         },
       },
     );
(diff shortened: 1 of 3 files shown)


packages/tokens-studio-for-figma/src/plugin/__tests__/setValuesOnNode.test.ts
```diff
@@ -265,7 +265,7 @@ describe('Can set values on node', () => {
         },
       },
     );
-    expect(setTextValuesOnTargetSpy).toHaveBeenCalled();
+    expect(setTextValuesOnTargetSpy).not.toHaveBeenCalled();
   });
```
akshay-gupta7, line 269: This update in condition is in line with my new logic, where in, when a node has both individual typography value tokens and a typography token, it will not call setTextValuesOnTarget directly, instead will route to tryApplyTypographyCompositeVariable.

packages/tokens-studio-for-figma/src/plugin/applyTypographyTokenOnNode.ts
```diff
@@ -8,80 +8,75 @@ import { NodeTokenRefMap } from '@/types/NodeTokenRefMap';
...
-      resolvedValue: resolvedValueObject,
-      baseFontSize,
-    });
+
+  // Apply matching style or fallback to applying values
+  if (matchingStyleId && (await trySetStyleId(node, 'text', matchingStyleId))) return;
+
+  // Apply typography token directly if no other properties exist
+  if (data.typography && resolvedToken && isSingleTypographyValue(resolvedToken.value) && !Object.keys(values).length) {
+    setTextValuesOnTarget(node, data.typography, baseFontSize);
```
akshay-gupta7, line 67: now we will call this setTextValuesOnTarget here when a text Node has only a typography token applied to it

macintoshhelper (approved):
LGTM

PR #3158, opened 2024-09-27
fix(tokenState): setTokensFromVariables description comparison

### LINKED ISSUE #3135: setTokensFromVariables does not correctly compare descriptions
**Describe the bug**
On import variables, when determining if a token is existing or updated, the comparison of the old token and new token description leads to false positives or negatives as it could compare different types of values.

This because of the fact that:
1. Figma provides empty descriptions as empty string.
2. Tokens Studio saves empty descriptions as a value of `undefined`

**To Reproduce**
Steps to reproduce the behaviour:
- Create token without description in Tokens Studio
- Export to Figma.
- Import to variables to Tokens Studio.
- See that the existing token is falsely determined as a new token.

**Expected behaviour**
See that no tokens have changed.

**Proposal**
1. Normalize description before check if they are equal.

From: 
https://github.com/tokens-studio/figma-plugin/blob/45697ef5023dc171587eadf2b1128b607ef5e0d0/packages/tokens-studio-for-figma/src/app/store/models/tokenState.tsx#L392

To:
```ts
const normalizedOldValueDescription = oldValue.description ?? '';
const normalizedTokenDescription = token.description ?? '';
if (isEqual(oldValue.value, token.value) && isEqual(normalizedOldValueDescription, normalizedTokenDescription)) {
```

2. Reuse logic from `setTokensFromStyles` in `setTokensFromVariables`
https://github.com/tokens-studio/figma-plugin/blob/45697ef5023dc171587eadf2b1128b607ef5e0d0/packages/tokens-studio-for-figma/src/app/store/models/tokenState.tsx#L349-L350

3. Abstract token determination to separate function for both `setTokensFromStyles` and `setTokensFromVariables`

---

If you agree that this would be a bug, I'm happy to provide a PR with one of the options above, depending on what the preference of the maintainers is.

mrtnvh:
### Why does this PR exist?

Closes #3135

Normalizes old token value and new token value descriptions before comparing equality.

### Testing this change

🚧 Could use some help with adding a test, I am unfamiliar with the state setup used in the project. 

In basis, the old value contains a description value of an empty string and the new description contains a value of `undefined`.

### DIFF
--- a/.changeset/perfect-fishes-smoke.md
+++ b/.changeset/perfect-fishes-smoke.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+fix(tokenState): setTokensFromVariables description comparison
--- a/packages/tokens-studio-for-figma/src/app/store/models/tokenState.tsx
+++ b/packages/tokens-studio-for-figma/src/app/store/models/tokenState.tsx
@@ -338,6 +338,7 @@ export const tokenState = createModel<RootModel>()({
       const newTokens: StyleToCreateToken[] = [];
       const existingTokens: StyleToCreateToken[] = [];
       const updatedTokens: StyleToCreateToken[] = [];
+
       // Iterate over received styles and check if they existed before or need updating
       Object.values(receivedStyles).forEach((values) => {
         values.forEach((token) => {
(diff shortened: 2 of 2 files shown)


six7 (approved):
TY! 🎉

PR #3171, opened 2024-10-10
Fix duplicate display studio sync

### LINKED ISSUE #3069: Sync: Plugin is showing connection twice
When connecting a project from Studio to the plugin, the connection is shown twice. 
This is not reproducible everytime. 

Also, it seems to happen when you have two providers that have the same repo but a different branch.

Additional context from @NuthanHyma: 

> I also faced the same issue but for me! when I created the API key with the Power-user Groups option it showed an error.
When I selected the Admin Groups option it worked and showed up twice

Video recording in https://hyma-team.slack.com/archives/C03B3KA9LM8/p1723632624884239?thread_ts=1723621691.556229&cid=C03B3KA9LM8

Image

akshay-gupta7:
Resolves #3069 

Fixes an error where the Tokens Studio Sync Provider in the plugin was showing up twice.

## Type of change

* [ ] None of the above

# How to test this

* Go to Settings Tab
* Click on Add New Sync Provider
* Choose Tokens Studio(Beta)
* Add API key and connection string
* Click on Save
* You should either see and error or a success. If you see an error, check your keys and click on save again. If successful, the Tokens Studio Sync provider should no show up only once in the list of sync providers

### DIFF
--- a/.changeset/shy-ants-dance.md
+++ b/.changeset/shy-ants-dance.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+fixed an issue where the studio sync provider was mistakenly displayed twice upon setting up the connection
--- a/packages/tokens-studio-for-figma/src/app/components/StorageItemForm/TokensStudioForm.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/StorageItemForm/TokensStudioForm.tsx
@@ -58,7 +58,7 @@ export default function TokensStudioForm({
           ...validationResult.data,
           internalId: validationResult.data.internalId || generateId(24),
         } as ValidatedFormValues;
-        setIsLoading(true); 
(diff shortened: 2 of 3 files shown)


six7 (changes requested):
This doesnt fix the issue for already duplicative registered providers. Also, the issue mentions this also happens for other sync providers such as Supernova, can you check that as well?

six7:
I'll be closing this in favor of https://github.com/tokens-studio/figma-plugin/pull/3173

Reasoning: Doing a uniqueness check on useEffect is an anti-pattern. We should always clean our data when we retrieve it. In our case, we already had a uniqueness check, but it was just not executed properly for some sync providers. In the case of Studio we *always* just compared the URL, never incldued the api key. In Supernova case, we didnt include the mapping. This now changes that, but also ALWAYS prefers the unique id.

Issue #3169, opened 2024-10-07
Published Changes in Figma Shows all Changes in the File Modified


keeganedwin:
**Describe the bug**
There seems to have been some change made in the plugin that's affecting the changes displayed when publishing Figma libraries. Previously, when users would make token changes and update anything in a file, Figma would only show the `modified changes` and not everything else. Currently, the plugin shows that everything in the file is `modified`. This Seems to affect styles, variables and components.

> Here's the [slack thread ](https://tokens-studio.slack.com/archives/C06MVD961K8/p1721153538418999) wherein this was issue was last discussed. As discussed in the thread, the the user of the thread continued to face this issue at the time but one (Simone) didn't. However Simone recently [reported](https://tokens-studio.slack.com/archives/C06MVD961K8/p1727098659276589?thread_ts=1721153538.418999&cid=C06MVD961K8) that they're facing it now as well.

**To Reproduce**
1. Export any - variables or styles - to Figma
2. Publish Figma file
3. Change one token and update it's variable/style
4. Publish the file
5. See error - Figma displays that more than 1 variable/style have been modified

**Expected behavior**
Only the modified changes should be displayed in Figma.

**Screenshots or Screencasts**
This recording is from a user who's facing this with styles
https://github.com/user-attachments/assets/7e5ec554-be8c-41d5-aa37-aaaf80be2cc5

six7:
Bringing conversations from Slack here:

- Turns out we didn't have a check for if variable values were the same before we updated them
- Styles already had this check, so this should not happen for Styles - let's focus this issue on Variables

PR #3188, opened 2024-10-24
Check variable value before update

six7:
### Why does this PR exist?

Closes https://github.com/tokens-studio/figma-plugin/issues/3169

Variables weren't checking if they are supposed to update, they just did.

With this change we're making it so that we only update the variable value when the variable value differs from the new value.

### Testing this change

Create a few variables, hit Publish in Figma, and then hit export again. In the production version you'd now get another blue dot on the library icon, in this version you don't.

### Additional Notes (if any)

https://github.com/user-attachments/assets/5549d755-8c2c-4190-aa9d-476aec6970ba

### DIFF
--- a/.changeset/swift-apricots-sin.md
+++ b/.changeset/swift-apricots-sin.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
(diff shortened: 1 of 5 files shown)


six7:
@keeganedwin were you able to test this?

keeganedwin:
@six7 tested the and seems to be working as showed in the video.

akshay-gupta7 (approved):

PR #3198, opened 2024-11-04
Fix bitbucket sync

### LINKED ISSUE #3196: BitBucket syncing issue
**Describe the bug**
When having multifile sync enabled, pushing tokens work and reflect on the remote. However, when you pull these tokens, some sets are removed and not shown in the plugin.

**To Reproduce**

1. Add syncing method bitbucket to the testfile below
2. Sync tokens (multi file) to Bitbucket (and see that all tokens are there)
3. Afterwards pull these tokens in the plugin and see that some sets are removed.

Test file: https://www.figma.com/design/dFR1AnD9trEJCPfSDslczR/BitBucket-sync?node-id=0-1&t=7Bq0Okcyz6v0gDKp-1

Related previous pR: https://github.com/tokens-studio/figma-plugin/issues/3070

Might relates to this issue that just came in: https://github.com/tokens-studio/figma-plugin/issues/3197

akshay-gupta7:
### Why does this PR exist?

Fixes #3196 

This PR increases the amount of json files returned to 100, which is the maximum allowed from their API.
The default was 10, hence the users were unable to fetch the set json files, when the number of sets was beyond 10.

### Testing this change

- Create a token set subfolder with more than 10 sets.
- Push it to Bitbucket.
- Then pull it from the same repo.
- Now it will should be pulling all the sets(unless the count is more than 100)

### DIFF
--- a/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
@@ -145,33 +145,41 @@ export class BitbucketTokenStorage extends GitTokenStorage {
    */
 
   private async fetchJsonFilesFromDirectory(url: string): Promise<FetchJsonResult> {
-    const paginatedUrl = `${url}?pagelen=100`;
+    let allJsonFiles: any[] = [];
+    let nextPageUrl: string | null = `${url}?pagelen=100`;
 
-    const response = await fetch(paginatedUrl, {
-      headers: {
-        Authorization: `Basic ${btoa(`${this.username}:${this.secret}`)}`,
-      },
-    });
+    while (nextPageUrl) {
+        const response = await fetch(nextPageUrl, {
+            headers: {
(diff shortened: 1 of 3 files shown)


packages/tokens-studio-for-figma/src/storage/BitbucketTokenStorage.ts
```diff
@@ -145,7 +145,9 @@ export class BitbucketTokenStorage extends GitTokenStorage {
    */
 
   private async fetchJsonFilesFromDirectory(url: string): Promise<FetchJsonResult> {
-    const response = await fetch(url, {
+    const paginatedUrl = `${url}?pagelen=100`;
+
+    const response = await fetch(paginatedUrl, {
```
six7, line 150: Can you extend the functionality of this PR to fetch as many times as needed? Like if i have 352 files we should fetch 4 times

akshay-gupta7: got it, done!

six7 (approved):
🎉

PR #3199, opened 2024-11-04
Fix gitlab gitkeep issue

### LINKED ISSUE #3160: Issue with .gitkeep Files and Branch Switching
**Describe the bug**
User has a problem where when the try to push on gitlab the screenshot provided below shows up and they cannot see what is the problem but can just close it. 

When switching branches, the presence of `.gitkeep` files, which are used to preserve empty directories in Git, seems to be related to a post 2.0 bug fix for GitLab. This mechanism may not be necessary and could be removed after thorough testing to ensure the bug is not reintroduced.

**To Reproduce**
1. Create a repository with empty directories containing .gitkeep files.
2. Switch between branches and observe the behavior of these directories.
3. Note any inconsistencies or issues that arise.

**Expected behavior**
Empty directories should be preserved without the need for .gitkeep files, especially if other files are added later.

Screenshot 2024-09-18 at 14 35 00

akshay-gupta7:
### Why does this PR exist?

Fixes #3160 

When gitlab creates empty folders/directories, it creates an empty .gitkeep files.
When our plugin users push tokens to those directories, they found that the gitkeep files still exist, thereby interfering with their token pipelines.

This PR adds ability to remove gitkeep files in non empty directories, the next time user pushes a commit to Gitlab

### Testing this change

- Push a commit to an existing directory with gitkeep files
- The commit should also take care of removing those redundant gitkeep files


keeganedwin:
Tested the fix and it works

packages/tokens-studio-for-figma/src/storage/GitlabTokenStorage.ts
```diff
@@ -198,33 +237,30 @@ export class GitlabTokenStorage extends GitTokenStorage {
         errorMessage: ErrorMessages.VALIDATION_ERROR,
       };
     } catch (err) {
-      // Raise error (usually this is an auth error)
       console.error(err);
+      return [];
```
six7, line 241: could this have side-effects? we're now no longer returning [] when none of the above happens 🤔

packages/tokens-studio-for-figma/src/storage/GitlabTokenStorage.ts
```diff
@@ -198,33 +237,30 @@ export class GitlabTokenStorage extends GitTokenStorage {
...
-    const pathToCreate = this.path.endsWith('.json') ? this.path : `${this.path}/.gitkeep`;
     try {
-      await this.gitlabClient.RepositoryFiles.show(this.projectId, pathToCreate, branch);
-    } catch (e) {
+      await this.gitlabClient.RepositoryFiles.show(this.projectId, gitkeepPath, branch);
+    } catch {
       await this.gitlabClient.RepositoryFiles.create(
         this.projectId,
-        pathToCreate,
+        gitkeepPath,
```
six7, line 263: i found it confusing that we are changing this to gitKeeppath. its still the path where we are creating, and not the gitkeep path?

packages/tokens-studio-for-figma/src/storage/GitlabTokenStorage.ts
```diff
@@ -236,40 +272,28 @@ export class GitlabTokenStorage extends GitTokenStorage {
...
-    }
+
+    const gitlabActions: CommitAction[] = Object.entries(changeset).map(([filePath, content]) => {
+      const action = tree.some((file) => file.path === filePath) ? 'update' : 'create';
+      return { action, filePath, content };
+    });
 
     try {
-      const response = await this.gitlabClient.Commits.create(
+      await this.gitlabClient.Commits.create(
```
six7, line 282: does this have a side-effect? we are kind of returning if it was a successful push, now we are depending on the gitkeep stuff. but if its a single file sync we should never even call the gitkeep action

packages/tokens-studio-for-figma/src/storage/GitlabTokenStorage.ts
```diff
@@ -236,40 +272,28 @@ export class GitlabTokenStorage extends GitTokenStorage {
...
       );
-      return !!response;
     } catch (e: any) {
       if (e.cause.description && String(e.cause.description).includes(ErrorMessages.GITLAB_PUSH_TO_PROTECTED_BRANCH_ERROR)) {
         throw new Error(ErrorMessages.GITLAB_PUSH_TO_PROTECTED_BRANCH_ERROR);
       }
       throw new Error(e);
     }
+
+    await this.cleanupGitkeepFiles(rootPath, branch, message);
```
six7, line 295: not needed for single file sync

six7 (approved):

Issue #3208, opened 2024-11-22
Export Variables: once Variables are exported, you can't update a reference to raw value and vice versa


rbosker:
**Describe the bug**
once Variables are exported, you can't update a reference to raw value and vice versa

**To Reproduce**
1. Create 2 color tokens: Token 1 having a raw hex value and Token 2 referencing Token 1
2. Export these as Variables (Themes export)
3. Change Token 2 to a raw HEX value and in Token 1 you reference Token 2 now
4. Export your variables again.
5. See that the Variable that first referenced another value stays as a referenced token and the color is not updated. The variable that was initially a raw value, updates the value, but isn't changing to a referenced token/variable.

**Expected behavior**
Expect that Token 1 updates to a referenced Variable in step 4. Expect Token 2 to update to a Hex value with the correct value.

**Screenshots or Screencasts**
https://github.com/user-attachments/assets/7c27dda7-dd21-40de-93d4-0b4b3e20795b

**Note**
This probably also applies for other token types.

liamco:
Not sure if this helps or muddies the waters but I just submitted a bug report where we're having a pretty similar issue; #3209

DarioSoller:
We are also affected by that bug at the moment. FYI: For us at the moment it's with some inset spacing tokens/variables, as this issue does not seem to be limited to a certain token type.

akshay-gupta7:
> We are also affected by that bug at the moment. FYI: For us at the moment it's with some inset spacing tokens/variables, as this issue does not seem to be limited to a certain token type.

@DarioSoller We have a fix for this in progress, which should be released soon.

DarioSoller:
@akshay-gupta7 nice that sounds promising. Actually for us this decides, if we will ship another design token release before Xmas or not. Looking forward to it. Thx for the quick update!

six7:
✅ This has been released with 2.2.3

PR #3219, opened 2024-12-04
fix variable setting for various cases

six7:
### Why does this PR exist?

Closes https://github.com/tokens-studio/figma-plugin/issues/3208

The prod version currently just checks for equality of old value vs new value, but forgets that alias references could exist which would always mean they're not equal - thus we're not updating the reference when the user goes from reference -> hard value. It worked before if we went from reference -> another reference (because another function is responsible for that)

This PR modifies the early return to not return if we have an alias reference, and instead go and update.

### Testing this change

Create variables of all types and export to Figma that are a hard value and the other a reference to another variable
Then update the variable that references to now be a hard value again.

### DIFF
--- a/.changeset/seven-keys-tease.md
+++ b/.changeset/seven-keys-tease.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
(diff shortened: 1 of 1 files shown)


rbosker:
Tested with exporting as sets and exporting as themes. Both work as expected and you can change a reference to raw value and vice versa again.

akshay-gupta7 (approved):
Working as expected!✨

PR #3226, opened 2024-12-11
fix applying typography tokens variable values

### LINKED ISSUE #3215: Applying Typography tokens doesn't apply separate Variable values, but raw values.
**Describe the bug**
When you're only using Variables and NOT text styles, applying Typography tokens doesn't apply separate Variable values, BUT raw values.

**To Reproduce**
1. Create a Font-Family and Spacing token (for example)
2. Create a Typography token referencing these values
3. Export to Variables, so the font family and spacing token are created as variable
4. Try to apply your typography token and see that raw values are applied

**Expected behavior**
Expect the separate variables to be applied, instead of the raw value.

**Note**
1. Separate tokens (font family/spacing) can be applied as variables
2. Composition token type doesn't seem to have this issue.

akshay-gupta7:
### Why does this PR exist?

Closes #3215 

Addresses the concern where a typography token is applied with exported variables and the applied variable is not displayed in Figma.

### Testing this change

- Create a typography token
- Export as variable
- Apply the token on a text node
- The corresponding variable of the applied token should be displayed in Figma


packages/tokens-studio-for-figma/src/plugin/applyTypographyTokenOnNode.ts
```diff
@@ -8,16 +8,40 @@ import { NodeTokenRefMap } from '@/types/NodeTokenRefMap';
...
-    fontWeight: isPrimitiveValue(data.fontWeights) ? String(data.fontWeights.startsWith('{') ? data.fontWeights : `{${data.fontWeights}}`) : undefined,
-    lineHeight: isPrimitiveValue(data.lineHeights) ? String(data.lineHeights.startsWith('{') ? data.lineHeights : `{${data.lineHeights}}`) : undefined,
-    fontSize: isPrimitiveValue(data.fontSizes) ? String(data.fontSizes.startsWith('{') ? data.fontSizes : `{${data.fontSizes}}`) : undefined,
-    letterSpacing: isPrimitiveValue(data.letterSpacing) ? String(data.letterSpacing.startsWith('{') ? data.letterSpacing : `{${data.letterSpacing}}`) : undefined,
-    paragraphSpacing: isPrimitiveValue(data.paragraphSpacing) ? String(data.paragraphSpacing.startsWith('{') ? data.paragraphSpacing : `{${data.paragraphSpacing}}`) : undefined,
-    textCase: isPrimitiveValue(data.textCase) ? String(data.textCase.startsWith('{') ? data.textCase : `{${data.textCase}}`) : undefined,
-    textDecoration: isPrimitiveValue(data.textDecoration) ? String(data.textDecoration.startsWith('{') ? data.textDecoration : `{${data.textDecoration}}`) : undefined,
+    fontFamily: formatValue(
+      data.fontFamilies ||
+        data.fontFamily ||
```
six7, line 22: when is it the singular? i console logged and we always have the plural for fontFamilies, fontWeights, lineHeights, fontSizes - so i'm not sure you need the singular version of these?

six7 (commented):
changeset please 🙏

packages/tokens-studio-for-figma/src/plugin/applyTypographyTokenOnNode.ts
akshay-gupta7: right, I was working with a solution with values, hence I missed to update the logic

six7 (approved):

PR #3234, opened 2024-12-23
fix ts sync env

six7:
fixes the env variable for ts sync


six7:
merging without approval as required fix

PR #3241, opened 2025-01-07
Fix style variable decimal import

### LINKED ISSUE #2706: Importing variables or styles with decimal numbers leads to unexpected results
As we're converting to a base font size token (e.g. 16px) we're converting px based values to rem based values. However, while doing that conversion we seem to not respect any max decimal size.

Happens both on variable import and text style import. There's likely a step inbetween where we should round to 3 decimal numbers

To reproduce
- Create a number variable with 33.6 as its value
- Click on Import variables

or

- Create a text style with 33.6 as the font size
- Click on Import styles

Image

### LINKED ISSUE #3213: Import Number Variables as dimensions in rem units results in NaNrem when Base font size setting is empty
**Describe the bug**
Import Number Variables as dimensions in rem units results in NaNrem when Base font size setting is empty.

image

Both Jan and Keegan recall this issue being related to this → https://github.com/tokens-studio/figma-plugin/issues/2742

**To Reproduce**

In a Figma file with number variables
1. Go to the settings page of the plugin
2. Open the Base Font Size Setting
3. Ensure the input is empty (no value)
- This replicates the default state of the plugin as a new user today. 

no-setting@2x

Import Variables
4. Go to Styles and Variables Menu from Tokens Page
5. Click on Import Variables
6. Select both optional checkboxes (dimension token type and value as rem)
7. The diff view will show all number variables with the value of NaNrem
- if you confirm the import, the tokens are created with this value. 

image

Solution steps
1. Go to the settings page of the plugin
2. Open the Base Font Size Setting
3. Ensure the input is not empty - type 16px 
4. Save and repeat the Import Variables steps above
- the values will now all be correctly resolved. 

image

**Expected behavior**

User Story:
If I am performing a task which will create broken tokens, 
- block me from completing the task
- surface the reason for the block in beginner-friendly language with instructions on how to self-solve. Link to external guides where needed. 

To Do:
- Define all possible errors 
- Sam to help write error messages based on the cause / links to self-help guides based on the errors. 
- Today we don't surface errors or skipped styles/variables on import, so that will need to be defined/created. 

If the default state of the plugin is contributing to issues with basic tasks, we need to change the default state. 
 - Can the Base font size setting be default to 16px? 
 - Can the Base font size input require a populated state? 

That would limit this issue to when a token referenced in this setting is broken. 
- This setting today follows me between Figma projects, files and repositories. 
- Can the value of this setting be limited to my token storage solution? If local, it is per Figma file. If remote, it is per sync connection. 

Video of the problem and solution

https://github.com/user-attachments/assets/b9804dba-14dd-4a62-a546-56b3a125a258

[Figma File](https://www.figma.com/design/dUVbLRvnvY2Hwnncs21dZG/Number-variables-only?node-id=0-1&t=IL4NKKyBHTVMFenX-1)

akshay-gupta7:
### Why does this PR exist?

Fixes #2706  
Fixes #3213 

This PR addresses the issue when we import a style with text size or a Number variable with decimal values.

The import of a number like 33.6 is reflected as 33.599... in the corresponding token in the plugin

Also, addresses the number variables imported as NaNrem when the baseFontSize in ui Setting is blank.

### Testing this change

- Create a style in Figma with textSize Decimal type or a Number type variable with a Decimal value
- Import in the plugin
- See if the decimal value is imported to the closest 3 places after the decimal
- For the second issue, go to settings and change baseFontSize to blank(thereby setting it to default 16px)
- Import a Number variable with both options of 'Convert Number to dimensions' and 'use rem for dimension values' checked.
- they should no longer be imported as NaNrem token values

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
@@ -16,10 +16,11 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
 
   let baseRem = 16;
   if (options.useRem) {
-    baseRem = await figma.clientStorage.getAsync('uiSettings').then(async (uiSettings) => {
-      const settings = JSON.parse(await uiSettings);
-      return settings.baseFontSize;
-    });
+    const uiSettings = await figma.clientStorage.getAsync('uiSettings');
+    const settings = JSON.parse(await uiSettings);
+    baseRem = settings?.baseFontSize && !isNaN(settings.baseFontSize)
+      ? Number(settings.baseFontSize)
+      : 16;
   }
 
   const localVariables = await getVariablesWithoutZombies();
--- a/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
@@ -100,9 +100,9 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
               tokenValue = `{${alias?.name.replace(/\//g, '.')}}`;
             } else if (typeof value === 'number') {
               if (options.useRem) {
-                tokenValue = `${(Number(tokenValue) / parseFloat(String(baseRem))).toFixed(3)}rem`;
+                tokenValue = `${Number((Number(tokenValue) / parseFloat(String(baseRem))).toFixed(3))}rem`;
               } else if (options.useDimensions) {
-                tokenValue = `${tokenValue.toFixed(3)}px`;
+                tokenValue = `${Number(tokenValue.toFixed(3))}px`;
               } else {
                 tokenValue = Number(tokenValue.toFixed(3));
               }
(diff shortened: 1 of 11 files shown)


packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -16,10 +16,11 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
...
   let baseRem = 16;
   if (options.useRem) {
-    baseRem = await figma.clientStorage.getAsync('uiSettings').then(async (uiSettings) => {
-      const settings = JSON.parse(await uiSettings);
-      return settings.baseFontSize;
-    });
+    const uiSettings = await figma.clientStorage.getAsync('uiSettings');
+    const settings = JSON.parse(await uiSettings);
+    baseRem = settings?.baseFontSize && !isNaN(settings.baseFontSize)
+      ? Number(settings.baseFontSize)
```
six7, line 22: this doesnt handle the case where baseFontSize is set with a unit, such as `16px` or `14px` - there should be logic in the plugin already doing that we should be able to reuse that

six7 (commented):
changeset and a test would be good

six7 (approved):
❤️

PR #3243, opened 2025-01-08
Fix ts theme metadata

roppazvan:
### Why does this PR exist?

* TS sync removed some metadata from themes like figmaModeId or figmaCollectionId. Fixed this issue by querying for that data and also updating it properly.
* Also updated the types and sdk packages

Closes #0000

### DIFF
--- a/packages/tokens-studio-for-figma/package.json
+++ b/packages/tokens-studio-for-figma/package.json
@@ -62,7 +62,7 @@
     "@supabase/supabase-js": "^2.0.5",
     "@supernovaio/supernova-sdk": "^1.9.3",
     "@tokens-studio/graph-engine": "^0.17.5",
-    "@tokens-studio/sdk": "^2.0.0-alpha.0",
+    "@tokens-studio/sdk": "^1.4.2",
     "@tokens-studio/tokens": "0.2.5",
     "@tokens-studio/types": "0.5.1",
     "@tokens-studio/ui": "0.9.0",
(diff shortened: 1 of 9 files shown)


packages/tokens-studio-for-figma/package.json
```diff
@@ -62,7 +62,7 @@
     "@supabase/supabase-js": "^2.0.5",
     "@supernovaio/supernova-sdk": "^1.9.3",
     "@tokens-studio/graph-engine": "^0.17.5",
-    "@tokens-studio/sdk": "^2.0.0-alpha.0",
+    "@tokens-studio/sdk": "^1.4.2",
```
six7, line 65: Is this on purpose? Sdk 1.4.2 is a prior version?

six7 (approved):
Changeset please

packages/tokens-studio-for-figma/package.json
roppazvan: yeah, 1.4.2 is the latest

six7: what happened to 2.0.0.-alpha?

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -427,7 +429,7 @@ export class TokensStudioTokenStorage extends RemoteTokenStorage<TokensStudioSav
       const responseData = await this.client.mutate({
         mutation: DELETE_THEME_GROUP_MUTATION,
         variables: {
-          branch: 'master',
+          branch: 'main',
```
akshay-gupta7, line 432: this should be DEFAULT_BRANCH as well right?

akshay-gupta7 (approved):

PR #3262, opened 2025-02-12
Tokens Studio sync - Raise sets limit to 1000

georgebuciuman:
### Why does this PR exist?

Closes #0000 

Current maximum limit of token sets synced from Tokens Studio provider has been raised from 100 to 1000


six7:
Changeset please

PR #3264, opened 2025-02-13
Fix default Token Format for Studio sync as DTCG

### LINKED ISSUE #3259: The tokens studio sync provider shows legacy tokens label
**Describe the bug**

**To Reproduce**
1. Create a tokens studio sync in the plugin
2. Load the tokens from the studio app
3. It is being loaded by default in "Legacy Format"

**Expected behavior**
The default should be W3C DTCG format

Image

**Figma file (optional)**

**JSON (optional)**

akshay-gupta7:
### Why does this PR exist?

Resolves #3259 

Currently, if the local storage or previous sync provider had tokens set in Legacy format, while switching to Tokens Studio Sync provider, it was being set to Leagcy as well, inspite of the tokens beign received from studio in DTCD format.
also, hides the badge of Legacy/DTCG in studio sync storage type as the default will always be DTCG

### Testing this change

- Create a sync provider in the plugin from studio
- Switch to Local Storage, change format to legacy
- Switch back to Tokens Studio storage and it should be set to DTCG by default again

### Additional Notes (if any)

Screenshot 2025-02-14 at 11 03 29 AM

### DIFF
--- a/.changeset/rotten-ads-roll.md
+++ b/.changeset/rotten-ads-roll.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fixes the default format of tokens to DTCG when tokens are synced froM Tokens Studio Platform.
--- a/packages/tokens-studio-for-figma/src/app/components/StorageItem.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/StorageItem.tsx
@@ -133,7 +133,9 @@ const StorageItem = ({ item, onEdit }: Props) => {
       <Box css={{ marginRight: '$1' }}>
         {isActive() ? (
           <Stack gap={2} align="center">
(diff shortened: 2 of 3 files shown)


six7 (changes requested):
I think this issue will be obsolote once you complete https://github.com/tokens-studio/studio-app/issues/2152 which will return them as DTCG

akshay-gupta7:
> I think this issue will be obsolote once you complete [tokens-studio/studio-app#2152](https://github.com/tokens-studio/studio-app/issues/2152) which will return them as DTCG

This issue does not appear to be particularly from studio sync, rather the tokenFormat persists the tokenFormat of the previous sync provider(or local), and if the previous one is legacy and then they choose studio sync, it sets to legacy regardless of the format it is receiving the tokens in

six7 (approved):
lets just add a changeset

PR #3268, opened 2025-02-17
Fix gitlab bug, gitkeep file created on every commit

akshay-gupta7:
### Why does this PR exist?

Fixes an issue in Gitlab sync. When a user pushes a change as a commit, gitkeep file is being created, leading to 3 commits for each change.

### Testing this change

- Create a sync in Gitlab Repo
- Make a change
- Commit the change
- It should now be only one commit(if gitkeep file was already removed previously)

### DIFF
--- a/.changeset/gentle-pets-confess.md
+++ b/.changeset/gentle-pets-confess.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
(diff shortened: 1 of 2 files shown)


six7 (approved):
can you get a review from Keegan or Robbert?

akshay-gupta7:
> can you get a review from Keegan or Robbert?

yes I have sent it to Keegan and one of our users Andy Sobol as well to confirm, I will wait for their feedback before merging

PR #3270, opened 2025-02-18
Fix import issue

### LINKED ISSUE #3260: When importing variables, import as `Theme groups`
Currently when users import variables we only create or update a token set with the name of the collection.

However, we do not create theme groups for them. This leaves users with half the job done, and half still to do manually. We can optimize this.

## Introducing theme group creation on import

When importing variables, we will from now on
- look at the name of the variable collection and the modes that would be imported, for this example let's use `Collection name: Theme` `Mode name: Light`
- inside the plugin, look at if the user is a `[Pro]` user. If they are NOT, do nothing
- if they *are* a [Pro] user, check to see if they have any theme groups, and if so, if any theme group's name matches the collection name
- if no theme group name that we need exists
  - create a theme group with the name of the Collection
  - inside that theme group, create theme options with the name of each `Mode`, with the "enabled sets" as the sets that is used for that (which we even today create or populate, so we have that logic in already, we just need to create the theme options with the right sets)
- if a theme group exists already with that collection name
  - Look at the theme options to see if we need to create any new ones, or if we can leave it as is

akshay-gupta7:
### Why does this PR exist?

Fixes #3260 

Currently, when users import variables, the plugin only creates corresponding sets for collections and variables.

This PR adds the feature to create theme groups as well, containing the corresponding sets created as well.

### Testing this change

- Create a variable collection with multiple modes
- Import it in the plugin
- The new sets will be created
- Now, new theme groups and options mapped to those sets will also be created

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
+++ b/packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
@@ -35,6 +35,8 @@ import { BackgroundJobs } from '@/constants/BackgroundJobs';
 import { defaultTokenResolver } from '@/utils/TokenResolver';
 import { getFormat } from '@/plugin/TokenFormatStoreClass';
 import { ExportTokenSet } from '@/types/ExportTokenSet';
+import { licenseKeySelector } from '@/selectors/licenseKeySelector';
+import { licenseKeyErrorSelector } from '@/selectors/licenseKeyErrorSelector';
 
 type ConfirmResult = ('textStyles' | 'colorStyles' | 'effectStyles' | string)[] | string;
 
@@ -68,6 +70,9 @@ export default function useTokens() {
   const store = useStore<RootState>();
   const tokensContext = useContext(TokensContext);
   const shouldConfirm = useMemo(() => updateMode === UpdateMode.DOCUMENT, [updateMode]);
+  const existingKey = useSelector(licenseKeySelector);
+  const licenseKeyError = useSelector(licenseKeyErrorSelector);
+  const proUser = Boolean(existingKey && !licenseKeyError);
   const VALID_TOKEN_TYPES = [
     TokenTypes.DIMENSION,
     TokenTypes.BORDER_RADIUS,
@@ -205,9 +210,11 @@ export default function useTokens() {
           useDimensions: userDecision.data.includes('useDimensions'),
           useRem: userDecision.data.includes('useRem'),
         },
+        themes,
+        proUser,
       });
     }
-  }, [confirm]);
+  }, [confirm, themes, proUser]);
 
   const removeTokensByValue = useCallback((data: RemoveTokensByValueData) => {
     track('removeTokensByValue', { count: data.length });
--- a/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
@@ -1,12 +1,13 @@
 /* eslint-disable no-param-reassign */
 import { figmaRGBToHex } from '@figma-plugin/helpers';
 import { notifyVariableValues } from './notifiers';
-import { PullVariablesOptions } from '@/types';
+import { PullVariablesOptions, ThemeObjectsList } from '@/types';
 import { VariableToCreateToken } from '@/types/payloads';
 import { TokenTypes } from '@/constants/TokenTypes';
 import { getVariablesWithoutZombies } from './getVariablesWithoutZombies';
+import { TokenSetStatus } from '@/constants/TokenSetStatus';
 
-export default async function pullVariables(options: PullVariablesOptions): Promise<void> {
+export default async function pullVariables(options: PullVariablesOptions, themes: ThemeObjectsList, proUser: boolean): Promise<void> {
   // @TODO should be specifically typed according to their type
   const colors: VariableToCreateToken[] = [];
   const booleans: VariableToCreateToken[] = [];
@@ -29,10 +30,23 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
 
   const localVariables = await getVariablesWithoutZombies();
 
-  localVariables.forEach((variable) => {
+  const collections = new Map<string, {
+    id: string,
+    name: string,
+    modes: { name: string, modeId: string }[]
+  }>();
+
+  localVariables.forEach(async (variable) => {
+    const collection = figma.variables.getVariableCollectionById(variable.variableCollectionId);
+    if (collection) {
+      collections.set(collection.name, {
+        id: collection.id,
+        name: collection.name,
+        modes: collection.modes.map((mode) => ({ name: mode.name, modeId: mode.modeId })),
+      });
+    }
     const variableName = variable.name.replace(/\//g, '.');
     try {
-      const collection = figma.variables.getVariableCollectionById(variable.variableCollectionId);
       switch (variable.resolvedType) {
         case 'COLOR':
           Object.entries(variable.valuesByMode).forEach(([mode, value]) => {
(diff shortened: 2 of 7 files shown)


packages/tokens-studio-for-figma/src/app/store/useTokens.tsx
```diff
@@ -68,6 +70,9 @@ export default function useTokens() {
   const store = useStore<RootState>();
   const tokensContext = useContext(TokensContext);
   const shouldConfirm = useMemo(() => updateMode === UpdateMode.DOCUMENT, [updateMode]);
+  const existingKey = useSelector(licenseKeySelector);
+  const licenseKeyError = useSelector(licenseKeyErrorSelector);
+  const proUser = Boolean(existingKey && !licenseKeyError);
```
six7, line 75: i think we should have a `prouser` property in uiState somewhere - i might be wrong though. basically the one that we use to display the `GET PRO` badge or the `PRO` badge - is that this one?

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -29,10 +30,23 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
...
 
-  localVariables.forEach((variable) => {
+  const collections = new Map<string, {
+    id: string,
+    name: string,
+    modes: { name: string, modeId: string }[]
+  }>();
+
+  localVariables.forEach(async (variable) => {
+    const collection = figma.variables.getVariableCollectionById(variable.variableCollectionId);
```
six7, line 40: is this a slow task? should we try to cache this? also.. this is `deprecated` according to the figma plugin api and we should use the async version: https://www.figma.com/plugin-docs/api/properties/figma-variables-getvariablecollectionbyid/

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -151,12 +165,59 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
...
-  const returnedObject = Object.entries(stylesObject).reduce<ResultObject>((acc, [key, value]) => {
-    if (value.length > 0) {
-      acc[key] = value;
-    }
-    return acc;
-  }, {});
+  // Process themes if pro user
+  const updatedThemes = [...themes];
+  if (proUser) {
+    collections.forEach((collection) => {
```
six7, line 171: can we also instead of doing a forEach do a promise.all to have them run all at the same time?

six7 (changes requested):

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -151,12 +165,59 @@ export default async function pullVariables(options: PullVariablesOptions): Prom
...
+            id: `${collection.name.toLowerCase()}-${mode.name.toLowerCase()}`,
+            name: mode.name,
+            group: collection.name,
+            selectedTokenSets: {
+              [`${collection.name}/${mode.name}`]: TokenSetStatus.ENABLED,
+            },
+            $figmaStyleReferences: {},
+            $figmaModeId: mode.modeId,
+            $figmaCollectionId: collection.id,
+          });
```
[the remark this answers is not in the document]
akshay-gupta7: yeah, I updated the logic a bit, to fix a bug for the variables being imported as themes even before clicking 'Import All'

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -36,8 +36,8 @@ export default async function pullVariables(options: PullVariablesOptions, theme
     modes: { name: string, modeId: string }[]
   }>();
 
-  localVariables.forEach(async (variable) => {
-    const collection = figma.variables.getVariableCollectionById(variable.variableCollectionId);
+  for (const variable of localVariables) {
+    const collection = await figma.variables.getVariableCollectionByIdAsync(variable.variableCollectionId);
```
six7, line 40: i could imagine this to still be slow if you have 4000 variables. can we cache this?

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
```diff
@@ -165,59 +165,44 @@ export default async function pullVariables(options: PullVariablesOptions, theme
...
-        }
-      });
-    });
+    await Promise.all(Array.from(collections.values()).map(async (collection) => {
+      await Promise.all(collection.modes.map(async (mode) => {
+        const collectionVariables = localVariables.filter((v) => v.variableCollectionId === collection.id);
+
+        const variableReferences = collectionVariables.reduce((acc, variable) => ({
+          ...acc,
+          [variable.name]: variable.id,
```
six7, line 177: we'll need to normalize the name here. variable names can contain a `/`. we should transform them to a `.`. there should be a utility to do that (we use that when importing variables already)

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
six7: ai generated suggestion (pls validate first)

```diff --git a/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts b/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
index b48786b62..1cc919667 100644
--- a/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
@@ -30,22 +30,30 @@ export default async function pullVariables(options: PullVariablesOptions, theme
 
   const localVariables = await getVariablesWithoutZombies();
 
+  // Cache for collection lookups
+  const collectionsCache = new Map<string, {
+    id: string,
+    name: string,
+    modes: { name: string, modeId: string }[]
+  }>();
+
   for (const variable of localVariables) {
-    const collection = await figma.variables.getVariableCollectionByIdAsync(variable.variableCollectionId);
+    // Check if collection is already cached
+    let collection = collectionsCache.get(variable.variableCollectionId);
+    
+    if (!collection) {
+      // Only fetch from Figma API if not cached
+      const collectionData = await figma.variables.getVariableCollectionByIdAsync(variable.variableCollectionId);
+      if (collectionData) {
+        collection = {
+          id: collectionData.id,
+          name: collectionData.name,
+          modes: collectionData.modes.map((mode) => ({ name: mode.name, modeId: mode.modeId })),
+        };
+        collectionsCache.set(variable.variableCollectionId, collection);
+      }
+    }
+    
     if (collection) {
-      collections.set(collection.name, {
-        id: collection.id,
-        name: collection.name,
-        modes: collection.modes.map((mode) => ({ name: mode.name, modeId: mode.modeId })),
-      });
-    }
- ```

packages/tokens-studio-for-figma/src/plugin/pullVariables.ts
akshay-gupta7: this utility?
https://github.com/tokens-studio/figma-plugin/blob/main/packages/tokens-studio-for-figma/src/utils/normalizeTokenName.ts
not sure why it says deprecated?

six7: Yeah that one. Not sure, but that does what we need

six7: Oh, in fact no - we need it the other way around

akshay-gupta7: So we need to replace the `. ` With `/` ?

six7: No vice versa, variable names contain / and we need to normalize to .

akshay-gupta7: Got it, I will create a new util if I don't find one

six7: 👍 we should have one though - when importing variables we already do the same at some point

six7 (approved):

PR #3272, opened 2025-02-19
replace tokens with raw in GetProjectData Query

akshay-gupta7:
### Why does this PR exist?

Fixes https://github.com/tokens-studio/studio-app/issues/2152

### DIFF
--- a/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
@@ -76,7 +76,6 @@ async function getProjectData(id: string, orgId: string, client: any): Promise<P
         const rawTokens = typeof tokenSet.raw === 'string'
           ? JSON.parse(tokenSet.raw)
           : tokenSet.raw;
-        // Creating new object to ensure it's extensible
         acc.tokens[tokenSet.name] = JSON.parse(JSON.stringify(rawTokens));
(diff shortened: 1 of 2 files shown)


packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -72,16 +73,11 @@ async function getProjectData(id: string, orgId: string, client: any): Promise<P
...
         if (!tokenSet.name) return acc;
-        acc.tokens[tokenSet.name] = tokenSet.tokens.map((tkn) => ({
-          name: tkn.name,
-          type: tkn.type,
-          description: tkn.description,
-          $extensions: tkn.extensions,
-          value: tkn.value,
-        }));
-
+        const rawTokens = typeof tokenSet.raw === 'string'
```
six7, line 76: do we need this? feels like we should be able to rely on the data structure to always be one?

six7 (commented):
just a comment on the condition - and pls a changeset

akshay-gupta7:
> just a comment on the condition - and pls a changeset

do we need a changeset for this since this is not particularly a user facing change?

six7:
> > just a comment on the condition - and pls a changeset
> 
> do we need a changeset for this since this is not particularly a user facing change?

you're right - we dont need one 👍

six7 (approved):

six7:
let's get this published on Monday, thank you! 🙏

PR #3305, opened 2025-04-04
fix font size style referencing a variable

### LINKED ISSUE #3299: Importing Styles with Variable References doesn't Reference the Correct Tokens
**Describe the bug**
In our docs we mention the following under importing styles and variables -
> Known limitation - Styles with Variable References
> You can still import them, but any connection to Variables will need to be manually created.

So currently it looks like the plugin is only instructed to fetch references of the base tokens it creates when importing text-styles. In this use-case:
1. Number variables are imported into the `Dimension` type.
2. Although in Figma the style is referencing a variable, on import it creates base tokens and references these _instead_ of referencing the Dimension (variables) token that it needs to.

The dimension token the typography token needs to reference is `Typography.heading.fontSize.1` which it does in Figma as a variable, but it's instead referencing `fontSize.7` token that the plugin created a base token of on importing text-styles.

**To Reproduce**
Available in test file

**Expected behavior**
I'm unaware of things work under the hood but I would expect the plugin to fetch the correct reference(s) when importing styles with variable references. 

https://github.com/user-attachments/assets/bdcea29c-ca8f-4b19-8bb9-dbec932c917b

**Figma file (optional)**
[Figma file](https://www.figma.com/design/fRW6949XiCalfFigZNYJig/Untitled?node-id=14-3&t=rKbiiRvJrhXJiGO5-4)

akshay-gupta7:
### Why does this PR exist?

Fixes #3299 

Currently, in our plugin we have a limitation of importing styles with variable references. To get around it users can import variables first and then styles. However even then the styles which have a fontsize reference to a variable end up creating a new base variable even though the variable exists as a token in our plugin.

This PR addresses that issue and links the correct referenced variable token.

### Testing this change

- Create a typography token with a reference variable in font size
- Import variables, convert numbers to dimensions
- Now import styles
- The created typography token will now reference the correct variable reference token, instead of creating a new one


packages/tokens-studio-for-figma/src/plugin/pullStyles.ts
```diff
@@ -93,45 +97,206 @@ export default function pullStyles(styleTypes: PullStyleOptions): void {
       if (!rawTextDecoration.includes(style.textDecoration)) rawTextDecoration.push(style.textDecoration);
     });
 
-    fontSizes = rawFontSizes
-      .sort((a, b) => a - b)
-      .map((size, idx) => ({
+    fontSizes = figmaTextStyles.map((style, idx) => {
```
six7, line 100: lots of duplicate code here! there's at least 4 times the same code. I created a PR that cleans this up, feel free to merge in or adjust.

https://github.com/tokens-studio/figma-plugin/pull/3309

six7 (changes requested):
Also, please a changeset 🙏  functionality wise this works fine! code isnt great - i'd recommend to read into DRY and making sure that stuff like this isnt repeated too often - the way i did it now is more isolated and allows a dedicated test file

packages/tokens-studio-for-figma/src/plugin/pullStyles.ts
```diff
@@ -93,45 +98,113 @@ export default function pullStyles(styleTypes: PullStyleOptions): void {
...
+      }
+
+      return {
+        name: `fontFamilies.${slugify(fontFamily)}`,
+        value: fontFamily,
+        type: TokenTypes.FONT_FAMILIES,
+      };
+    });
+
+    // Remove console.log and duplicate fontSizes assignment
```
six7, line 196: whats this one

packages/tokens-studio-for-figma/src/plugin/pullStyles.ts
six7, line 1: test coverage comment shows we've decreased test coverage in this file https://github.com/tokens-studio/figma-plugin/pull/3305#issuecomment-2781459361

six7 (approved):

PR #3326, opened 2025-04-20
Move to client storage

### LINKED ISSUE #3310: Move "restore" token values to client storage for remote sync providers
## Background

Figma is introducing a 100KB limit for shared plugin data. Our plugin currently stores token data in `figma.root` containing the stringified version of all user tokens. For large token sets, this limit will be reached, preventing data from being saved.

For users with remote sync providers, token values that act as "live" tokens (tokens not yet pushed) should be stored in client storage (which has a 5MB limit) instead of shared plugin data. That means the current changes of users are personal, and not viewable by others.

## Implementation Plan

1. Create a new `ClientValuesProperty` in `figmaStorage` folder that uses client storage
2. Always use client storage for recover local changes - for all providers when saving their fallback values on the document
4. When users use a remote storage provider and they launch the plugin, first check client storage - which is going to be our "recover local changes"

## Definition of done
- [ ] Remote storage providers do not save anything on shared plugin data anymore
- [ ] Remote storage providers save their in-progress tokens on clientStorage
- [ ] When restoring, we read from clientStorage
- [ ] There is a check similar to checkForChanges that lets us detect if there's changes that the user wants to restore (TBD if that is just checkForChanges on the clientStorage, or something else)
- [ ] There is tests for this new behavior

## Testing

1. Test with a user using local storage
2. Test with a user using remote storage (Git, etc.)
3. Test migration from old storage to new storage
4. Test with token sets of various sizes

akshay-gupta7:
### Why does this PR exist?

Closes #3310 

This Pull Request stores the local changes which aren't synced to remote Providers in the client storage instead of shared Plugin Storage now.
It uses a prefix of a random UUID and file Name, to distinguish tokens between different files.

### Testing this change

- Setup a figma file with tokens synced to a sync provider
- Make some changes to tokens, close the plugin without syncing
- When you open the plugin again, it recovers local changes from the client storage
- to check if tokens and themes are stored in client storage, go to dev console, and type `figma.clientStorage.keysAsync().then(console.log)` to see the keys of tokens/values and tokens/themes, indicating that the tokens and themes did indeed save to the client storage instead of shared plugin storage.


packages/tokens-studio-for-figma/src/utils/figma/updateLocalTokensData.ts
```diff
@@ -18,12 +20,24 @@ type Payload = {
...
+    await ClientStorageProperty.write('tokens/themes', payload.themes);
+    await ClientStorageProperty.write('tokens/values', payload.tokens); // will help with migration of sharedPluginData to clientstorage as well
+    await ThemesProperty.write(payload.themes);
+    await ValuesProperty.write(payload.tokens);
+  } else {
+    await ValuesProperty.write({});
+    await ThemesProperty.write([]);
+    await ClientStorageProperty.write('tokens/themes', payload.themes);
+    await ClientStorageProperty.write('tokens/values', payload.tokens);
+  }
```
six7, line 40: does this include logic yet as per:

- if i use remote storage, we should _always_ store recover local changes data in clientStorage (we need to figure out for what key? we could store one key per file name?)
- if i use local storage, we should _always_ store in shared plugin data for now, lets not mix those concerns in this PR

six7 (commented):
@akshay-gupta7 can you separate those PRs:

1 PR to be only about the move to client storage for recover local changes stuff (and only for that, not for regular local document storage)
1 PR for the indicator in the bottom left with the dialog

that way we can ship the move to client storage for recover local changes stuff relatively soon, and can then follow up with just the dialog and size hint once we have the right wording and assets from Sam / Lucy.

packages/tokens-studio-for-figma/src/plugin/node.ts
```diff
@@ -107,8 +120,21 @@ export async function getTokenData(): Promise<{
...
+      const fileKey = await getFileKey();
+      const fileName = encodeURIComponent(figma.root.name);
+      const prefix = `${fileKey}/${fileName}/tokens`;
+      values = await ClientStorageProperty.read(`${prefix}/values`) ?? {};
+      themes = await ClientStorageProperty.read(`${prefix}/themes`) ?? [];
+    }
+
     const activeTheme = await ActiveThemeProperty.read(figma.root) ?? {};
     const version = await VersionProperty.read(figma.root);
     const updatedAt = await UpdatedAtProperty.read(figma.root);
```
six7, line 140: one other thing that i noticed just now:

the `checkForChanges` property.. we set that to `true` whenever we make changes. and then, when the user successfully pushed, we set it to `false` again.

we need to change this to `clientStorage` too. as in, if i make changes to a document, other people opening the file should not be shown that there's changes to recover (they wouldnt have access anyway).

i'd say try this out with 2 users:
1 makes changes, closes the plugin without pushing
the other opens the plugin. they should in that case not see the recover local changes dialog. 

if things already work like that, great - but i assume it would lead to some inconsistencies

packages/tokens-studio-for-figma/src/utils/figma/updateLocalTokensData.ts
```diff
@@ -18,12 +21,23 @@ type Payload = {
...
   await VersionProperty.write(pjs.version);
-  await ThemesProperty.write(payload.themes);
-  await ValuesProperty.write(payload.tokens);
+  // Check storage size and storage method
+  if (payload.storageProvider === StorageProviderType.LOCAL) {
+    await ThemesProperty.write(payload.themes);
+    await ValuesProperty.write(payload.tokens);
+  } else {
+    const fileKey = await getFileKey();
+    const fileName = encodeURIComponent(figma.root.name);
```
six7, line 36: i think we cannot use the filename as a unique identifier - this will be different when users rename, and it wont be unique.

why do we even need it? i'd remove it, the unique identifier you generate through `getFileKey` should be enough.

packages/tokens-studio-for-figma/src/utils/figma/updateLocalTokensData.ts
```diff
@@ -18,12 +21,23 @@ type Payload = {
...
+    const prefix = `${fileKey}/${fileName}/tokens`;
+    await ClientStorageProperty.write(`${prefix}/themes`, payload.themes);
```
six7, line 38: i think we need to account for the situation where the user is close to their 5MB limit and we try to save a large token set.

what happens if that fails?

in addition, i think.. we should do some garbage collection from time to time. like right now we dont ever `DELETE` those from client storage, meaning every time i do this in a new file, the data will pile up. 

can we make it so that we
- on successful push, reset the client storage property
- (specified in a comment further above), move the `checkForChanges` to client storage, and in the step where we're setting it to false also clean up the client storage?
- delete all client storage tokens except for the... 5 last saved ones? that feels like a number that means we always have enough storage left

six7 (changes requested):
overall i think this is on a great path 🎉

packages/tokens-studio-for-figma/src/app/store/updateSources.tsx
```diff
@@ -160,10 +160,16 @@ export default async function updateTokensOnSources({
     ? defaultTokenResolver.setTokens(mergeTokenGroups(tokens, usedTokenSet))
     : null;
 
+  let jsonSize;
+
   if (mergedTokens) {
     try {
-      const jsonSize = (new TextEncoder().encode(JSON.stringify(tokenValues)).length) / 1024;
-      track('tokens_size', { jsonSize });
+      const combinedData = { tokens: tokenValues, themes };
```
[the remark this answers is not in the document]
six7: I dont think this is done?

packages/tokens-studio-for-figma/src/plugin/node.ts
```diff
@@ -107,12 +121,32 @@ export async function getTokenData(): Promise<{
...
+    let values = {};
+    let themes: ThemeObjectsList = [];
+    const fileKey = await getFileKey();
+    const prefix = `${fileKey}/tokens`;
+
+    if (storageType.provider === StorageProviderType.LOCAL) {
+      values = await ValuesProperty.read(figma.root) ?? {};
+      themes = await ThemesProperty.read(figma.root) ?? [];
+    } else {
+      values = JSON.parse(decompressFromUTF16(await ClientStorageProperty.read(`${prefix}/values`))) ?? {};
```
six7, line 134: Could we move the compression into the actual read write of the actual property (like my comment in the other PR)

packages/tokens-studio-for-figma/src/plugin/node.ts
```diff
@@ -107,12 +121,32 @@ export async function getTokenData(): Promise<{
...
+      if (themes.length === 0) {
+        themes = await ThemesProperty.read(figma.root) ?? [];
+      }
+    }
+
     const activeTheme = await ActiveThemeProperty.read(figma.root) ?? {};
     const version = await VersionProperty.read(figma.root);
     const updatedAt = await UpdatedAtProperty.read(figma.root);
-    const checkForChanges = await CheckForChangesProperty.read(figma.root);
+    const checkForChanges = await ClientStorageProperty.read(`${prefix}/checkForChanges`) ?? false;
```
six7, line 149: Could there be an issue with this flow when the user has checkforchanges true on the old version and then starts the new version? Could you try that one? My assumption is we dont catch that, we should ideally also pick up that prop from the old way for the migration period

akshay-gupta7: added the original checkforchanges back, just for legacy data, maybe we can remove it post a few weeks

six7: we wont be able to do a few weeks - as users could ... just have a break for 2 months or even 6 months until they reopen the plugin

packages/tokens-studio-for-figma/src/app/store/updateSources.tsx
```diff
@@ -160,10 +161,17 @@ export default async function updateTokensOnSources({
...
+      const compressedData = compressToUTF16(JSON.stringify(combinedData));
```
six7, line 169: lets not combine the two, but track individually

packages/tokens-studio-for-figma/src/utils/figma/updateLocalTokensData.ts
six7: can you add something for this? maybe.. something that keeps track of the date of when these were saved? and then when we notice we do hit an error because of that limit we delete the oldest?

akshay-gupta7: I added a separate key called lastupdated, with the same fileKey, so that we can fetch it and delete old data based on that date

six7 (approved):

PR #3330, opened 2025-04-23
compress shared Plugin Data token values

### LINKED ISSUE #3328: Add compression to storing and reading sharedPluginData (themes and token values)
When saving and reading tokens from figma's root (token values and themes), we should use compression as outlined in https://github.com/tokens-studio/figma-plugin/pull/3325 - that way we can get our kb numbers down.

## Definition of done
- [ ] We compress when saving `values` and `themes` to figma.root
- [ ] We decompress when reading those
- [ ] We are able to differentiate between compressed values and non-compressed values, so a user using an older version of the plugin doesnt face any problems

akshay-gupta7:
### Why does this PR exist?

Closes #3328 

This PR compresses the data(mainly tokens and themes), using lz-string, while storing it in the localSharedPluginData of a user's file. While reading it decompresses it first.

Also, for legacy data, or users opening the plugin after a long time, it will still check whether their data is compressed or not, using a key added in this PR called `isCompressed`, which is also stored in the sharedPluginData, which once set to true,i.e the user is now storing compressed data, cannot be set to false again

### Testing this change

- Testing this is tricky and not straightforward
- Pull this branch, open the plugin, preferably with data and check if it opens fine
- then run this on the console to see if the data is compressed or not
- `figma.root.getSharedPluginData('tokens','themes');` - should return unreadable data(indicating that the string has been compressed), before pulling the branch it should return normal readable token values
- `figma.root.getSharedPluginData('tokens','values');`- should return unreadable data(indicating that the string has been compressed), before pulling the branch it should return normal readable themes dataa
- `figma.root.getSharedPluginData('tokens','isCompressed');` - should return empty/null before opening the plugin, and once plugin is opened, always `true`

### DIFF
--- a/.changeset/metal-mangos-fold.md
+++ b/.changeset/metal-mangos-fold.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+We now compress Figma's shared plugin data for tokens and themes values stored on the document, significantly reducing file sizes for very large token sets.
--- a/packages/tokens-studio-for-figma/package.json
+++ b/packages/tokens-studio-for-figma/package.json
@@ -106,6 +106,7 @@
     "launchdarkly-js-client-sdk": "^2.24.2",
     "launchdarkly-react-client-sdk": "^2.28.0",
     "lodash.debounce": "^4.0.8",
+    "lz-string": "^1.5.0",
     "mixpanel-figma": "^2.0.1",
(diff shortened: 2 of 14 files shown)


six7 (approved):
code-wise seems good - can we get it functionally tested? its such core functionality i'd want a few eyes on that

also, add a changeset - can say something like `We now compress shared plugin data for token values and themes values stored on the document to reduce file sizes for very large token sets`

PR #3339, opened 2025-05-07
Chunk Themes and Values when they cross 95KB in storage

akshay-gupta7:
### Why does this PR exist?

While using SharedPluginData in Figma, users currently have the freedom to store any amount of data.
However, starting May 13th, figma will impose a data limit of 100KB per key in sharedPluginData.

As a workaround we decided to chunk the themes and values into multiple keys while writing if their compressed size, surpasses 95KB, so for every 95KB, there will be a key generated like themes_chunk_0, themes_chunk_1, values_chunk_0, values_chunk_1, and so on. 
Also, added keys like themes_meta and values_meta will store the amount of chunks that have been created so that it is easy while reading and combining the same data.

### Testing this change

- Switch to local storage
- Load a token set/themes with compressed size greater than 100KB
- Then go to Figma dev console, type `figma.root.getSharedPluginDataKeys('tokens');`
- See the chunks that were created , like themes_chunk_0, themes_chunk_1, etc. or values_chunk and so on
- Also, confirm from the `values_meta` and the `themes_meta` keys that the count of chunks is correct, in that case chunking is happening accurately

### DIFF
--- a/packages/tokens-studio-for-figma/src/figmaStorage/FigmaStorageProperty.ts
+++ b/packages/tokens-studio-for-figma/src/figmaStorage/FigmaStorageProperty.ts
@@ -134,22 +134,22 @@ export class FigmaStorageProperty<V = string> {
           }
 
           if (metadata.type === 'chunked' && metadata.count) {
-            // If it's chunked, read and reassemble all chunks
-            const chunks: string[] = [];
-
-            for (let i = 0; i < metadata.count; i += 1) {
+            // Create an array of promises to read all chunks
+            const chunkPromises = Array.from({ length: metadata.count }, (_, i) => {
               const chunkKey = `${key}${CHUNK_SUFFIX_PREFIX}${i}`;
-              const chunk = node?.getSharedPluginData(namespace, chunkKey);
+              return node?.getSharedPluginData(namespace, chunkKey);
+            });
 
-              if (!chunk) {
-                console.error(`Missing chunk ${i} of ${metadata.count} for key ${key}`);
-                return null; // Missing chunk
-              }
+            // Wait for all chunks to be read
+            const chunks = await Promise.all(chunkPromises);
 
-              chunks.push(chunk);
+            // Check if any chunks are missing
+            if (chunks.some((chunk) => !chunk)) {
+              console.error('One or more chunks are missing');
+              return null;
             }
 
-            // Reassemble the chunks
+            // Join chunks and parse
             const value = chunks.join('');
             return value ? this.parse(value, isCompressed) : null;
           }
--- a/packages/tokens-studio-for-figma/src/figmaStorage/FigmaStorageProperty.ts
+++ b/packages/tokens-studio-for-figma/src/figmaStorage/FigmaStorageProperty.ts
@@ -1,61 +1,10 @@
-import { getUTF16StringSize } from '@/utils/getUTF16StringSize';
+import { readSharedPluginData } from '@/utils/figmaStorage/readSharedPluginData';
+import { writeSharedPluginData } from '@/utils/figmaStorage/writeSharedPluginData';
 
 export enum FigmaStorageType {
   CLIENT_STORAGE = 'client_storage',
   SHARED_PLUGIN_DATA = 'shared_plugin_data',
 }
-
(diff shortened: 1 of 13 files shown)


packages/tokens-studio-for-figma/src/figmaStorage/FigmaStorageProperty.ts
```diff
@@ -3,6 +3,85 @@ export enum FigmaStorageType {
...
+  }
+}
+
+/**
+ * Splits a string into chunks of specified maximum byte size
+ * @param str String to split
+ * @param maxBytes Maximum bytes per chunk
+ * @returns Array of string chunks
+ */
+function splitIntoChunks(str: string, maxBytes: number): string[] {
```
six7, line 48: can you extract this to its own file and add tests for this?

packages/tokens-studio-for-figma/src/figmaStorage/FigmaStorageProperty.ts
```diff
@@ -32,21 +117,173 @@ export class FigmaStorageProperty<V = string> {
...
+   * @param isCompressed Whether the data is compressed
+   */
   public async write(value: V | null, node: BaseNode = figma.root, isCompressed = false) {
     if (this.storageType === FigmaStorageType.CLIENT_STORAGE) {
       await figma.clientStorage.setAsync(this.key, value ? this.stringify(value, isCompressed) : null);
-    } else if (this.storageType === FigmaStorageType.SHARED_PLUGIN_DATA) {
+      return;
+    }
+
+    if (this.storageType === FigmaStorageType.SHARED_PLUGIN_DATA) {
```
six7, line 185: can we move this write logic into its own utility file and add tests?

packages/tokens-studio-for-figma/src/figmaStorage/FigmaStorageProperty.ts
```diff
@@ -32,21 +117,173 @@ export class FigmaStorageProperty<V = string> {
       const keyParts = this.key.split('/');
       const namespace = keyParts[0];
       const key = keyParts.slice(1).join('/');
+
+      // First, check if there's metadata for chunked data
+      const metaKey = `${key}${METADATA_SUFFIX}`;
```
six7, line 122: can we move this read logic into its own utility file and add tests?

packages/tokens-studio-for-figma/src/figmaStorage/__tests__/ValuesProperty.test.ts
```diff
@@ -17,14 +17,54 @@ describe('ValuesProperty', () => {
...
   it('should be able to read', async () => {
-    mockRootGetSharedPluginData.mockReturnValueOnce(JSON.stringify(mockValues));
-    expect(await ValuesProperty.read()).toEqual(mockValues);
+    // Mock metadata indicating single storage
+    mockRootGetSharedPluginData.mockReturnValueOnce(JSON.stringify({ type: 'single' }));
+
+    // Mock the actual data
+    mockRootGetSharedPluginData.mockReturnValueOnce(compressedMockValues);
+
+    expect(await ValuesProperty.read(figma.root, true)).toEqual(mockValues);
```
six7, line 47: we should add tests here to verify the chunking logic works (as in when there's chunks, it reads them and outputs the expected

packages/tokens-studio-for-figma/src/figmaStorage/__tests__/ValuesProperty.test.ts
```diff
@@ -17,14 +17,54 @@ describe('ValuesProperty', () => {
...
   it('should be able to write', async () => {
+    // Mock metadata check for single storage
+    mockRootGetSharedPluginData.mockReturnValueOnce(null);
+
     await ValuesProperty.write(mockValues);
-    expect(mockRootSetSharedPluginData).toBeCalledTimes(1);
-    expect(mockRootSetSharedPluginData).toBeCalledWith('tokens', 'values', compressedMockValues);
+
+    // Should set metadata first (indicating single storage)
+    expect(mockRootSetSharedPluginData).toHaveBeenCalledWith(
```
six7, line 31: we should add a test here for the chunking part

six7 (approved):
got some test comments that i will address myself 👍

PR #3354, opened 2025-05-19
update error messages on studio sync

### LINKED ISSUE #3304: Studio sync: If it fails to pull, show an error
If we fail to pull from Studio when the user presses the Pull and we are unable to pull, show an error.

Today we silently fail but show an error in console

We should also do that when you are setting up a new storage provider - right now we just show the `{}` in the error field 😬  e.g. if the token set couldnt be retrieved successfully

akshay-gupta7:
### Why does this PR exist?

Closes #3304 

### Additional Notes (if any)

Screenshot 2025-05-27 at 2 13 03 PM
Screenshot 2025-05-27 at 2 13 32 PM

### DIFF
--- a/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
@@ -59,6 +59,7 @@ async function getProjectData(id: string, orgId: string, client: any): Promise<P
     });
 
     if (!data.data?.project?.branch) {
+      notifyToUI('Project or branch not found in Tokens Studio', { error: true });
       return null;
     }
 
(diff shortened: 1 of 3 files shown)


packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -59,6 +59,7 @@ async function getProjectData(id: string, orgId: string, client: any): Promise<P
     });
 
     if (!data.data?.project?.branch) {
+      notifyToUI('Project or branch not found in Tokens Studio', { error: true });
```
six7, line 62: does this also show up in the error message users receive on the Settings screen? The one that says `{}` today?

packages/tokens-studio-for-figma/src/storage/TokensStudioTokenStorage.ts
```diff
@@ -118,7 +119,26 @@ async function getProjectData(id: string, orgId: string, client: any): Promise<P
...
+      } else if (e.message.includes('Not Found') || e.message.includes('404')) {
+        errorMessage = 'Project or organization not found in Tokens Studio.';
+      } else if (e.message.includes('Permission') || e.message.includes('403')) {
+        errorMessage = 'You do not have permission to access this Tokens Studio project.';
+      } else {
+        errorMessage = `Error connecting to Tokens Studio: ${e.message.substring(0, 100)}`;
+      }
+    }
+
+    notifyToUI(errorMessage, { error: true });
```
six7, line 141: same message as above - the issue itself was about the {} error message that we get when setting up a new provider but something fails - or if connecting to a token set - can we (in addition to the notifyToUI) also make sure we display it there?

six7:
i dont think you addressed my prior comments - can you please add images to the PR that show the new functionality of the error showing up in the UI? i dont think right now it would show up in the error that appears (instead of the {} that we have today, which the issue is about)

six7 (approved):

PR #3396, opened 2025-05-29
Fix/3227

### LINKED ISSUE #3227: Tokens applying as Variables instead of Tokens
**Describe the bug**
When you export certain token types as Variables and try to apply it as a token to an element in Figma, the Variable is applied instead of the Token.
The Variable icon should be shown when a Variable is directly applied from Figma. The same accounts for the styles icon (4 dots).
In my tests I noticed strange behavior for the token types: Border Radius, Border Width and the Dimension token (all can be applied to different sides/places, so that is probably related to the issue).

**To Reproduce**
1. Create tokens and export these as Variables (see screenrecording)
2. Try to apply the tokens to an element and check the Inspect panel afterwards.
3. See that Variables in some cases are applied instead of the tokens.

**Note**
Please also check other token types to see if I didn't miss any type.

**Expected behavior**
These Tokens should be applied as Token and not as Variable (so the Variable icon shouldn't show up in the inspect tab).

https://github.com/user-attachments/assets/5787c215-9671-428b-a9d1-76fc53d8d76d

**Figma file (optional)**
https://www.figma.com/design/p4GCFwId9x3mRGvMjD8dG2/Tokens-applying-as-Variables-instead-of-Tokens?node-id=0-1&t=b0GGuh5dFKgUFpHv-1

akshay-gupta7:
### Why does this PR exist?

Closes #3227 

Currently, when users apply border radius/border width on a layer in Figma, it is applied as variable when we apply as token from our plugin, when we inspect in our Inspect Panel. Tokens should be displayed as variables when we apply a variable from Figma directly.
This PR addresses that, when we apply tokens as token, it displays as token, and when we apply as variables, it displays as variable.

### Testing this change

- Export tokens as variables
- Apply border radius/border width as token
- Check inspect panel on the layer, it should be displayed as tokens
- Now create another layer and apply it as variables exported directly from Figma
- It should now reflect the applied variable in the Inspect panel in our plugin.

### Additional Notes (if any)

https://github.com/user-attachments/assets/c1837aa0-5d62-4eee-bb9f-9baf23886d29

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/applyBorderRadiusValuesOnNode.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/applyBorderRadiusValuesOnNode.ts
@@ -23,16 +23,7 @@ export async function applyBorderRadiusValuesOnNode(
       switch (individualBorderRadius.length) {
         case 1:
           if ('cornerRadius' in node) {
-            if (
-              !(
-                (await tryApplyVariableId(node, 'topLeftRadius', data.borderRadius))
-                && (await tryApplyVariableId(node, 'topRightRadius', data.borderRadius))
-                && (await tryApplyVariableId(node, 'bottomRightRadius', data.borderRadius))
-                && (await tryApplyVariableId(node, 'bottomLeftRadius', data.borderRadius))
-              )
-            ) {
-              node.cornerRadius = transformValue(String(values.borderRadius), 'borderRadius', baseFontSize);
-            }
+            node.cornerRadius = transformValue(String(values.borderRadius), 'borderRadius', baseFontSize);
           }
           break;
         case 2:
--- a/packages/tokens-studio-for-figma/src/plugin/applyBorderRadiusValuesOnNode.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/applyBorderRadiusValuesOnNode.ts
@@ -23,7 +23,16 @@ export async function applyBorderRadiusValuesOnNode(
       switch (individualBorderRadius.length) {
         case 1:
           if ('cornerRadius' in node) {
-            node.cornerRadius = transformValue(String(values.borderRadius), 'borderRadius', baseFontSize);
+            if (
+              !(
+                (await tryApplyVariableId(node, 'topLeftRadius', data.borderRadius))
+                && (await tryApplyVariableId(node, 'topRightRadius', data.borderRadius))
+                && (await tryApplyVariableId(node, 'bottomRightRadius', data.borderRadius))
+                && (await tryApplyVariableId(node, 'bottomLeftRadius', data.borderRadius))
+              )
+            ) {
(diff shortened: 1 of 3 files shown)


packages/tokens-studio-for-figma/src/plugin/applyBorderRadiusValuesOnNode.ts
```diff
@@ -23,16 +23,7 @@ export async function applyBorderRadiusValuesOnNode(
...
-              !(
-                (await tryApplyVariableId(node, 'topLeftRadius', data.borderRadius))
-                && (await tryApplyVariableId(node, 'topRightRadius', data.borderRadius))
-                && (await tryApplyVariableId(node, 'bottomRightRadius', data.borderRadius))
-                && (await tryApplyVariableId(node, 'bottomLeftRadius', data.borderRadius))
-              )
-            ) {
-              node.cornerRadius = transformValue(String(values.borderRadius), 'borderRadius', baseFontSize);
-            }
+            node.cornerRadius = transformValue(String(values.borderRadius), 'borderRadius', baseFontSize);
```
six7: this completely removes the functionality to have border radius tokens applied as variables? we _want_ tokens to be applied as variables if they exist, so i dont see why this was removed

akshay-gupta7: when I tried to apply it and it gave me an error for corner radius, because of this:
```
type VariableBindableNodeField =
  | 'height'
  | 'width'
  | 'characters'
  | 'itemSpacing'
  | 'paddingLeft'
  | 'paddingRight'
  | 'paddingTop'
  | 'paddingBottom'
  | 'visible'
  | 'topLeftRadius'
  | 'topRightRadius'
  | 'bottomLeftRadius'
  | 'bottomRightRadius'
  | 'minWidth'
  | 'maxWidth'
  | 'minHeight'
  | 'maxHeight'
  | 'counterAxisSpacing'
  | 'strokeWeight'
  | 'strokeTopWeight'
  | 'strokeRightWeight'
  | 'strokeBottomWeight'
  | 'strokeLeftWeight'
  | 'opacity'
```
cornerRadius or border radius doesn't exist on this:
https://www.figma.com/plugin-docs/api/VariableBindableNodeField/

six7: yes they do: 
```
  | 'topLeftRadius'
  | 'topRightRadius'
  | 'bottomLeftRadius'
  | 'bottomRightRadius'
```

akshay-gupta7: reverted in [6150981](https://github.com/tokens-studio/figma-plugin/pull/3396/commits/6150981b50705968faa0756b012a2568e5fe3ceb)

six7 (approved):

PR #3423, opened 2025-06-12
Fix load tokens when imported

### LINKED ISSUE #3408: Tokens Lost if you don't click Apply To when using Local Document
**Describe the bug**
There seems to be an issue when working on `local document`, where if you load tokens into the plugin, close and reopen it, the plugin doesn't save the tokens to the document. The plugin only saves them locally if you specifically click the `Apply To` button after loading tokens. 

**Expected behavior**
The plugin shouldn't force users to click the `Apply To` after loading tokens as that is not its functionality. 

TIMESTAMP 2:30
https://github.com/user-attachments/assets/7fd76204-006b-4216-92c1-8caffa32dda8

akshay-gupta7:
### Why does this PR exist?

Resolves #3408

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/store/models/effects/tokenState/setDefaultTokens.ts
+++ b/packages/tokens-studio-for-figma/src/app/store/models/effects/tokenState/setDefaultTokens.ts
@@ -12,7 +12,11 @@ export function setDefaultTokens(dispatch: RematchDispatch<RootModel>) {
       themes: [],
       activeTheme: {},
       usedTokenSet: { core: TokenSetStatus.SOURCE, light: TokenSetStatus.ENABLED, theme: TokenSetStatus.ENABLED },
-      shouldUpdate: true,
+    });
+
+    dispatch.tokenState.updateDocument({
+      shouldUpdateNodes: false,
+      updateRemote: false,
     });
   };
 }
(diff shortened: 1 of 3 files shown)


packages/tokens-studio-for-figma/src/app/store/models/effects/tokenState/setDefaultTokens.ts
```diff
@@ -12,6 +12,7 @@ export function setDefaultTokens(dispatch: RematchDispatch<RootModel>) {
       themes: [],
       activeTheme: {},
       usedTokenSet: { core: TokenSetStatus.SOURCE, light: TokenSetStatus.ENABLED, theme: TokenSetStatus.ENABLED },
+      shouldUpdate: true,
```
six7: this only partially solves the issue.

there was a reason this isnt included in the setDefaultTokens call, and that is.. if you load tokens in a huge document and you'd have "apply to document" set, it means that on load of tokens it will run a document wide change. which can take quite a long time and isnt expected.

intead of just changing shouldUpdate to true - we should rather send an atomic update that just sets the new token values, and doesnt perform the whole other orchestration that comes through `shouldUpdate: true`

six7: you can reproduce this by:
- load the default tokens
- change a token from those (eg black to red)
- create a rectangle and apply that token you just changed
- create 1.000 replicas of that rectangle
- change mode to "Apply to document"
- Load the default tokens again.

Notice how it will - in this PR - make changes to aaaaaalll rectangles - without you ever requesting it. that's what `shouldUpdate` will do.

PR #3426, opened 2025-06-17
Fix plugin crashing when a huge token set is exported to variables

### LINKED ISSUE #3425: Plugin crashes when exporting variables from large token sets
**Describe the bug**
The plugin crashes when attempting to export a large token set to create variables in Figma. This seems to happen consistently when the token set is significantly large, potentially due to Figma's memory allocation limits which leads to abrupt crashes.

**To Reproduce**
- Open the plugin in a Figma file.
- Export a  large token set (e.g., 1000+ tokens).
- Plugin becomes unresponsive and/or crashes.

**Expected behavior**

**Figma file (optional)**

**JSON (optional)**

akshay-gupta7:
### Why does this PR exist?

Closes #3425 
We are observing that when users are attempting to export sets with huge tokens, (typically >1000 atleast), the plugin is freezing and then crashing.
Error logs point to a memory leakage on Figma's end. This PR addresses our approach to handle such huge sets.

This PR changes the approach to manage such huge tokens in batches of 100, instead of processing all at one, if token count is above 1000.

### Testing this change

- Export a  large token set (e.g., 1000+ tokens).
- This should now display progress bar and the plugin shouldn't crash, due to the memory leakage issue

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
@@ -30,12 +30,12 @@ async function processBatch(
   referenceVariableCandidates: ReferenceVariableType[],
   renamedVariableKeys: string[],
 ) {
-  await Promise.all(batch.map(async (token) => {
+  return Promise.all(batch.map(async (token) => {
     const variableType = convertTokenTypeToVariableType(token.type, token.value);
     // If id matches the variableId, or name patches the token path, we can use it to update the variable instead of re-creating.
     // This has the nasty side-effect that if font weight changes from string to number, it will not update the variable given we cannot change type.
     // In that case, we should delete the variable and re-create it.
-    const variable = variablesInFigma.find((v) => (v.key === token.variableId && !v.remote) || v.name === token.path) || figma.variables.createVariable(token.path, collection, variableType);
+    const variable = variablesInFigma.find((v) => (v.key === token.variableId && !v.remote) || v.name === token.path) ?? figma.variables.createVariable(token.path, collection, variableType);
 
     if (variable) {
       // First, rename all variables that should be renamed (if the user choose to do so)
@@ -116,12 +116,11 @@ export default async function setValuesOnVariable(
   // Process tokens in batches to prevent memory leaks with large token sets
   // Use smaller batch sizes for very large token sets to prevent memory issues
   const totalTokens = tokens.length;
-  let BATCH_SIZE = 100;
-  if (totalTokens > 10000) {
-    BATCH_SIZE = 50;
-  } else if (totalTokens > 5000) {
-    BATCH_SIZE = 75;
-  }
+  // clamp batch size between 100 and 50, for tiny sets -> 100, for sets>10000 -> 50
+  const BATCH_SIZE = Math.round(Math.min(Math.max(100 - totalTokens / 200, 50), 100));
+
+  // Calculate delay based on total tokens: totalTokens / 200, clamped between 10 and 50ms
+  const DELAY = Math.max(10, Math.min(50, Math.round(totalTokens / 200)));
 
   const tracker = new ProgressTracker(BackgroundJobs.PLUGIN_UPDATEPLUGINDATA);
 
@@ -158,16 +157,10 @@ export default async function setValuesOnVariable(
       }
       tracker.reportIfNecessary();
 
-      // For very large token sets, add longer delays to prevent memory issues
+      // Add delay between batches to prevent memory issues
       if (i + BATCH_SIZE < totalTokens) {
-        let delay = 10;
-        if (totalTokens > 10000) {
-          delay = 50;
-        } else if (totalTokens > 5000) {
-          delay = 25;
-        }
         await new Promise<void>((resolve) => {
-          setTimeout(resolve, delay);
+          setTimeout(resolve, DELAY);
         });
       }
     }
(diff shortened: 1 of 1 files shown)


packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
```diff
@@ -7,13 +7,100 @@ import { convertTokenTypeToVariableType } from '@/utils/convertTokenTypeToVariab
...
+  variablesInFigma: Variable[],
+  collection: VariableCollection,
+  mode: string,
+  baseFontSize: string,
+  shouldRename: boolean,
+  variableKeyMap: Record<string, string>,
+  referenceVariableCandidates: ReferenceVariableType[],
+  renamedVariableKeys: string[],
+) {
+  await Promise.all(batch.map(async (token) => {
```
jorenbroekema:
```suggestion
  return Promise.all(batch.map(async (token) => {
```
since we don't need to await for anything inside the scope of this function, we can just return it, which returns the Promise that's a result of Promise.all()

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
```diff
@@ -7,13 +7,100 @@ import { convertTokenTypeToVariableType } from '@/utils/convertTokenTypeToVariab
...
+    const variableType = convertTokenTypeToVariableType(token.type, token.value);
+    // If id matches the variableId, or name patches the token path, we can use it to update the variable instead of re-creating.
+    // This has the nasty side-effect that if font weight changes from string to number, it will not update the variable given we cannot change type.
+    // In that case, we should delete the variable and re-create it.
+    const variable = variablesInFigma.find((v) => (v.key === token.variableId && !v.remote) || v.name === token.path) || figma.variables.createVariable(token.path, collection, variableType);
```
jorenbroekema:
```suggestion
    const variable = variablesInFigma.find((v) => (v.key === token.variableId && !v.remote) || v.name === token.path) ?? figma.variables.createVariable(token.path, collection, variableType);
```
prefer nullish coalescing for fallbacks, so you don't end up using a fallback for a valid falsy value like 0 or ''

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
```diff
@@ -25,78 +112,84 @@ export default async function setValuesOnVariable(
...
+
+  // Process tokens in batches to prevent memory leaks with large token sets
+  // Use smaller batch sizes for very large token sets to prevent memory issues
+  const totalTokens = tokens.length;
+  let BATCH_SIZE = 100;
+  if (totalTokens > 10000) {
+    BATCH_SIZE = 50;
+  } else if (totalTokens > 5000) {
+    BATCH_SIZE = 75;
+  }
```
jorenbroekema:
```suggestion
  // clamp batch size between 100 and 50, for tiny sets -> 100, for sets>10000 -> 50
  const BATCH_SIZE = Math.round(Math.min(Math.max(100 - totalTokens / 200, 50), 100));
```

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
```diff
@@ -25,78 +112,84 @@ export default async function setValuesOnVariable(
...
+      }
+      tracker.reportIfNecessary();
+
+      // For very large token sets, add longer delays to prevent memory issues
+      if (i + BATCH_SIZE < totalTokens) {
+        let delay = 10;
+        if (totalTokens > 10000) {
+          delay = 50;
+        } else if (totalTokens > 5000) {
+          delay = 25;
```
jorenbroekema: Put this logic near the same LOC that contains the batch size logic. DELAY should be a constant just like BATCH_SIZE.
You can also use a clamp math function for this, see my other comment for an example. `totalTokens / 200`, rounded, clamped between 10 and 50.

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
```diff
@@ -25,78 +112,84 @@ export default async function setValuesOnVariable(
...
-    }));
+    }
+
+    postToUI({
+      type: MessageFromPluginTypes.COMPLETE_JOB,
+      name: BackgroundJobs.PLUGIN_UPDATEPLUGINDATA,
+    });
+
+    if (totalTokens > 1000) {
+      notifyUI(`Successfully processed ${totalTokens} tokens!`, { timeout: 2000 });
```
jorenbroekema: We don't notify for small tokensets? why not?

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
```diff
@@ -25,78 +112,75 @@ export default async function setValuesOnVariable(
...
-          referenceVariableCandidates.push({
-            variable,
-            modeId: mode,
-            referenceVariable: referenceTokenName,
-          });
-        }
+    for (let i = 0; i < totalTokens; i += BATCH_SIZE) {
+      const batch = tokens.slice(i, i + BATCH_SIZE);
+
+      await processBatch(
```
jorenbroekema, line 143: Would it still work if you parallelized this? Rather than 1 batch waiting for the previous batch to complete? Or does that defeat the purpose of this PR which is to actually sequence things in batches to avoid the memory issues?
Might be good to add a comment if so, so someone in the future doesn't parallelize this and cause a regression of the bug

jorenbroekema (approved):

packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts
akshay-gupta7: sure, I will add that comment

PR #3427, opened 2025-06-18
FIx crashing of plugin when importing styles with lots of variable references

### LINKED ISSUE #3353: Plugin Crashes on Style Import with Missing Variables
**Bug: Plugin Crashes During Style Import Due to Missing Variables**

When attempting to import styles in version 2.5.1 of the plugin, the process crashes if certain referenced variables do not exist.

**Steps to Reproduce:**
1. Attempt to import styles from the provided Figma file link.
2. Ensure that the styles reference variables that may not exist in the file.

**Expected Behavior:**
The plugin should successfully import the styles, disregarding any missing variables without crashing the entire process.

**Actual Behavior:**
The import process crashes when it encounters a missing variable, for example, `font-size/54`, which causes the plugin to break.

**Suggested Fix:**
Implement a try/catch mechanism during the reference lookup; if a variable is not found, the plugin should use the raw value from the style instead of terminating the process.

Slack thread: https://hyma-team.slack.com/archives/C04V8U8UH0F/p1747199683619199?thread_ts=1747199683.619199&cid=C04V8U8UH0F

## Attachments

📷 **Image attachment**: [CleanShot 2025-05-14 at 06.14.32@2x.png](https://hyma-team.slack.com/files/U0692JZALKC/F08SBKUF86N/cleanshot_2025-05-14_at_06.14.32_2x.png) (PNG, 160 KB)

📷 **Image attachment**: [Screenshot 2025-05-16 at 7.53.05 PM.png](https://hyma-team.slack.com/files/U0692JZALKC/F08SPBNBSR3/screenshot_2025-05-16_at_7.53.05___pm.png) (PNG, 1240 KB)

<details>
<summary>Show all messages</summary>
marco.krenn wrote: @akshay i just tried importing styles with 2.5.1 and the plugin crashed with

akshay wrote: @marco.krenn if its a test file or a duplicate one if its production, can I have access to those, because I am being able to import styles, maybe there is a specific type crashing it

six.jan wrote: same, able to import color,text,shadow

akshay wrote: yeah

marco.krenn wrote: trying to export the file it's in another org so need to bring it to us

marco.krenn wrote: @akshay @six.jan this is the file
<https://www.figma.com/design/zMeQRBqw5YppHPrLdaVnRc/Foundation---Token-Studio--Copy----Partial-file-saved-15-05-2025?node-id=12-2&amp;p=f&amp;t=bV0aftZsYAqDNkr0-0|https://www.figma.com/design/zMeQRBqw5YppHPrLdaVnRc/Foundation---Token-Studio--Copy[…]al-file-saved-15-05-2025?node-id=12-2&amp;p=f&amp;t=bV0aftZsYAqDNkr0-0>

marco.krenn wrote: @akshay did you have some time to look into the import on that file?

akshay wrote: yes, just checked, there is no issue with the styles

akshay wrote: there is some issue with the variables which is hijacking the import process

akshay wrote: for example this variable referenced in the style font size is `font-size/54` which does not exist in the file

marco.krenn wrote: oh, but this should not break the plugin when we only import the style, but i guess it tries to make a lookup

akshay wrote: yes you are right it should not be breaking the entire process

six.jan wrote: oh got it, so the issue is that a single variable currently - if it does not exist - causes the whole thing to break? agree, we should be more resilient and try/catch the reference lookup, if no variable is found, just use the raw value (which AFAIK should be on the style still?)

six.jan wrote: @issue_assistant create issue in tokens-studio/figma-plugin
</details>

akshay-gupta7:
### Why does this PR exist?

Closes #3353 

Currently, when we import styles with a lot of variable references, it breaks down because of memory limits in the Figma sandbox

This PR runs those imports in batches to avoid said breakdowns.

### Testing this change

- Open a figma file with lots of variables, and a lot of which would be connected to styles
- Import the styles into our plugin to create tokens
- The styles should be imported smoothly and the references of variables would also be created correctly.
- Also added a progress bar to display the user the state of the styles' import

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/LoadingBar.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/LoadingBar.tsx
@@ -26,6 +26,7 @@ export const backgroundJobTitles = {
   [BackgroundJobs.UI_ATTACHING_LOCAL_STYLES]: 'Attaching local styles to theme...',
   [BackgroundJobs.UI_CREATEVARIABLES]: 'Creating variables...',
   [BackgroundJobs.UI_CREATE_STYLES]: 'Creating styles...',
+  [BackgroundJobs.UI_IMPORT_STYLES]: 'Importing styles...',
 };
 
 export default function LoadingBar() {
--- a/packages/tokens-studio-for-figma/src/constants/BackgroundJobs.ts
+++ b/packages/tokens-studio-for-figma/src/constants/BackgroundJobs.ts
@@ -21,4 +21,5 @@ export enum BackgroundJobs {
   UI_RENAME_TOKEN_ACROSS_SETS = 'ui_rename_token_across_sets',
   UI_CREATEVARIABLES = 'ui_create_variables',
   UI_CREATE_STYLES = 'ui_create_styles',
+  UI_IMPORT_STYLES = 'ui_import_styles',
 }
--- a/packages/tokens-studio-for-figma/src/plugin/processTextStyleProperty.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/processTextStyleProperty.ts
@@ -16,27 +16,148 @@ export function processTextStyleProperty(
   const boundVariables = style.boundVariables as Record<string, { id: string; } | undefined>;
   if (boundVariables?.[propertyKey]?.id) {
     const variable = localVariables.find((v) => v.id === boundVariables[propertyKey]?.id);
-    if (variable && tokens) {
+    if (variable) {
       const normalizedName = variable.name.replace(/\//g, '.');
 
-      // Look for an existing token with this name
-      const existingToken = Object.entries(tokens.values).reduce<SingleToken | null>((found, [_, tokenSet]) => {
-        if (found) return found;
-        const foundToken = Array.isArray(tokenSet) ? tokenSet.find((token) => typeof token === 'object'
-          && token !== null
-          && 'name' in token
-          && token.name === normalizedName) : null;
-        return foundToken || null;
-      }, null);
+      // Get the variable's actual value and transform it if needed
+      console.log(`Processing variable ${normalizedName} for property ${propertyKey}`);
+      console.log('Variable object:', JSON.stringify(variable, null, 2));
+      console.log('Variable valuesByMode:', JSON.stringify(variable.valuesByMode, null, 2));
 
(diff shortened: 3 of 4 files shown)


anathaniel-TS:
🛡️ **Hyma Compliance Check**

4 file(s) scanned · 0 findings

anathaniel-TS:
🛡️ **Hyma Compliance Check**

4 file(s) scanned · 0 findings

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **1 finding(s) across 1 file(s)**

| 🟢 LOW | Policy Gap | `packages/tokens-studio-for-figma/src/plugin/pullStyles.ts` | Batch processing swallows errors silently, so partial style imports complete with no log or user signal |

<details>

**[Policy Gap] Batch processing swallows errors silently, so partial style imports complete with no log or user signal**
- **File**: `packages/tokens-studio-for-figma/src/plugin/pullStyles.ts` (lines ~40-78 `processStylesWithVariablesInBlocks`, lines ~80-115 `processTypographyInBlocks`)
- **Implication**: Both new block processors wrap each batch in `try { … } catch (error) { /* Continue with next block instead of failing completely */ }` with an empty body. If a block throws, the remaining styles in that block are dropped and the loop moves on: the import reports success, the progress bar still advances to `COMPLETE_JOB`, and neither the user nor any log records that the result is incomplete. That is a loss of integrity signal on a data-transformation path (SOC2 CC7.2 monitoring / PI1.4 completeness), and it makes a silently truncated token set hard to distinguish from a correct one during later investigation. The same swallowed `catch` was flagged independently by the Copilot reviewer (`#discussion_r2167493182`). No personal data is involved — the payload is design tokens and style metadata.
- **Recommended action**: Before or after merging, log the caught error (e.g. `console.error` with the block start index, or route through the existing plugin error notifier) and surface a warning to the UI when any block fails, so an incomplete import is visible rather than silent.

Scope note: the new `BackgroundJobs.UI_IMPORT_STYLES` job and the `postToUI` progress messages in `pullStyles.ts` use the internal Figma plugin↔iframe `postMessage` channel only — no external host, no analytics or Sentry event, and no new dependency. `processTextStyleProperty.ts` reads `variable.valuesByMode` and falls back to style values entirely in-memory. No storage APIs, credential handling, entitlement gating or personal-data fields are touched.

</details>

PR #3476, opened 2025-07-30
Fix import text styles

### LINKED ISSUE #3458: Importing styles with variable references on font style/weight not working
**Describe the bug**
When we import text styles and they have a variable reference in font weight/style, they seem to be showing up as undefined

**To Reproduce**
1. Create a style with a variable reference in their font style and/or weight
2. In our plugin, Import styles
3. The corresponding tokens will be broken as the font style/weight will show up as undefined

**Expected behavior**
Ideally it should create a token for the referenced variable, and if it cannot find the variable, it should create a token with the resolved value and link it in the typography token

**Figma file (optional)**

**JSON (optional)**

akshay-gupta7:
### Why does this PR exist?

Closes #3458

### DIFF
--- a/.changeset/soft-needles-do.md
+++ b/.changeset/soft-needles-do.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fix issue of resolving font styles and font weight to the correct referenced variables when importing text styles into the plugin.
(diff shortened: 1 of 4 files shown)


six7:
trying to follow the steps i get this error

image

six7 (changes requested):
this feels like bandaids for a problem that is deeper, i'll do some investigations.

PR #3485, opened 2025-08-01
Add search input to Manage Themes dialog

six7:
Adds a search input to the Manage Themes dialog, for new themes and existing themes, letting users search for token sets.

image

image

image


akshay-gupta7 (commented):
it's working as expected, just test coverage is considerably reduced

PR #3496, opened 2025-08-13
fix azure issue

### LINKED ISSUE #3472: ADO sync weird issue for a user
**Describe the bug**
A plugin user who use Azure Devops to sync their tokens, is facing an issue of tokens being asked to be removed everytime when they pull, when in-fact, there is no tokens that have been removed

**To Reproduce**
Hard to reproduce for now

**Expected behavior**

**Figma file (optional)**

**JSON (optional)**

akshay-gupta7:
### Why does this PR exist?

Closes #3472

### DIFF
--- a/packages/tokens-studio-for-figma/src/storage/ADOTokenStorage.ts
+++ b/packages/tokens-studio-for-figma/src/storage/ADOTokenStorage.ts
@@ -92,18 +92,48 @@ export class ADOTokenStorage extends GitTokenStorage {
       ? Object.entries(params).reduce<string>((acc, [key, value]) => `${acc}${key}=${value}&`, '') + apiVersion
       : apiVersion;
     const input = `${orgUrl}/${projectId ? `${projectId}/` : ''}_apis/git/repositories/${repositoryId}/${gitResource}?${paramString}`;
-    const res = await fetch(
-      input,
-      {
-        method,
-        headers: {
-          'Content-Type': 'application/json',
-          Authorization: `Basic ${btoa(`:${token}`)}`,
-        },
-        body,
-      },
-    );
-    return res;
+
+    // Retry logic for Azure DevOps IP transition network issues
+    let lastError: any;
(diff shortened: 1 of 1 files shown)


packages/tokens-studio-for-figma/src/storage/GithubTokenStorage.ts
```diff
@@ -287,20 +309,26 @@ export class GithubTokenStorage extends GitTokenStorage {
...
+        changes: [
+          {
+            message,
+            files: changeset,
+            filesToDelete,
+            ignoreDeletionFailures,
+          },
+        ],
+      }),
+      {
```
six7, line 327: Instead of adding this everywhere lets make it the default, customizable but not something you have to add

packages/tokens-studio-for-figma/src/utils/retryWithBackoff.ts
```diff
@@ -0,0 +1,98 @@
+export interface RetryOptions {
```
six7, line 1: lets also add test coverage for this file.

you can also add a cypress test, like make sure that the first 2 attempts fail, and the third goes through - we'd expect the user to not see an error and everything look normal

packages/tokens-studio-for-figma/src/storage/GithubTokenStorage.ts
akshay-gupta7: resolved in https://github.com/tokens-studio/figma-plugin/pull/3496/commits/8e152d61285f529ef8489ad27129d0c9c6541f80

packages/tokens-studio-for-figma/src/utils/retryWithBackoff.ts
akshay-gupta7: will raise a separate PR for this, created a ticket to track
https://github.com/tokens-studio/figma-plugin/issues/3503

PR #3538, opened 2025-08-31
fix theme reoder

six7:


packages/tokens-studio-for-figma/cypress/e2e/manage-themes.cy.js
six7, line 1: @copilot lets not do the test twice, it doesnt test anything new. Instead, validate drag and drop is a actually working, see https://dev.to/samelawrence/how-i-implement-drag-and-drop-in-my-cypress-tests-ppg

PR #3548, opened 2025-09-04
Update on drop

six7:
## Summary

  Adds a new "Auto-apply theme on drop" setting that automatically applies the current
  theme's tokens when component instances are inserted into Figma documents.

##  What's New

  - New Setting: Added toggle in Settings dropdown (gear icon) to enable/disable
  auto-apply functionality
  - Automatic Token Application: When enabled, tokens are automatically applied to newly
  inserted component instances
  - Seamless Workflow: Eliminates the need to manually apply tokens after dropping
  components

##  Technical Implementation

  - Uses Figma's documentchange event listener to detect component instance creation
  - Clean message-passing architecture from plugin to UI thread
  - Respects user preferences through Redux state management
  - Only applies tokens when the setting is enabled

##  User Experience

  - Setting is disabled by default to maintain existing behavior
  - Located in Settings dropdown alongside other configurable options
  - Works with any theme that's currently active
  - Provides immediate feedback when dropping component instances

##  Files Changed

  - Added new Redux state and selectors for the setting
  - Implemented plugin-side event detection for instance creation
  - Added UI message handling and token application logic
  - Updated Settings dropdown with new toggle option
  - Added TypeScript types and proper error handling

This feature streamlines the design workflow by automatically applying theme tokens to
component instances, reducing manual steps and improving consistency across designs.


six7:
@copilot lint fails and tests fail, fix pls. Also make sure we add tests that validate this new behavior. And see if we can simplify, the change with the document change is the right one not sure if we have leftover stuff in the changes that we dont need.

packages/tokens-studio-for-figma/src/i18n/lang/en/tokens.json
```diff
@@ -84,6 +84,10 @@
     "shouldUpdateStyles": {
       "title": "Update Figma Styles on apply",
       "description": "Updates local Figma Styles when themes change or whenever applied"
+    },
+    "autoApplyThemeOnDrop": {
```
six7, line 88: @copilot also generate translations for the other languages

packages/tokens-studio-for-figma/src/app/components/SettingsDropdown.tsx
```diff
@@ -99,6 +104,19 @@ export default function SettingsDropdown() {
...
+            checked={autoApplyThemeOnDrop}
+            onCheckedChange={handleAutoApplyThemeOnDrop}
+          >
+            <DropdownMenu.ItemIndicator>
+              <Check />
+            </DropdownMenu.ItemIndicator>
+            {t('update.autoApplyThemeOnDrop.title')}
+            <Box css={{ color: '$fgMuted', fontSize: '$xxsmall' }}>
+              {t('update.autoApplyThemeOnDrop.description')}
+            </Box>
```
akshay-gupta7, line 118: should we disable this for non pro users

six7: no, this is a consumer-faced feature, and as such we make it available to non-pro users

PR #3550, opened 2025-09-04
fix persistent push indicator issue

### LINKED ISSUE #3522: "Push" button indicator persists after pulling changes
**Describe the bug**
The blue dot indicator on the "Push" button, which signals that local changes are ready to be pushed, does not disappear after a user pulls the latest updates from the remote repository via the GitHub provider.

**To Reproduce**
1. Make changes 
2. Push to Github
3. Pull from Github

**Expected behavior**
Pulling changes should discard local (config) changes or override them.

https://github.com/user-attachments/assets/b9cccd55-58a9-41a9-bcfe-66e15ce7b65e

**Figma file (optional)**

**JSON (optional)**

akshay-gupta7:
### Why does this PR exist?

Closes #3522  

The blue dot indicator was incorrectly showing that there were always changes to push to the remote repository, even when the local and remote states were actually identical. This happened because of a format mismatch in how themes were being compared

This PR fixes the persistent blue dot by ensuring the lastSyncedState uses the same cleaned theme format (excluding disabled token sets) that's used in current state comparisons, eliminating the format mismatch that caused false change detection

### Testing this change

- Rename a token set, included in a theme
- Push the change
- the blue dot should be gone now

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/store/models/reducers/tokenState/setTokenData.ts
+++ b/packages/tokens-studio-for-figma/src/app/store/models/reducers/tokenState/setTokenData.ts
@@ -36,7 +36,19 @@ export function setTokenData(state: TokenState, payload: SetTokenDataPayload): T
   const tokenValues = Array.isArray(payload.values) ? payload.values : removeIdPropertyFromTokens(payload.values);
 
   // When the remote data has changed, we will update the last synced state
-  const lastSyncedState = payload.hasChangedRemote ? JSON.stringify(compact([tokenValues, payload.themes, TokenFormat.format]), null, 2) : state.lastSyncedState;
+  const lastSyncedState = payload.hasChangedRemote
+    ? (() => {
+      const cleanedThemes = (payload.themes ?? []).map((theme) => ({
+        ...theme,
+        selectedTokenSets: Object.fromEntries(
+          Object.entries(theme.selectedTokenSets).filter(
+            ([setName, status]) => allAvailableTokenSets.includes(setName) && status !== 'disabled',
+          ),
+        ),
+      }));
+      return JSON.stringify(compact([tokenValues, cleanedThemes, TokenFormat.format]), null, 2);
+    })()
+    : state.lastSyncedState;
 
   // @README (1) for the sake of normalization we will set the DISABLED status for all available token sets
   // this way we can always be certain the status is available. This behavior is also reflected in the createTokenSet logic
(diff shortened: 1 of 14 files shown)


packages/tokens-studio-for-figma/src/app/store/models/reducers/tokenState/setTokenData.ts
```diff
@@ -36,7 +36,19 @@ export function setTokenData(state: TokenState, payload: SetTokenDataPayload): T
   const tokenValues = Array.isArray(payload.values) ? payload.values : removeIdPropertyFromTokens(payload.values);
 
   // When the remote data has changed, we will update the last synced state
-  const lastSyncedState = payload.hasChangedRemote ? JSON.stringify(compact([tokenValues, payload.themes, TokenFormat.format]), null, 2) : state.lastSyncedState;
+  const lastSyncedState = payload.hasChangedRemote
+    ? (() => {
+      const cleanedThemes = (payload.themes ?? []).map((theme) => ({
```
six7: instead of doing this here, let's clean up in the area where we get payload.themes. like, where we are handing over this themes object, apparently this needs to be cleaned up in THAt step rather. (applies to all files changed in this PR)

six7: @copilot give my comment a try

akshay-gupta7: resolved in [abstract shared logic to utility](https://github.com/tokens-studio/figma-plugin/pull/3550/commits/148175dd8f14ec67b00a8e511694e4ba4ff5418a)

akshay-gupta7: @six7 cleaning up only in the area where we get payload.themes is not working as it fails in the comparison, so doing it there is making it redundant, 
let me know in case I didn't understand it correctly

packages/tokens-studio-for-figma/src/app/store/providers/ado/ado.tsx
```diff
@@ -229,7 +232,9 @@ export const useADO = () => {
               themes: content.themes,
               metadata: content.metadata,
             });
-            const stringifiedRemoteTokens = JSON.stringify(compact([sortedValues, content.themes, TokenFormat.format]), null, 2);
+            const cleanedThemes = cleanThemesSelectedTokenSets(content.themes, Object.keys(sortedValues));
+
+            const stringifiedRemoteTokens = JSON.stringify(compact([sortedValues, cleanedThemes, TokenFormat.format]), null, 2);
             dispatch.tokenState.setLastSyncedState(stringifiedRemoteTokens);
```
six7: what if instead of doing this in each of the providers... you just did it once in `setLastSyncedState`? i want to avoid that we have to do this in every provider, it feels like something we should be able to do in just 1 file.

akshay-gupta7: resolved in [clean themes in last synced state](https://github.com/tokens-studio/figma-plugin/pull/3550/commits/2ccb2dcd0cec9af7922508ec4f05c05d2bbcfdf8)

Issue #3317, opened 2025-04-11
Figma Displaying all Elements in the File are Modified


keeganedwin:
**Describe the bug**
This bus is related to #3169 that was solved but it looks like it's back. Creating a new ticket with a narrowed down focus on the possible issue.

The [user](https://tokens-studio.slack.com/archives/C02DLQXNY6Q/p1744115060816919) that reported this exported `styles & variables` without making any changes in the plugin. However although no changes were made, the Figma assets who that every style and variable in the file was modified. More [users](https://tokens-studio.slack.com/archives/C02JAL48UDN/p1744316796349699) have reported this as members consuming these libraries will not know what are the actual **incoming** changes. 

After taking a closer look into the issue, it looks like even if no changes are made, the plugin re-exports styles & variables regardless, and this is what's causing Figma to think that changes have been made. The plugin isn't creating new IDs, but it's more of a **touch base** situation. 

**To Reproduce**
1. Check variables to see what already exists
2. Export `styles & variables` for brands only - no changes made to tokens
3. Check variables to see no new created variable(s)
4. Open the file's libraries 
5. Check changes 
6. See error - all 20 variables are modified even though nothing was actually changed

**Expected behavior**
Not sure how we would go about fixing this

**Screenshots or Screencasts**
https://github.com/user-attachments/assets/12897629-a0d2-4f9a-b553-0351551301aa

**Figma File:**
[File](https://www.figma.com/design/oIo8nDwLmvKGWcfl1d70JL/Publish-Changes?node-id=6015-74&t=Xwqwmuuqn9wPgdcL-4)

PR #3551, opened 2025-09-04
fix reference recreation

six7:
Fixes https://github.com/tokens-studio/figma-plugin/issues/3317

# Fix Variable Reference Recreation Issue

## Problem
When updating variables with alias references, the plugin was unnecessarily recreating variables even when the actual resolved values were identical to the new values being applied. This happened because the comparison logic was comparing the new color value directly with the alias object (`{type: 'VARIABLE_ALIAS', id: 'VariableID:1:9'}`) instead of checking if the alias already points to the correct variable based on the token's reference.

## Root Cause
The `setColorValuesOnVariable` function was not properly handling variable aliases during value comparison. When a variable had an alias reference, the comparison would always fail because it was comparing a color object with an alias object, leading to unnecessary updates.

### 1. Enhanced Color Type Support
- **Added RGB type support**: Extended `isFigmaColorObject` to accept both RGB and RGBA objects (RGB objects don't require the 'a' property)
- **Created union type**: `RGBOrRGBA = RGB | RGBA` for better type safety

### 2. Smart Reference-Based Comparison for Aliases
- **Added `getReferencedVariableName` function**: Gets the variable name from an alias ID
- **Reference extraction**: Extracts reference name from token's `rawValue` (e.g., `{accent.default}` → `accent.default`)
- **Name normalization**: Uses `normalizeVariableName` to ensure consistent comparison
- **Direct comparison**: Compares normalized referenced variable name with normalized reference name from `rawValue`

### 3. Improved Comparison Logic
- **For aliases with references**: Compare the referenced variable name with the token's `rawValue` reference
- **For direct colors**: Compare actual color values as before
- **Early return**: If alias already points to correct variable (based on reference), no update needed

### 4. Enhanced Debugging
- **Added comprehensive logging**: Console logs for variable updates across all variable types (color, number, boolean, string)
- **Better visibility**: Logs show variable names, existing values, and new values for easier debugging

## Files Changed
- `packages/tokens-studio-for-figma/src/plugin/setColorValuesOnVariable.ts` - Main fix for alias comparison using reference-based logic
- `packages/tokens-studio-for-figma/src/plugin/setNumberValuesOnVariable.ts` - Added debugging logs
- `packages/tokens-studio-for-figma/src/plugin/setValuesOnVariable.ts` - Added debugging logs and pass `rawValue` to color function

## Key Insight
Instead of resolving aliases to color values and comparing colors (which was unnecessarily complex), we now simply compare the **normalized variable names** based on the token's reference. If the alias already points to a variable with the same normalized name as the token's reference, there's no change needed.

## How It Works
1. **Extract reference**: From token's `rawValue` (e.g., `{accent.default}` → `accent.default`)
2. **Get referenced variable**: From the alias ID (e.g., `accent/default`)
3. **Normalize both**: Convert to consistent format (e.g., `accent.default` vs `accent.default`)
4. **Compare**: If they match, skip the update; if not, proceed with update

## Testing
- Variables with alias references now properly compare based on token references instead of alias objects
- RGB objects (without alpha) are now supported alongside RGBA objects
- No regression in existing functionality for direct color values
- Much simpler and more efficient comparison logic

## Impact
- **Performance improvement**: Eliminates unnecessary variable updates when aliases already point to correct variables
- **Better user experience**: Reduces flickering and unnecessary changes in the UI
- **Enhanced debugging**: Better visibility into when and why variables are being updated
- **Type safety**: Improved TypeScript support for both RGB and RGBA color objects
- **Simpler code**: Much cleaner and more maintainable comparison logic

### DIFF
--- a/packages/tokens-studio-for-figma/src/plugin/setColorValuesOnVariable.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/setColorValuesOnVariable.ts
@@ -1,19 +1,6 @@
 import { isVariableWithAliasReference } from '@/utils/isAliasReference';
 import { convertToFigmaColor } from './figmaTransforms/colors';
-
-// Helper function to check if two colors are approximately equal within a threshold
-function isColorApproximatelyEqual(
-  color1: { r: number; g: number; b: number; a: number },
-  color2: { r: number; g: number; b: number; a: number },
-  threshold: number = 0.0001,
-): boolean {
-  return (
-    Math.abs(color1.r - color2.r) < threshold
-    && Math.abs(color1.g - color2.g) < threshold
-    && Math.abs(color1.b - color2.b) < threshold
-    && Math.abs(color1.a - color2.a) < threshold
-  );
-}
+import { isColorApproximatelyEqual } from '@/utils/isColorApproximatelyEqual';
 
 type RGB = { r: number; g: number; b: number };
 type RGBOrRGBA = RGB | RGBA;
--- a/packages/tokens-studio-for-figma/src/plugin/setNumberValuesOnVariable.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/setNumberValuesOnVariable.ts
@@ -28,7 +28,6 @@ export default function setNumberValuesOnVariable(variable: Variable, mode: stri
       }
     }
 
-    console.log('Setting number value on variable', variable.name, variable.valuesByMode[mode], value);
     variable.setValueForMode(mode, value);
   } catch (e) {
     console.error('Error setting numberVariable on variable', variable.name, e);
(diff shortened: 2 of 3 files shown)


six7:
@copilot few changes pls:
- make sure the way we did this is DRY, and works for the other variable types too not just color. We might even perform the variable alias check outside as its not reliant on the type
- create tests for the changes we made here: make sure that we dont create new variables or update the value when it didnt change.

six7:
@copilot the logic we have on deleting variables - i think we're still skipping early or doing something wrong. We still delete variables that were not touched.

six7:
@copilot tests are failing, can you pls fix them

akshay-gupta7 (commented):
This is working, 
However, there are a lot of debugging console log statements throughout the files,
Do we intend to keep that

PR #3581, opened 2025-09-16
fix issue where plugin creates empty file or folder or branch when se…

### LINKED ISSUE #3547: Gitlab Sync: Cannot Create a New File or Folder Path During Setup
**Describe the bug**
When setting up a new sync connection with Gitlab, the plugin requires the user to specify a file or folder path that already exists in the remote repository. It is not possible to define a new, non-existent path and have the plugin create it on the initial sync. This limits users who want to sync their files to a fresh, new location within their repo without first creating it manually via the Gitlab.

**To Reproduce**
1. Setup a new Gitlab provider
2. Add credentials
3. Enter a new and empty file-path

**Expected behavior**
A user should be able to enter a new file path (e.g., ⁠`tokens`) in the sync configuration. If the path does not exist, the plugin should interpret this as a directive to create a new folder (or an empty file) at the specified location in the Gitlab repository.

**Figma file (optional)**

**JSON (optional)**

akshay-gupta7:
…tting up sync

### Why does this PR exist?

Closes #3547  
When a user types an empty file or folder when setting up  sync which currently doesn't exist in the repo, the sync fails as it throws an error mentioning the file doesn't exist. This PR addresses that by creating that file/folder/branch.

This PR ignores the not found error and instead returns an empty array, which is an indication to create that file/folder on the repository.

### Testing this change

- Set up a new sync with github/Gitlab/Bitbucket/ADO
- Add a file/folder which currently doesn't exist
- Click on save and it should create the folder/file and push the tokens

### DIFF
--- a/.changeset/clever-students-laugh.md
+++ b/.changeset/clever-students-laugh.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Add ability to create empty file, folders or branch in a repository when setting up a new sync provider


six7 (changes requested):
remember DRY

six7 (approved):

PR #3697, opened 2025-11-24
remove user name to mix panel and hash user ids

### LINKED ISSUE #3698: Adjust Mixpanel to not track figmaUserId and name, use hashed figmaUserId instead as unique element
we cant use figma ids in mixpanel - we need to make sure we hash it - and we need to use the same hashing later in mixpanel as a bulk operation so that we still have a unique identifier.

first step is to come up with a strategy on how we can do this, both in mixpanel and in the plugin when we send the figma id to mixpanel

Todo
- Drop user name entirely
- figmaUserId needs to be hashed with some secret
- On mixpanel: We also need to bulk hash all stored figmaUserIds with that same secret

akshay-gupta7:
### Why does this PR exist?

Closes #3698  

Currently, Mixpanel events were sent using the user’s plain name, which may pose privacy concerns and isn’t ideal for user tracking. This PR replaces the usage of plain user names with hashed user IDs to improve user data privacy.

- Removes sending of plain user names to Mixpanel.
- Introduces hashing for Figma user IDs before sending them to Mixpanel events.
- Updates any related code that referenced Mixpanel’s userName property.
- Ensures that user identification for analytics works without exposing sensitive information.

### Testing this change

1. Check Mixpanel events via the analytics dashboard to confirm user names are no longer present—only hashed IDs should show.
2. Run the plugin and trigger event logging actions.
3. Inspect outgoing payloads to ensure no plain text user names are included.

### DIFF
--- a/packages/tokens-studio-for-figma/.changeset/hash-figma-id-mixpanel.md
+++ b/packages/tokens-studio-for-figma/.changeset/hash-figma-id-mixpanel.md
@@ -1,4 +1,0 @@
----
-"@tokens-studio/figma-plugin": patch
----
-Hash Figma user ID before sending to Mixpanel, and remove user name from all Mixpanel events. Uses a secret from the environment for hashing, ensuring privacy and consistent identification for bulk Mixpanel operations.


akshay-gupta7:
@copilot can you investigate and fix the tests in analytics.test.tsx in context to this PR

packages/tokens-studio-for-figma/.changeset/hash-figma-id-mixpanel.md
```diff
@@ -0,0 +1,4 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+Hash Figma user ID before sending to Mixpanel, and remove user name from all Mixpanel events. Uses a secret from the environment for hashing, ensuring privacy and consistent identification for bulk Mixpanel operations.
```
six7: @akshay-gupta7 lets not include a changeset for this - its not user facing

akshay-gupta7: reoslved in [remove changeset](https://github.com/tokens-studio/figma-plugin/pull/3697/commits/3ec155c56b3597a62e8d4d48094c259070ffa987)

PR #3787, opened 2026-02-19
variable scoping enhancements

### LINKED ISSUE #3774: Scoping enhancements
- Add hiddenFromPublishing (true/false/indeterminate)
- Allow turning off all scopes

akshay-gupta7:
### Why does this PR exist?

Closes #3774

### DIFF
--- a/packages/tokens-studio-for-figma/src/app/components/FigmaVariableForm.tsx
+++ b/packages/tokens-studio-for-figma/src/app/components/FigmaVariableForm.tsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
 import {
   IconButton, Heading, Checkbox, Text, Stack, Label,
 } from '@tokens-studio/ui';
@@ -25,6 +26,7 @@ export default function FigmaVariableForm({
   handleFigmaVariableChange: (scopes: VariableScope[], codeSyntax: Partial<Record<CodeSyntaxPlatform, string>>, hiddenFromPublishing?: boolean) => void;
   handleRemoveFigmaVariable: () => void;
 }) {
+  const { t } = useTranslation(['tokens']);
   const [figmaVariableVisible, setFigmaVariableVisible] = React.useState(false);
 
   const currentScopes = useMemo(() => internalEditToken?.$extensions?.['com.figma.scopes'] as VariableScope[] || [], [internalEditToken]);
@@ -123,11 +125,11 @@ export default function FigmaVariableForm({
 
   let hiddenFromPublishingTooltip: string;
   if (currentHiddenFromPublishing === true) {
-    hiddenFromPublishingTooltip = 'Hidden from publishing: ON — this variable will be hidden in published libraries';
+    hiddenFromPublishingTooltip = t('figmaVariable.hiddenFromPublishing.tooltipOn');
   } else if (currentHiddenFromPublishing === false) {
-    hiddenFromPublishingTooltip = 'Hidden from publishing: OFF — this variable will be visible in published libraries';
+    hiddenFromPublishingTooltip = t('figmaVariable.hiddenFromPublishing.tooltipOff');
   } else {
-    hiddenFromPublishingTooltip = 'Hidden from publishing: not set — the variable\'s current Figma setting will be preserved';
+    hiddenFromPublishingTooltip = t('figmaVariable.hiddenFromPublishing.tooltipNotSet');
   }
 
   if (!shouldShowFigmaVariableSection) {
@@ -212,7 +214,7 @@ export default function FigmaVariableForm({
                   </span>
                 </Tooltip>
                 <Label css={{ fontWeight: '$sansRegular', fontSize: '$small' }} htmlFor="hiddenFromPublishing">
-                  Hidden from publishing
+                  {t('figmaVariable.hiddenFromPublishing.label')}
                 </Label>
               </Box>
             </Box>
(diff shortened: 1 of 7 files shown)


six7 (commented):
## Review: hiddenFromPublishing indeterminate state & related issues

### 🔴 Issue 1 (Critical): `hiddenFromPublishing` indeterminate state is broken

When a token has NO `hiddenFromPublishing` set (indeterminate/undefined), the code coerces it to `false` — and then any scope or codeSyntax change will save that `false` onto the token, overriding whatever the variable already had in Figma.

**In `FigmaVariableForm.tsx`:**
```typescript
const currentHiddenFromPublishing = useMemo(() => 
  internalEditToken?.$extensions?.['com.figma.hiddenFromPublishing'] as boolean || false,
  [internalEditToken]
);
```
The `|| false` converts `undefined` → `false`. Then every callback (`handleScopeChange`, `handleCodeSyntaxPlatformChange`, etc.) passes this value through:
```typescript
handleFigmaVariableChange(newScopes, currentCodeSyntax, currentHiddenFromPublishing); // passes false, not undefined
```

In `EditTokenForm.tsx` `handleFigmaVariableChange`, since `false` is a boolean, this stamps `com.figma.hiddenFromPublishing: false` onto the token:
```typescript
if (typeof hiddenFromPublishing === 'boolean') {
  newExtensions['com.figma.hiddenFromPublishing'] = hiddenFromPublishing; // false is a boolean!
}
```

Now when syncing to Figma, the variable's existing `hiddenFromPublishing` gets overwritten to `false` even though the user never touched it.

Additionally, `addFigmaVariable` explicitly passes `false` instead of `undefined`:
```typescript
handleFigmaVariableChange([], {}, false); // should be undefined
```

**Fix:** `currentHiddenFromPublishing` should be `boolean | undefined`, defaulting to `undefined` (not `false`). Callbacks should only pass `hiddenFromPublishing` when it was explicitly set by the user.

---

### 🔴 Issue 2 (Bug): Duplicate `setInternalEditToken` in `EditTokenForm.tsx`

In `handleFigmaVariableChange`, after the new `hiddenFromPublishing` block, there are TWO consecutive `setInternalEditToken` calls:

```typescript
// FIRST call — BUG: spreads newExtensions at TOP LEVEL of token
setInternalEditToken({
  ...internalEditToken,
  ...newExtensions,   // ← puts 'com.figma.scopes', 'com.figma.hiddenFromPublishing' etc. as top-level keys
});

// SECOND call — correct: sets $extensions
setInternalEditToken({
  ...internalEditToken,
  $extensions: newExtensions as SingleToken['$extensions'],
});
```

The first call is a bug — it spreads extension keys directly onto the token object instead of inside `$extensions`. The first `setInternalEditToken` should be removed entirely.

---

### 🟡 Issue 3: No three-way checkbox UI for `hiddenFromPublishing`

The issue spec says "Add hiddenFromPublishing (true/false/indeterminate)", but the UI only renders a two-state checkbox:
```tsx
<Checkbox checked={currentHiddenFromPublishing} ... />
```
There is no way for the user to set the value back to indeterminate once they've toggled it. The codebase already has a three-way checkbox pattern (checked/unchecked/indeterminate) that should be used here.

---

### 🟡 Issue 4: No tests

No test updates for the new `hiddenFromPublishing` logic in `updateVariablesFromPlugin.ts`, `setValuesOnVariable.ts`, or the UI components. Tests should cover at minimum:
- `undefined` hiddenFromPublishing leaves variable untouched
- `true`/`false` explicitly sets the value on the variable
- Scope changes don't accidentally stamp `hiddenFromPublishing: false` onto tokens that had it unset

packages/tokens-studio-for-figma/src/app/components/FigmaVariableForm.tsx
```diff
@@ -99,6 +107,29 @@ export default function FigmaVariableForm({
...
+    } else {
+      nextValue = undefined; // back to indeterminate — removes the extension key
+    }
+    track('Set hidden from publishing', { checked: nextValue });
+    handleFigmaVariableChange(currentScopes, currentCodeSyntax, nextValue);
+  }, [currentHiddenFromPublishing, currentScopes, currentCodeSyntax, handleFigmaVariableChange]);
+
+  let hiddenFromPublishingTooltip: string;
+  if (currentHiddenFromPublishing === true) {
+    hiddenFromPublishingTooltip = 'Hidden from publishing: ON — this variable will be hidden in published libraries';
```
six7: these are too long. Just make it shorter. Just say `Will be hidden in published libraries`
Will be visible in published libraries
Current Figma setting will be preserved

packages/tokens-studio-for-figma/src/app/components/FigmaVariableForm.tsx
```diff
@@ -163,6 +194,29 @@ export default function FigmaVariableForm({
...
+                  <span>
+                    <Checkbox
+                      id="hiddenFromPublishing"
+                      checked={currentHiddenFromPublishing === undefined ? 'indeterminate' : currentHiddenFromPublishing}
+                      onCheckedChange={handleHiddenFromPublishingCheckedChange}
+                    />
+                  </span>
+                </Tooltip>
+                <Label css={{ fontWeight: '$sansRegular', fontSize: '$small' }} htmlFor="hiddenFromPublishing">
+                  Hidden from publishing
```
six7: translations

packages/tokens-studio-for-figma/src/app/components/FigmaVariableForm.tsx
six7: plus translations

packages/tokens-studio-for-figma/src/app/components/FigmaVariableForm.tsx
akshay-gupta7: resolved in [516774f](https://github.com/tokens-studio/figma-plugin/pull/3787/commits/516774f3aba23397896a47f3bfc06889943cbaee)

packages/tokens-studio-for-figma/src/app/components/FigmaVariableForm.tsx
akshay-gupta7: resolved om [516774f](https://github.com/tokens-studio/figma-plugin/pull/3787/commits/516774f3aba23397896a47f3bfc06889943cbaee)

PR #3825, opened 2026-03-31
filter sets in themes by status

akshay-gupta7:
### Why does this PR exist?

https://github.com/user-attachments/assets/7ad0e0a9-8f00-4bef-81b3-01c7b7e9c780

Screenshot 2026-04-01 at 12 31 07 PM


six7 (approved):
You just need to add translations to the other languages

PR #3836, opened 2026-04-08
fix: resolve mode-specific baseFontSize for rem scaling in variable e…

### LINKED ISSUE #3800: Bug: Token reference in baseFontSize no longer resolves per theme during Variables/Styles export
## Summary

When a token is referenced in the plugin's `baseFontSize` setting, it should resolve per theme context during export — allowing different themes to produce different rem→px conversions. This appears to have stopped working in 2.11.0. Instead, all themes now resolve rem values using the currently active theme's baseline, producing identical dimensions across modes.

Blocks responsive typography workflows that rely on per-theme baseFontSize.

## Environment

- **Plugin version:** 2.11.0
- **Product area:** Plugin Core / Theming / Variables & Styles export
- **Last known working version:** Pre-2.11.0 

## Reproduction Steps

1. Import the attached token files into the plugin (multi-file setup)
2. The `baseFontSize` should reference a token that resolves differently per theme (e.g., `{typography.baseline}` → 16 for Mobile, 17 for Tablet, 18 for Desktop)
3. Export Variables and Styles
4. Apply a generated text style (e.g., `body1.normal`) to a text element
5. Switch mode from Mobile to Tablet or Desktop
6. **Observe:** All modes show identical text dimensions

**Expected:** Each mode should show different text dimensions because `1rem` resolves to different px values per theme (16px / 17px / 18px).

**Actual:** All modes resolve to the same px value — whichever theme is currently active in the plugin.

## Token Structure

```
global.json
├── primitives.typography.baseline-mobile: 16
├── primitives.typography.baseline-tablet: 17
└── primitives.typography.baseline-desktop: 18

device/mobile.json
├── typography.baseline: → {primitives.typography.baseline-mobile}
├── typography.font.size.md: 1rem
└── typography.font.lineHeight.snug: 1.25rem

device/tablet.json
├── typography.baseline: → {primitives.typography.baseline-tablet}
├── typography.font.size.md: 1rem
└── typography.font.lineHeight.snug: 1.25rem

device/desktop.json
├── typography.baseline: → {primitives.typography.baseline-desktop}
├── typography.font.size.md: 1rem
└── typography.font.lineHeight.snug: 1.25rem

components/typography.json
└── typography.styles.body1.normal (typography composite):
    fontSize: {typography.font.size.md}
    lineHeight: {typography.font.lineHeight.snug}
```

**Themes ($themes.json):**
- Mobile → enables: global + device/mobile + components/typography
- Tablet → enables: global + device/tablet + components/typography
- Desktop → enables: global + device/desktop + components/typography

All three themes are in a "Device" group, mapped to a single Figma Variable Collection with three modes.

## Investigation Lead

The 2.11.0 release introduced **"Smart Defaults"** — the plugin now auto-selects the first theme from each group for quicker setup. This change to theme auto-selection/context handling could be affecting how `baseFontSize` token references are resolved during export. Previously, the export may have resolved the baseFontSize token reference independently per theme context; the new auto-selection behavior might be causing the active theme's resolved value to "leak" into all other theme contexts.

## Customer Context
- **Slack thread:** https://tokens-studio.slack.com/archives/C02DLQXNY6Q/p1772727369198849

akshay-gupta7:
…xport

### Why does this PR exist?

Closes #3800  

This PR addresses issues with `baseFontSize` resolution when exporting variables to Figma. Previously, numeric or aliased font sizes (especially those involving math evaluation) were not correctly handled during per-mode exports, leading to inconsistent scaling across different modes in the same collection.

- **Improved `baseFontSize` Resolution**: Updated `updateVariables.ts` to correctly handle `number` types and evaluate aliases for `baseFontSize`, ensuring consistent string conversion.
- **Fixed Variable Update Logic**: Corrected a logical check in `setStringValuesOnVariable.ts` that was preventing some existing variable values from being updated correctly.
- **Enhanced Test Reliability**: 
    - Added tests for math-evaluated font sizes and per-mode rem scaling.
    - Updated `updateVariables.test.ts` to use isolated variable mocks, ensuring that scaling validation is accurate and free from side effects of shared mocks.

### Testing this change

1. Open the plugin and ensure `baseFontSize` is set (either as a numeric value or an alias like `{typography.baseline}`).
2. Export variables to a Figma collection with multiple modes.
3. Verify that `rem` values (e.g., `1rem` spacing) are correctly translated to pixels based on the specific `baseFontSize` defined for each theme/mode.

**Automated Tests:**
Run the updated test suite:
```bash
yarn test src/plugin/updateVariables.test.ts
```

### Additional Notes (if any)

The test mock isolation was necessary because `base/font-size` and other variables were sharing the same `setValueForMode` mock, which could hide bugs where specific variables weren't being updated.

### DIFF
--- a/.changeset/fresh-brooms-build.md
+++ b/.changeset/fresh-brooms-build.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fix per-mode `baseFontSize` resolution and variable export, adding support for numeric/aliased values and improved testing.
--- a/.changeset/fresh-brooms-build.md
+++ b/.changeset/fresh-brooms-build.md
@@ -2,4 +2,4 @@
 "@tokens-studio/figma-plugin": patch
 ---
 
-Fix per-mode `baseFontSize` resolution and variable export, adding support for numeric/aliased values and improved testing.
+Support per-mode `baseFontSize` resolution via aliases, ensuring correct rem-to-pixel scaling when exporting variables across different themes. Also includes improved handling of numeric/aliased font sizes in variable exports and more robust testing for dimension conversions.
--- a/packages/tokens-studio-for-figma/src/plugin/setBooleanValuesOnVariable.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/setBooleanValuesOnVariable.ts
@@ -4,8 +4,8 @@ export default function setBooleanValuesOnVariable(variable: Variable, mode: str
   try {
     const existingVariableValue = variable.valuesByMode[mode];
     if (
-      existingVariableValue === undefined
-      || !(typeof existingVariableValue === 'boolean' || isVariableWithAliasReference(existingVariableValue))
+      existingVariableValue !== undefined
+      && !(typeof existingVariableValue === 'boolean' || isVariableWithAliasReference(existingVariableValue))
     ) return;
 
     const newValue = value === 'true';
--- a/packages/tokens-studio-for-figma/src/plugin/setColorValuesOnVariable.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/setColorValuesOnVariable.ts
@@ -23,8 +23,8 @@ export default function setColorValuesOnVariable(variable: Variable, mode: strin
     const { color, opacity } = convertToFigmaColor(value);
     const existingVariableValue = variable.valuesByMode[mode];
     if (
-      !existingVariableValue
-      || !(isFigmaColorObject(existingVariableValue) || isVariableWithAliasReference(existingVariableValue))
+      existingVariableValue
+      && !(isFigmaColorObject(existingVariableValue) || isVariableWithAliasReference(existingVariableValue))
     ) return;
 
     const newValue = { ...color, a: opacity };
--- a/packages/tokens-studio-for-figma/src/plugin/setNumberValuesOnVariable.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/setNumberValuesOnVariable.ts
@@ -16,8 +16,8 @@ export default function setNumberValuesOnVariable(variable: Variable, mode: stri
     }
     const existingVariableValue = variable.valuesByMode[mode];
     if (
-      existingVariableValue === undefined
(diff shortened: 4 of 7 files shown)


.changeset/fresh-brooms-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fix per-mode `baseFontSize` resolution and variable export, adding support for numeric/aliased values and improved testing.
```
six7: can we be more specific what actually now works? we had aliased and numeric support before - this makes it seem like a net new thing, users will be curious what is now possible

packages/tokens-studio-for-figma/src/plugin/setBooleanValuesOnVariable.ts
```diff
@@ -4,8 +4,8 @@ export default function setBooleanValuesOnVariable(variable: Variable, mode: str
   try {
     const existingVariableValue = variable.valuesByMode[mode];
     if (
-      existingVariableValue === undefined
-      || !(typeof existingVariableValue === 'boolean' || isVariableWithAliasReference(existingVariableValue))
+      existingVariableValue !== undefined
```
six7, line 7: is this intentional? its the opposite of before? before we had it as "if its undefined or if its not a boolean return", now we have "if its NOT undefined AND if its not a boolean return" meaning now if the value is undefined we dont return?

akshay-gupta7: The previous version would return early if the value was undefined, which meant we could never set the initial value for a mode. The new logic allows undefined values to proceed to initialization, but still protects against type mismatches if an existing value of a different type is already present. This now matches the logic we use in other setters like `setNumberValuesOnVariable.ts`

.changeset/fresh-brooms-build.md
akshay-gupta7: updated in [update changeset](https://github.com/tokens-studio/figma-plugin/pull/3836/commits/3f93299376ca0afd533d0f1bf4b89b1f6c8bf423)

six7 (approved):

PR #3889, opened 2026-06-01
Fix typography Resolution and Application whilst using Tokens Studio OAuth

akshay-gupta7:
### Why does this PR exist?

Typography tokens were broken in two ways for Tokens Studio OAuth sync users: they weren't resolving correctly in the UI, and even when they appeared resolved, applying them to a text layer in Figma had no visual effect until the user manually clicked "Pull from studio OAuth sync."

**Fix 1 — Typography token resolution** (`tokenHelpers.ts`)

Added a second pass to `mergeServerResolvedTokens` that re-resolves composite tokens (typography, border, shadow) whose individual properties reference atomic tokens updated in the server delta. Previously, only tokens directly named in the delta were updated; composite tokens referencing those atoms kept stale locally-resolved values.

**Fix 2 — Typography token application** (`tokenHelpers.ts` + `tryApplyTypographyCompositeVariable.ts`)

The OAuth server returns composite tokens as JSON-stringified strings in the resolution delta. `mergeServerResolvedTokens` was blindly assigning this string to `.value`, clobbering the object. Downstream code then read `.fontFamily` off a string → `undefined` for every property → nothing applied.

Additionally, resolved values can arrive in formats the apply path can't consume: DTCG dimension objects (`{value: 24, unit: 'px'}`) from local variable resolution, and JSON-array font families (`'["Comic","Sans","MS"]'`) from the server. `parseFloat({...})` on a dimension object produces NaN, assigning NaN to a node throws, and the single outer `try/catch` was silently aborting the entire property loop.

Changes:
- Parse server's JSON-stringified composite value back to an object when the local `.value` is an object type
- Normalize DTCG dimension objects → `'24px'` / `'100%'` and JSON-array font families → `'Comic, Sans, MS'` before applying
- Wrap each property application in its own `try/catch` so one malformed value can't abort the rest
- Add a NaN guard before assigning transformed numeric values to the node

**Why pull always worked:** the Pull button's `setTokenData` reducer sets `serverResolvedTokens: null`, so `mergeServerResolvedTokens` returns locally-resolved object values untouched — the JSON-string clobber never occurred.

### Testing this change

1. Connect a Figma file to a Tokens Studio OAuth project with typography tokens that reference other tokens (e.g. `fontFamily` referencing `fontFamily.brand`)
2. Select a text layer and apply a typography token — it should visually update immediately without needing to pull
3. Verify tokens with DTCG-format dimension sub-values and multi-family font stacks also apply correctly
4. Confirm the pull flow still works as before

### Additional Notes (if any)

Root cause was specific to Tokens Studio OAuth sync — for all other storage providers `serverResolvedTokens` is null and `mergeServerResolvedTokens` is a no-op, so they were unaffected.

### DIFF
--- a/.changeset/heavy-shoes-tickle.md
+++ b/.changeset/heavy-shoes-tickle.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fixed two issues affecting typography tokens for Tokens Studio OAuth users: typography tokens referencing server-resolved values were not resolving correctly, and even when resolved they did not visually update text layers on apply (font, size, line height and other properties were silently dropped until a manual pull).
--- a/.changeset/heavy-shoes-tickle.md
+++ b/.changeset/heavy-shoes-tickle.md
@@ -2,4 +2,4 @@
 "@tokens-studio/figma-plugin": patch
 ---
 
-Fixed two issues affecting typography tokens for Tokens Studio OAuth users: typography tokens referencing server-resolved values were not resolving correctly, and even when resolved they did not visually update text layers on apply (font, size, line height and other properties were silently dropped until a manual pull).
+Fixed two issues affecting typography tokens for Studio platform users: typography tokens referencing server-resolved values were not resolving correctly, and even when resolved they did not visually update text layers on apply (font, size, line height and other properties were silently dropped until a manual pull).
--- a/packages/tokens-studio-for-figma/src/plugin/tryApplyTypographyCompositeVariable.ts
+++ b/packages/tokens-studio-for-figma/src/plugin/tryApplyTypographyCompositeVariable.ts
@@ -9,6 +9,32 @@ import { ApplyVariablesStylesOrRawValues } from '@/constants/ApplyVariablesStyle
 // Cache to track font loading promises to prevent race conditions in Promise.all
 const fontLoadingPromises = new Map<string, Promise<void>>();
 
+function dtcgUnitToCssUnit(unit: unknown): string {
+  if (unit === 'PIXELS') return 'px';
+  if (unit === 'PERCENT') return '%';
+  return typeof unit === 'string' ? unit : '';
+}
+
+// Resolved typography values can arrive in formats the raw-apply path can't consume:
+//  - DTCG dimension objects like { value: 24, unit: 'px' } (from variable/local resolution)
+//  - JSON-array font families like '["Comic","Sans","MS"]' (from server resolution)
+// transformValue/setFontStyleOnTarget expect plain strings (e.g. '24px', 'Comic, Sans, MS'),
+// so we normalize here. parseFloat on a raw object yields NaN, which throws when assigned.
+function normalizeTypographyPropertyValue(key: string, raw: any): any {
+  if (raw && typeof raw === 'object' && !Array.isArray(raw) && 'value' in raw) {
+    return `${raw.value}${dtcgUnitToCssUnit(raw.unit)}`;
+  }
+  if ((key === 'fontFamily' || key === 'fontFamilies') && typeof raw === 'string' && raw.trim().startsWith('[')) {
+    try {
+      const parsed = JSON.parse(raw);
+      if (Array.isArray(parsed)) return parsed.map((f) => String(f)).join(', ');
+    } catch {
+      // Not valid JSON — fall through and use the original string.
+    }
+  }
+  return raw;
+}
+
 export async function tryApplyTypographyCompositeVariable({
   target, value, baseFontSize, resolvedValue,
 }: {
@@ -26,9 +52,13 @@ export async function tryApplyTypographyCompositeVariable({
 
   if (typeof value === 'string') return;
 
+  const normalizedValue = Object.fromEntries(
+    Object.entries(value).map(([key, val]) => [key, normalizeTypographyPropertyValue(key, val)]),
+  ) as typeof value;
+
   try {
     // We iterate over all keys of the typography object and apply variables if available, otherwise we apply the value directly
-    for (const [originalKey, val] of Object.entries(value).filter(([_, keyValue]) => typeof keyValue !== 'undefined')) {
+    for (const [originalKey, val] of Object.entries(normalizedValue).filter(([_, keyValue]) => typeof keyValue !== 'undefined')) {
       if (typeof val === 'undefined') return;
       let successfullyAppliedVariable = false;
       if (resolvedValue[originalKey]?.toString().startsWith('{') && resolvedValue[originalKey].toString().endsWith('}') && shouldCreateStylesWithVariables) {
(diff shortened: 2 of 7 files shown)


packages/tokens-studio-for-figma/src/utils/tokenHelpers.ts
```diff
@@ -105,16 +105,69 @@ export function mergeServerResolvedTokens(
...
   });
+
+  // Second pass: re-resolve composite tokens (typography, border, shadow) whose
+  // individual properties reference atomic tokens that were updated in the server delta.
+  // Without this step, e.g. a typography token referencing {fontFamily.brand} keeps its
+  // locally-resolved (possibly stale or unresolved) value even after fontFamily.brand is
+  // updated by the server.
+  return firstPass.map((token) => {
+    const rawValue = token.resolvedValueWithReferences;
+    if (!rawValue || typeof rawValue !== 'object' || Array.isArray(rawValue)) {
```
six7, line 144: Verified gap: this `Array.isArray(rawValue)` guard makes the second pass skip multi-value (array) shadow tokens, even though the comment just above and the PR description both say shadow is covered. A multi-shadow token whose e.g. `color` references an atom present in the server delta keeps its stale locally-resolved value.

Two extra wrinkles:
- Single-object box-shadow tokens (`TokenBoxshadowValue` is `object | object[]`) fall through to the object branch and *are* handled, so behaviour is inconsistent between the two shadow shapes.
- Even without the array guard, the matching loop only inspects top-level `key: '{ref}'` string entries, so it would never catch references nested inside array elements (`value[0].color`).

Suggest either implementing array + nested-property handling, or narrowing the 'shadow' claim in this comment and the PR description to avoid implying coverage that isn't there.

packages/tokens-studio-for-figma/src/utils/tokenHelpers.ts
```diff
@@ -105,16 +105,69 @@ export function mergeServerResolvedTokens(
...
+      // string (e.g. `value.fontFamily`) and gets `undefined`, silently dropping the apply.
+      let nextValue: SingleToken['value'] = serverValue;
+      if (
+        typeof serverValue === 'string'
+        && typeof token.value === 'object'
+        && token.value !== null
+        && (serverValue.startsWith('{') || serverValue.startsWith('['))
+      ) {
+        try {
+          nextValue = JSON.parse(serverValue);
```
six7: Minor: there's an existing `src/utils/tryParseJson.ts` helper that does exactly this try/parse/catch (it returns `null` and logs on failure). Reusing it here — and in `normalizeTypographyPropertyValue` in tryApplyTypographyCompositeVariable.ts — would remove the duplicated inline try/catch. Just guard for the `null` return.

packages/tokens-studio-for-figma/src/plugin/tryApplyTypographyCompositeVariable.ts
```diff
@@ -26,9 +52,13 @@ export async function tryApplyTypographyCompositeVariable({
...
 
+  const normalizedValue = Object.fromEntries(
+    Object.entries(value).map(([key, val]) => [key, normalizeTypographyPropertyValue(key, val)]),
+  ) as typeof value;
+
   try {
     // We iterate over all keys of the typography object and apply variables if available, otherwise we apply the value directly
-    for (const [originalKey, val] of Object.entries(value).filter(([_, keyValue]) => typeof keyValue !== 'undefined')) {
+    for (const [originalKey, val] of Object.entries(normalizedValue).filter(([_, keyValue]) => typeof keyValue !== 'undefined')) {
       if (typeof val === 'undefined') return;
```
six7: Pre-existing nit, but it's in code this PR is hardening: `if (typeof val === 'undefined') return;` aborts the *entire* loop on an undefined value, which directly contradicts the new per-property isolation goal. It's currently unreachable (entries are pre-filtered for undefined just above), but `continue` would be the correct/consistent choice if it ever did fire.

six7 (commented):
Re-reviewed after digging into the shadow path and the server-delta format (`fetchServerResolvedTokens` returns a flat `{ name: stringValue }` map; composites come back JSON-stringified). Two clarifications and one confirmed gap below.

**Confirmed:** the *application* clobber is fixed for shadows too — the first-pass `JSON.parse` handles `'[...]'` arrays, so a shadow that's directly in the delta now applies without a re-pull. 

**Confirmed gap:** the *second-pass re-resolution* does **not** cover multi-value (array) shadow tokens (details inline). Since the whole reason the second pass exists is that the server delta carries atoms but not the composites referencing them, array shadows referencing a changed atom will still go stale — exactly the class of bug this PR targets.

**Tests:** none of the new behaviour is covered. Worth adding unit tests to `mergeServerResolvedTokens.test.ts` (JSON-string composite parse, second-pass atom re-resolution, and an array-shadow case to pin down whatever behaviour you decide on) and to `tryApplyTypographyCompositeVariable.test.ts` (DTCG dimension `{value,unit}` → `'24px'`, JSON-array fontFamily normalization, per-property isolation so one bad prop doesn't abort the rest). Both suites already exist and pass, so extending them is cheap.

Nothing here blocks the typography fix, which looks correct and is a clear improvement.

.changeset/heavy-shoes-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Fixed two issues affecting typography tokens for Tokens Studio OAuth users: typography tokens referencing server-resolved values were not resolving correctly, and even when resolved they did not visually update text layers on apply (font, size, line height and other properties were silently dropped until a manual pull).
```
six7: reword, nobody will know what oauth users are - "Studio platform users"

packages/tokens-studio-for-figma/src/plugin/tryApplyTypographyCompositeVariable.ts
```diff
@@ -5,10 +5,33 @@ import { ResolvedTypographyObject } from './ResolvedTypographyObject';
 import { transformTypographyKeyToFigmaVariable } from './transformTypographyKeyToFigmaVariable';
 import { SingleTypographyToken } from '@/types/tokens';
 import { ApplyVariablesStylesOrRawValues } from '@/constants/ApplyVariablesStyleOrder';
+import { tryParseJson } from '@/utils/tryParseJson';
 
 // Cache to track font loading promises to prevent race conditions in Promise.all
 const fontLoadingPromises = new Map<string, Promise<void>>();
 
+function dtcgUnitToCssUnit(unit: unknown): string {
```
six7: extract to its own file, add a unit test

packages/tokens-studio-for-figma/src/plugin/tryApplyTypographyCompositeVariable.ts
```diff
@@ -5,10 +5,33 @@ import { ResolvedTypographyObject } from './ResolvedTypographyObject';
...
+  if (unit === 'PERCENT') return '%';
+  return typeof unit === 'string' ? unit : '';
+}
+
+// Resolved typography values can arrive in formats the raw-apply path can't consume:
+//  - DTCG dimension objects like { value: 24, unit: 'px' } (from variable/local resolution)
+//  - JSON-array font families like '["Comic","Sans","MS"]' (from server resolution)
+// transformValue/setFontStyleOnTarget expect plain strings (e.g. '24px', 'Comic, Sans, MS'),
+// so we normalize here. parseFloat on a raw object yields NaN, which throws when assigned.
+function normalizeTypographyPropertyValue(key: string, raw: any): any {
```
six7: extract to own file, add a unit test

.changeset/heavy-shoes-tickle.md
akshay-gupta7: resolved in [8d6e2dd](https://github.com/tokens-studio/figma-plugin/pull/3889/commits/8d6e2dd6882f9ce5005bd959e463508f8bd0c33c)

packages/tokens-studio-for-figma/src/plugin/tryApplyTypographyCompositeVariable.ts
akshay-gupta7: resolved in [refactor: extract dtcgUnitToCssUnit and normalizeTypographyPropertyVa…](https://github.com/tokens-studio/figma-plugin/pull/3889/commits/8e50c6f8df5981c36df4a99012b4996ccf59aa2a)

packages/tokens-studio-for-figma/src/plugin/tryApplyTypographyCompositeVariable.ts
akshay-gupta7: resolved in [refactor: extract dtcgUnitToCssUnit and normalizeTypographyPropertyVa…](https://github.com/tokens-studio/figma-plugin/pull/3889/commits/8e50c6f8df5981c36df4a99012b4996ccf59aa2a)

six7 (approved):

PR #3946, opened 2026-08-01
feat: fail license API requests over to the legacy host during migration

six7:
## Why

We're moving the plugin's license API from `licence.tokens.studio` to `api-production.tokens.studio`. The two run on separate datastores, and the license data hasn't been imported yet, so neither host can serve every user on its own for the duration of the migration.

This replaces the configurable license API URL this PR originally introduced. Rather than a hidden override for internal testers, every request now goes to the new API first and falls back to the old one when it can't answer — so both stay live until the import completes. The follow-up PR drops the legacy host and this becomes a single-host client again.

## What changed

- **`src/utils/licenseApiUrl.ts`** — `LICENSE_API_HOSTS`, in the order they're tried: `api-production.tokens.studio`, then `LICENSE_API_URL` (defaulting to `licence.tokens.studio`). Deduped, so pointing the env var at the primary doesn't issue every request twice. Keeps the regex normalizer from the original PR — the `URL` constructor is still unavailable in the Figma sandbox.
- **`src/utils/licenseApi.ts`** — `fetchLicenseApi()`. Falls through to the next host on a transport error, a timeout, any non-200, a body it can't read as JSON, or a payload the caller rejects. Reads the body once via `text()`; the previous helpers called `res.json()` in two branches, which throws on the second read. The last host's `message` is what surfaces, so error text is unchanged when both fail.
- **`validateLicense` / `getLicenseKey` / `removeLicense`** — rewritten on top of it.
- **`src/mocks/handlers.ts`** — msw now serves both hosts.

### A timeout, because there wasn't one

`fetch` has no default timeout, and `addLicenseFactory` awaits these calls on the startup path. An unreachable host previously hung startup indefinitely rather than failing over — precisely the case this PR exists for. Requests now abort after 8s.

### Guarding against a silent downgrade

`validate-license` also rejects a 200 that carries neither a plan nor entitlements. A partially imported record can answer exactly that way, and `addLicenseKey` treats any error-free response as `VERIFIED` — so a paying user would be quietly dropped to the free tier. Marked in the source to be deleted along with the legacy host: with one host there is nowhere to fall through to, and the guard would start turning entitlement-less responses into hard errors.

### Telemetry, because the cutover needs a gate

Each fallback emits a `License API fallback` event: `{endpoint, primaryStatus, primaryReason, fallbackStatus, resolvedBy, migrationGap, figmaId}`.

`migrationGap` is precomputed as *the primary could not answer and the legacy host could* — the literal definition of a record the import missed. Users with no license fail on both hosts and are excluded, which matters: they'd otherwise give the metric a permanent floor it could never clear, since "user not found" and "user has no license" are the same response. **`migrationGap` reaching zero is the signal that it's safe to turn the old system off** — not the import finishing. `figmaId` goes through `track()`'s existing hashing, so a gap is traceable to an account.

Sentry now reports only when *no* host responded, which is what the per-helper `captureException` was actually catching, without a not-found storm.

## Verification

287 suites / 1663 tests green. 19 new tests cover each fallback trigger, the timeout, option pass-through on the PUT, and both telemetry cases.

Also driven against the real plugin in Figma Desktop over the bridge — interceptor injected before the bundle loads, a key seeded into `clientStorage`, and the plugin relaunched so the genuine startup path ran. Responses stubbed at the network boundary; everything above it is the shipped bundle. Confirmed the running build was this branch's before trusting any of it.

```
━━━ gap ━━━  (today: the new API has no record yet)
   2061ms  GET api-production.tokens.studio  validate-license → 404 {"message":"License not found"}
   2073ms  GET licence.tokens.studio         validate-license → 200 {"plan":"pro","entitlements":["pro"]}
   2078ms  EVENT "License API fallback" primary=404/status fallback=200 migrationGap=true
  RESULT  status=verified plan="pro" entitlements=["pro"]

━━━ migrated ━━━  (after the import — the legacy stub was a 500 tripwire; it never fired)
   1558ms  GET api-production.tokens.studio  validate-license → 200 {"plan":"pro",...}
  RESULT  status=verified plan="pro" entitlements=["pro"]

━━━ downgrade ━━━  (half-imported record: 200, valid JSON, nothing in it)
   1723ms  GET api-production.tokens.studio  validate-license → 200 {"plan":null,"entitlements":[]}
   1734ms  GET licence.tokens.studio         validate-license → 200 {"plan":"pro","entitlements":["pro"]}
   1745ms  EVENT "License API fallback" primary=200/incomplete fallback=200 migrationGap=true
  RESULT  status=verified plan="pro" entitlements=["pro"]     ← the downgrade did not happen

━━━ nokey ━━━  (ordinary user with no license — not a migration gap)
    875ms  GET api-production.tokens.studio  validate-license → 404
    889ms  GET licence.tokens.studio         validate-license → 404
    891ms  EVENT "License API fallback" primary=404 fallback=404 migrationGap=false   ← excluded from the gate
  RESULT  status=error error="No license key found"

━━━ hang ━━━  (new API accepts the connection, never answers)
   8807ms  GET api-production.tokens.studio  validate-license → ABORTED after 8003ms
   8811ms  GET licence.tokens.studio         validate-license → 200 {"plan":"pro",...}
   8813ms  EVENT "License API fallback" primary=null/timeout fallback=200 migrationGap=true
  RESULT  status=verified plan="pro" entitlements=["pro"]
```

`AbortController` fired at 8003ms inside the real Figma plugin iframe, so the sandbox concern was unfounded. `detach-license` was deliberately not exercised live — it mutates real license state.

## Notes for whoever lands this

- **`detach-license` fails over like the reads do.** While both systems are live, a detach the primary accepts leaves the seat attached on the legacy host. Re-run the import as a delta sync before switching the old one off.
- **Both migration exits are commented in `licenseApiUrl.ts`**: when the legacy host is dropped from `LICENSE_API_HOSTS`, the `isComplete` guard in `validateLicense` has to go with it.
- Unstubbed, `api-production` currently answers a bogus key in **75ms** — about 7× faster than the legacy host's 533ms — but with a **500** rather than a 404, as does the legacy host. That doesn't affect this PR (any non-200 falls back), but it does mean the new API's error rate is 100% noise during the window. Being fixed separately on the API side.

### DIFF
--- a/.changeset/license-api-failover.md
+++ b/.changeset/license-api-failover.md
@@ -1,0 +1,5 @@
+---
+"@tokens-studio/figma-plugin": patch
+---
+
+Send license requests to the new production API first and fall back to the previous one when it cannot answer, so both can stay live while license data is migrated.
--- a/packages/tokens-studio-for-figma/.env.example
+++ b/packages/tokens-studio-for-figma/.env.example
@@ -1,4 +1,6 @@
 MIXPANEL_ACCESS_TOKEN=
 ENVIRONMENT=development
+# Legacy license API. Requests go to https://api-production.tokens.studio first
+# and only fall back here while the license data migration is in progress.
 LICENSE_API_URL=https://licence.tokens.studio
 TOKENS_STUDIO_APP_URL=production.tokens.studio
--- a/packages/tokens-studio-for-figma/.env.production.example
+++ b/packages/tokens-studio-for-figma/.env.production.example
@@ -1,4 +1,6 @@
 MIXPANEL_ACCESS_TOKEN=
 ENVIRONMENT=production
+# Legacy license API. Requests go to https://api-production.tokens.studio first
+# and only fall back here while the license data migration is in progress.
 LICENSE_API_URL=https://licence.tokens.studio
 TOKENS_STUDIO_APP_URL=production.tokens.studio
--- a/packages/tokens-studio-for-figma/src/mocks/handlers.ts
+++ b/packages/tokens-studio-for-figma/src/mocks/handlers.ts
@@ -1,5 +1,6 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { rest } from 'msw';
+import { LICENSE_API_HOSTS } from '@/utils/licenseApiUrl';
 
 export const LICENSE_FOR_VALID_RESPONSE = 'validate-c421c421-c421c421c-4c21c421-c421c42';
 export const LICENSE_FOR_ERROR_RESPONSE = 'validate-error-c421c421-c421c421c421421-41';
@@ -53,12 +54,17 @@ export const mockDetachLicenseHandler = jest.fn((req, res, ctx) => {
   return res();
 });
 
+// Both hosts are served by the same mocks so the failover resolves on the
+// primary by default. Tests that exercise the fallback override the primary
+// with server.use().
 export const handlers = [
-  rest.get(`${process.env.LICENSE_API_URL}/get-license`, mockGetLicenseHandler),
+  ...LICENSE_API_HOSTS.flatMap((host) => [
+    rest.get(`${host}/get-license`, mockGetLicenseHandler),
 
-  rest.get(`${process.env.LICENSE_API_URL}/validate-license`, mockValidateLicenseHandler),
+    rest.get(`${host}/validate-license`, mockValidateLicenseHandler),
 
-  rest.put(`${process.env.LICENSE_API_URL}/detach-license`, mockDetachLicenseHandler),
+    rest.put(`${host}/detach-license`, mockDetachLicenseHandler),
+  ]),
 
   rest.post(`${process.env.TOKEN_FLOW_APP_URL}/api/tokens`, (req, res, ctx) => res(ctx.status(200), ctx.json({ result: 'test-id' }))),
 ];
--- a/packages/tokens-studio-for-figma/src/utils/getLicenseKey.test.ts
+++ b/packages/tokens-studio-for-figma/src/utils/getLicenseKey.test.ts
@@ -1,12 +1,50 @@
 import * as Sentry from '@sentry/react';
 import { mockFetch } from '../../tests/__mocks__/fetchMock';
+import { LICENSE_API_HOSTS } from './licenseApiUrl';
 import getLicenseKey from './getLicenseKey';
+
+const [PRIMARY, LEGACY] = LICENSE_API_HOSTS;
+
+const json = (status: number, body: unknown) => Promise.resolve({ status, text: () => Promise.resolve(JSON.stringify(body)) });
 
 describe('getLicenseKey', () => {
   const sentrySpy = jest.spyOn(Sentry, 'captureException');
-  it('should call Sentry exception', async () => {
-    mockFetch.mockImplementationOnce(() => Promise.reject());
-    await getLicenseKey('jan');
-    expect(sentrySpy).toBeCalledTimes(1);
+  const sentryMessageSpy = jest.spyOn(Sentry, 'captureMessage');
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockFetch.mockReset();
+  });
+
+  it('should call Sentry when no host can be reached', async () => {
+    mockFetch.mockImplementation(() => Promise.reject());
+
+    const result = await getLicenseKey('jan');
+
+    expect(result).toEqual({ error: 'Error fetching license' });
+    expect(sentryMessageSpy).toBeCalledTimes(1);
+  });
+
+  it('should not call Sentry when the legacy host answers', async () => {
+    mockFetch
+      .mockImplementationOnce(() => Promise.reject())
+      .mockImplementationOnce(() => json(200, { key: 'a-license-key' }));
+
+    const result = await getLicenseKey('jan');
+
+    expect(result).toEqual({ key: 'a-license-key' });
+    expect(mockFetch).toHaveBeenNthCalledWith(1, `${PRIMARY}/get-license?userId=jan`, expect.anything());
+    expect(mockFetch).toHaveBeenNthCalledWith(2, `${LEGACY}/get-license?userId=jan`, expect.anything());
+    expect(sentrySpy).not.toBeCalled();
+    expect(sentryMessageSpy).not.toBeCalled();
+  });
+
+  it('should surface the last error message when neither host has the license', async () => {
+    mockFetch
+      .mockImplementationOnce(() => json(404, { message: 'Not found' }))
+      .mockImplementationOnce(() => json(404, { message: 'No license key found' }));
+
+    expect(await getLicenseKey('jan')).toEqual({ error: 'No license key found' });
+    expect(sentryMessageSpy).not.toBeCalled();
   });
 });
--- a/packages/tokens-studio-for-figma/src/utils/getLicenseKey.ts
+++ b/packages/tokens-studio-for-figma/src/utils/getLicenseKey.ts
@@ -1,24 +1,16 @@
-import { handleReactError } from './error/handleReactError';
+import { fetchLicenseApi } from './licenseApi';
 
 type GetLicenseKeyResponse =
   { key: string; }
   | { error: string; };
 
 export default async function getLicenseKey(userId: string | null): Promise<GetLicenseKeyResponse> {
-  try {
-    const res = await fetch(`${process.env.LICENSE_API_URL}/get-license?userId=${userId}`);
-    if (res.status === 200) {
-      return await res.json();
-    }
+  const result = await fetchLicenseApi<{ key: string }>(`/get-license?userId=${userId}`, {
+    endpoint: 'get-license',
+    figmaId: userId,
+  });
 
-    const { message } = await res.json();
-    return {
-      error: message,
-    };
-  } catch (e) {
-    handleReactError(e);
-    return {
-      error: 'Error fetching license',
-    };
-  }
+  if (result.ok) return result.data;
+
+  return { error: result.message ?? 'Error fetching license' };
 }
--- a/packages/tokens-studio-for-figma/src/utils/licenseApi.test.ts
+++ b/packages/tokens-studio-for-figma/src/utils/licenseApi.test.ts
@@ -1,0 +1,177 @@
+import * as Sentry from '@sentry/react';
+import { LICENSE_API_TIMEOUT_MS, fetchLicenseApi } from './licenseApi';
+import { LICENSE_API_HOSTS } from './licenseApiUrl';
+import { track } from './analytics';
+
+jest.mock('./analytics', () => ({ track: jest.fn() }));
+jest.mock('@sentry/react', () => ({ addBreadcrumb: jest.fn(), captureMessage: jest.fn() }));
+
+const [PRIMARY, LEGACY] = LICENSE_API_HOSTS;
+
+const json = (status: number, body: unknown) => ({ status, text: async () => JSON.stringify(body) });
+const raw = (status: number, text: string) => ({ status, text: async () => text });
+
+// Never settles on its own — only when the timeout aborts it.
+const hangs = () => (_url: string, init: RequestInit) => new Promise((_resolve, reject) => {
+  init.signal?.addEventListener('abort', () => {
+    const error = new Error('Aborted');
+    error.name = 'AbortError';
+    reject(error);
+  });
+});
+
+const mockFetch = (...responses: any[]) => {
+  const fetchMock = jest.fn();
+  responses.forEach((response) => {
+    fetchMock.mockImplementationOnce(
+      typeof response === 'function' ? response : () => Promise.resolve(response),
+    );
+  });
+  global.fetch = fetchMock as any;
+  return fetchMock;
+};
+
+const lastTrackCall = () => (track as jest.Mock).mock.calls[(track as jest.Mock).mock.calls.length - 1];
+
+describe('fetchLicenseApi', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('resolves on the primary host without touching the legacy one', async () => {
+    const fetchMock = mockFetch(json(200, { key: 'abc' }));
+
+    const result = await fetchLicenseApi<{ key: string }>('/get-license?userId=1', { endpoint: 'get-license' });
+
+    expect(result).toMatchObject({ ok: true, data: { key: 'abc' } });
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    expect(fetchMock.mock.calls[0][0]).toEqual(`${PRIMARY}/get-license?userId=1`);
+    expect(track).not.toHaveBeenCalled();
+  });
+
+  it.each([
+    ['a 500', json(500, { message: 'boom' })],
+    ['a 404', json(404, { message: 'No license key found' })],
+    ['a 403', json(403, { message: 'Forbidden' })],
+    ['an unparseable 200', raw(200, '<html>gateway</html>')],
+    ['an empty 200', raw(200, '')],
+  ])('falls back to the legacy host on %s from the primary', async (_label, primaryResponse) => {
+    const fetchMock = mockFetch(primaryResponse, json(200, { key: 'abc' }));
+
+    const result = await fetchLicenseApi<{ key: string }>('/get-license?userId=1', { endpoint: 'get-license' });
+
+    expect(result).toMatchObject({ ok: true, data: { key: 'abc' } });
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+    expect(fetchMock.mock.calls[1][0]).toEqual(`${LEGACY}/get-license?userId=1`);
+  });
+
+  it('falls back when the primary cannot be reached at all', async () => {
+    const fetchMock = mockFetch(() => Promise.reject(new Error('Network request failed')), json(200, { key: 'abc' }));
+
+    const result = await fetchLicenseApi<{ key: string }>('/get-license?userId=1', { endpoint: 'get-license' });
+
+    expect(result).toMatchObject({ ok: true });
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+    expect(lastTrackCall()[1]).toMatchObject({ primaryReason: 'transport', primaryStatus: null });
+  });
+
+  it('gives up on a hanging primary and falls back', async () => {
+    jest.useFakeTimers();
+    const fetchMock = mockFetch(hangs(), json(200, { key: 'abc' }));
+
+    const pending = fetchLicenseApi<{ key: string }>('/get-license?userId=1', { endpoint: 'get-license' });
+    await Promise.resolve();
+    jest.advanceTimersByTime(LICENSE_API_TIMEOUT_MS);
+
+    const result = await pending;
+
+    expect(result).toMatchObject({ ok: true, data: { key: 'abc' } });
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+    expect(lastTrackCall()[1]).toMatchObject({ primaryReason: 'timeout' });
+    jest.useRealTimers();
+  });
+
+  it('falls back when the primary answers 200 with an incomplete payload', async () => {
+    const fetchMock = mockFetch(
+      json(200, { plan: null, entitlements: [] }),
+      json(200, { plan: 'pro', entitlements: ['pro'] }),
+    );
+
+    const result = await fetchLicenseApi<{ plan?: string; entitlements?: string[] }>('/validate-license', {
+      endpoint: 'validate-license',
+      isComplete: (data) => Boolean(data?.plan) || (data?.entitlements?.length ?? 0) > 0,
+    });
+
+    expect(result).toMatchObject({ ok: true, data: { plan: 'pro' } });
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+    expect(lastTrackCall()[1]).toMatchObject({ primaryReason: 'incomplete', primaryStatus: 200 });
+  });
+
+  it('surfaces the legacy host error message when both hosts fail', async () => {
+    mockFetch(json(404, { message: 'not found on new' }), json(404, { message: 'No license key found' }));
+
+    const result = await fetchLicenseApi('/get-license?userId=1', { endpoint: 'get-license' });
+
+    expect(result).toMatchObject({ ok: false, message: 'No license key found' });
+  });
+
+  it('returns a null message when neither host produced one', async () => {
+    mockFetch(() => Promise.reject(new Error('nope')), () => Promise.reject(new Error('nope')));
+
+    const result = await fetchLicenseApi('/get-license?userId=1', { endpoint: 'get-license' });
+
+    expect(result).toMatchObject({ ok: false, message: null });
+  });
+
+  it('passes request options through to every host', async () => {
+    const fetchMock = mockFetch(json(500, { message: 'boom' }), json(200, 'detached'));
+
+    await fetchLicenseApi<string>('/detach-license', {
+      endpoint: 'detach-license',
+      init: { method: 'PUT', headers: { 'Content-Type': 'application/json' }, body: '{"licenseKey":"a"}' },
+    });
+
+    expect(fetchMock.mock.calls[1][1]).toMatchObject({ method: 'PUT', body: '{"licenseKey":"a"}' });
+  });
+
+  describe('telemetry', () => {
+    it('flags a migration gap when only the legacy host can answer', async () => {
+      mockFetch(json(404, { message: 'not found' }), json(200, { key: 'abc' }));
+
+      await fetchLicenseApi('/get-license?userId=1', { endpoint: 'get-license', figmaId: 'figma-user-1' });
+
+      expect(track).toHaveBeenCalledWith('License API fallback', expect.objectContaining({
+        endpoint: 'get-license',
+        primaryStatus: 404,
+        primaryReason: 'status',
+        fallbackStatus: 200,
+        resolvedBy: LEGACY,
+        migrationGap: true,
+        figmaId: 'figma-user-1',
+      }));
+    });
+
+    // Users with no license 404 on both hosts forever. Counting them as gaps
+    // would leave the cutover metric with a floor it can never clear.
+    it('does not flag a migration gap when neither host has the record', async () => {
+      mockFetch(json(404, { message: 'not found' }), json(404, { message: 'No license key found' }));
+
+      await fetchLicenseApi('/get-license?userId=1', { endpoint: 'get-license' });
+
+      expect(lastTrackCall()[1]).toMatchObject({ migrationGap: false, resolvedBy: null });
+    });
+
+    it('reports to Sentry only when every host is unreachable', async () => {
+      mockFetch(json(404, { message: 'a' }), json(404, { message: 'b' }));
+      await fetchLicenseApi('/get-license?userId=1', { endpoint: 'get-license' });
+      expect(Sentry.captureMessage).not.toHaveBeenCalled();
+
+      mockFetch(() => Promise.reject(new Error('down')), () => Promise.reject(new Error('down')));
+      await fetchLicenseApi('/get-license?userId=1', { endpoint: 'get-license' });
+      expect(Sentry.captureMessage).toHaveBeenCalledWith(
+        'License API unreachable: get-license',
+        expect.objectContaining({ level: 'error' }),
+      );
+    });
+  });
+});
--- a/packages/tokens-studio-for-figma/src/utils/licenseApi.ts
+++ b/packages/tokens-studio-for-figma/src/utils/licenseApi.ts
@@ -1,0 +1,164 @@
+import * as Sentry from '@sentry/react';
+import { track } from './analytics';
+import { LICENSE_API_HOSTS } from './licenseApiUrl';
+
+// There is no timeout on fetch by default. Without one an unreachable host
+// hangs the startup process forever instead of falling back.
+export const LICENSE_API_TIMEOUT_MS = 8000;
+
+type FailureReason =
+  | 'transport' // network error, DNS, CORS
+  | 'timeout' // exceeded LICENSE_API_TIMEOUT_MS
+  | 'status' // responded, but not with a 200
+  | 'parse' // 200 with a body we could not read as JSON
+  | 'incomplete'; // 200, valid JSON, but the payload failed the caller's check
+
+export type LicenseApiAttempt = {
+  host: string;
+  status: number | null;
+  reason: FailureReason | null;
+  message?: string;
+};
+
+export type LicenseApiResult<T> =
+  | { ok: true; data: T; attempts: LicenseApiAttempt[] }
+  | { ok: false; message: string | null; attempts: LicenseApiAttempt[] };
+
+type LicenseApiOptions<T> = {
+  /**
+   * Marks the endpoint in telemetry. Not the raw path — that carries the
+   * license key and user id.
+   */
+  endpoint: string;
+  init?: RequestInit;
+  /**
+   * Hashed before it is sent. Lets a migration gap be traced back to the
+   * account whose record did not make it across.
+   */
+  figmaId?: string | null;
+  /**
+   * Guard against a half-migrated record answering 200 with a valid-looking
+   * but empty payload, which would otherwise silently downgrade the user.
+   * Returning false falls through to the next host.
+   */
+  isComplete?: (data: T) => boolean;
+};
+
+async function attemptHost<T>(
+  host: string,
+  path: string,
+  { init, isComplete }: Pick<LicenseApiOptions<T>, 'init' | 'isComplete'>,
+): Promise<LicenseApiAttempt & { data?: T }> {
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), LICENSE_API_TIMEOUT_MS);
+
+  try {
+    const res = await fetch(`${host}${path}`, { ...init, signal: controller.signal });
+    // Read the body once — res.json() cannot be called twice, and both the
+    // success and the error path need it.
+    const text = await res.text();
+
+    let body: any;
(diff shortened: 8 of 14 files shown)


anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **1 finding(s) across 4 file(s)**

| 🟡 MEDIUM | Access Control | `src/plugin/controller.ts` | Hidden runtime override of license/entitlement API endpoint |

<details>

**[Access Control] Hidden runtime override of license/entitlement API endpoint**
- **File**: `packages/tokens-studio-for-figma/src/plugin/controller.ts` (lines ~104-136), plus `src/app/internal/registerLicenseApiConsoleCommands.ts`, `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts`, `src/utils/licenseApiUrl.ts`, `src/figmaStorage/LicenseApiUrlOverrideProperty.ts`
- **Implication**: Adds hidden developer console commands (`__setTokensStudioLicenseApiUrl` / `__resetTokensStudioLicenseApiUrl`) that persist an override of the license API base URL to Figma client storage. All license calls — `get-license`, `detach-license`, `validate-license` — carry `userId`, `licenseKey`, `userName` and receive `email`, so a persisted override can reroute personal-data/credential-bearing requests to an arbitrary host. The commands are registered unconditionally at startup (no production/build guard visible), and the override survives across sessions.
- **Recommended action**: Before merging, confirm the override is gated to non-production/dev builds, validate/allowlist permitted target hosts (reject arbitrary domains), and record in the ROPA that license endpoints process personal data (user id, name, email, licence key) so a redirection risk is documented.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **2 finding(s) across 3 file(s)**

| 🟡 MEDIUM | Access Control | `src/utils/licenseApiUrl.ts` | Unrestricted runtime override of license API base URL (no domain allow-list) |
| 🟢 LOW | Policy Gap | `src/utils/validateLicense.ts` | Personal data (email, userId, userName, licenseKey) sent as URL query params to an overridable endpoint |

<details>

**[Access Control] Unrestricted runtime override of license API base URL (no domain allow-list)**
- **File**: `src/utils/licenseApiUrl.ts` (lines 1-46), `src/plugin/controller.ts` (lines 104-136), `src/app/internal/registerLicenseApiConsoleCommands.ts` (lines 1-47)
- **Implication**: A hidden, non-enumerable console command (`__setTokensStudioLicenseApiUrl`) lets any caller repoint all license API traffic to an arbitrary `http(s)` host, persisted to `clientStorage` across reloads. `normalizeLicenseApiUrl` validates only scheme/format, not destination — there is no allow-list restricting the target to tokens.studio staging hosts. Because the license endpoints transmit personal data (see finding 2), a redirect (e.g. via a script injected into the iframe context or a social-engineered console paste) could exfiltrate that data to a non-DPA-covered / attacker-controlled endpoint, outside subprocessor governance and with no audit trail. The command is intentionally hidden from `Object.keys`/tab-complete, reducing discoverability of the control.
- **Recommended action**: Restrict the override to an allow-list of approved tokens.studio staging domains, and/or gate the entire feature behind a development build flag so it is not compiled into production/public releases. Retain a persistent, visible indicator and log when a non-default URL is active.

**[Policy Gap] Personal data (email, userId, userName, licenseKey) sent as URL query params to an overridable endpoint**
- **File**: `src/utils/validateLicense.ts` (lines 9-14), `src/utils/getLicenseKey.ts` (lines 9-11)
- **Implication**: License requests pass `licenseKey`, `userId` and `userName` (and return `email`) as query-string parameters. Query strings are commonly retained in server/proxy/CDN access logs. Combined with the runtime override above, the base host receiving this personal data can be changed at runtime, widening the exposure surface for GDPR-relevant identifiers.
- **Recommended action**: Move personal-data parameters from the query string into the POST body/headers, and confirm any host reachable via the override is covered by an executed DPA and logged in the ROPA/subprocessor register.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **3 finding(s) across 3 file(s)**

| 🟡 MEDIUM | Access Control | `src/utils/licenseApiUrl.ts` | Runtime override repoints license/entitlement validation endpoint via hidden console commands |
| 🟡 MEDIUM | Subprocessor | `src/utils/validateLicense.ts` | Personal data (userId, userName, licenseKey) sent to a runtime-overridable, potentially unvetted endpoint |
| 🟢 LOW | Policy Gap | `src/plugin/controller.ts` | Hidden, non-enumerable override with no audit trail and a hardcoded default endpoint |

<details>

**[Access Control] Runtime override repoints license/entitlement validation endpoint via hidden console commands**
- **File**: `src/utils/licenseApiUrl.ts` (lines 30-46; enablers in `src/plugin/controller.ts` lines 105-136, `src/app/internal/registerLicenseApiConsoleCommands.ts` lines 24-47, `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts`)
- **Implication**: License validation, get-license and detach-license calls now resolve their target host at runtime from a `clientStorage`-persisted override (`getLicenseApiUrl()`) instead of the compiled `LICENSE_API_URL` env var. Anyone with DevTools access to the plugin sandbox/iframe can invoke the hidden `__setTokensStudioLicenseApiUrl(...)` command to repoint entitlement checks at an arbitrary backend, bypassing the intended license authority. This is an access-control / entitlement-gating change.
- **Recommended action**: Before merging, confirm the override is restricted to internal/non-production builds (gate behind a dev/staging flag) or formally document it as an accepted internal-testing control, and record it in the access-control register with the reset-to-default behaviour verified.

**[Subprocessor] Personal data (userId, userName, licenseKey) sent to a runtime-overridable, potentially unvetted endpoint**
- **File**: `src/utils/validateLicense.ts` (lines 9-14; same pattern in `src/utils/getLicenseKey.ts` and `src/utils/removeLicense.ts`)
- **Implication**: `validateLicense` transmits `licenseKey`, `userId` and `userName` as URL query-string parameters (and receives `email`) to whatever host the override points at. Combined with the new runtime override this creates a path for personal data to be transmitted to a domain that is not a vetted subprocessor, and query-string PII is additionally exposed in transit and server/access logs.
- **Recommended action**: Before merging, ensure any non-default override target is an approved subprocessor, and consider moving `licenseKey`/`userId`/`userName` from the query string into a POST body; confirm the license API endpoint is captured in the subprocessor/ROPA register.

**[Policy Gap] Hidden, non-enumerable override with no audit trail and a hardcoded default endpoint**
- **File**: `src/plugin/controller.ts` (lines 105-136; default at `src/utils/licenseApiUrl.ts` lines 23-25)
- **Implication**: The override commands are intentionally hidden (registered as non-enumerable `window`/`globalThis` properties, no UI) and write the endpoint to `clientStorage` without any log or audit record of who changed it or when. The default authority `https://licence.tokens.studio` is hardcoded as a fallback. No secret/credential was committed, but the hidden mechanism lacks a traceable audit trail for endpoint changes.
- **Recommended action**: Before merging, confirm the hardcoded default is the intended production authority, and add a log/audit entry whenever the override is set or reset so endpoint changes are traceable for SOC2 change-management evidence.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **2 finding(s) across 2 file(s)**

| 🟡 MEDIUM | Subprocessor | `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` | License API endpoint becomes runtime-overridable to any domain with no allowlist |
| 🟡 MEDIUM | Policy Gap | `packages/tokens-studio-for-figma/src/plugin/controller.ts` | Hidden, undocumented dev commands persist a license-API-URL override to client storage |

<details>

**[Subprocessor] License API endpoint becomes runtime-overridable to any domain with no allowlist**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` (lines 23-45)
- **Implication**: The license API base URL (default `https://licence.tokens.studio`, previously the fixed `process.env.LICENSE_API_URL`) is now resolved through `getLicenseApiUrl()`, which returns a mutable override. `normalizeLicenseApiUrl` accepts any `http`/`https`/`localhost` host with no domain allowlist. This URL is used by `validateLicense.ts`, `getLicenseKey.ts` and `removeLicense.ts`, which transmit personal data (userId, licenseKey, userName) and receive email in the response. An override therefore allows license-related PII to be routed to an arbitrary external endpoint outside the approved subprocessor list, and the persisted override survives restarts (Figma client storage), which is a GDPR Art. 28/44 (subprocessor / cross-border transfer) and SOC2 CC9.2 concern.
- **Recommended action**: Constrain overrides to an explicit allowlist of Tokens Studio-controlled hosts (or gate strictly to non-production/dev builds), confirm the license endpoint and the PII it processes are recorded in the ROPA and subprocessor register, and add a data-flow note that overrides must never point at unapproved domains in production.

**[Policy Gap] Hidden, undocumented dev commands persist a license-API-URL override to client storage**
- **File**: `packages/tokens-studio-for-figma/src/plugin/controller.ts` (lines 104-136; also `src/app/internal/registerLicenseApiConsoleCommands.ts`)
- **Implication**: Non-enumerable global commands (`__setTokensStudioLicenseApiUrl` / `__resetTokensStudioLicenseApiUrl` / `__getTokensStudioLicenseApiUrl`) are registered in both the plugin sandbox and UI scope and persist the override via `LicenseApiUrlOverrideProperty` (Figma CLIENT_STORAGE) with no user disclosure, consent, or audit trail. A hidden mechanism that silently changes where license/PII traffic is sent is a transparency/governance gap (GDPR Art. 5(1)(a); SOC2 CC5/CC9.2) and could be abused to exfiltrate license data.
- **Recommended action**: Document these developer hooks in an internal control record, ensure they are stripped or hard-disabled in production builds, and log/surface any active override so it is auditable.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **3 finding(s) across 4 file(s)**

| 🔴 HIGH | Secrets / Subprocessor | `src/utils/licenseApiUrl.ts` | Hidden, ungated runtime override lets license API base URL be repointed to any host |
| 🟡 MEDIUM | Policy Gap | `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts` | Endpoint change that redirects user PII produces no audit trail |
| 🟢 LOW | Access Control | `src/app/asyncMessageHandlers/startup.tsx` | Override applied from any `message` event without origin validation |

<details>

**[Secrets / Subprocessor] Hidden runtime override for the license API URL**
- **File**: `src/utils/licenseApiUrl.ts` (new, L1–46); `src/plugin/controller.ts` (L104–136, hidden `__setTokensStudioLicenseApiUrl`/`__reset`/`__get`); `src/app/internal/registerLicenseApiConsoleCommands.ts` (new, L1–47); `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts` (new); used by `src/utils/validateLicense.ts` (L11), `getLicenseKey.ts` (L10), `removeLicense.ts` (L13)
- **Implication**: A hidden, undocumented runtime override lets the license-validation API base URL be repointed to an arbitrary http(s) host with **no allowlist**. The three license fetches carry personal data — `validateLicense` sends `licenseKey`, `userId`, `userName` and receives `email`; `getLicenseKey` sends `userId`; `removeLicense` sends `licenseKey`+`userId`. `normalizeLicenseApiUrl` accepts any host (only enforces http/https, and downgrades localhost to plain http). Registration is **not** gated by a dev/NODE_ENV flag — `registerLicenseApiConsoleCommands()` runs unconditionally in `startup()` and controller.ts registers sandbox commands via an unconditional IIFE, so these ship in the production plugin. The override persists in CLIENT_STORAGE and re-applies on every startup, creating a silent PII-exfiltration channel for anyone with dev-console access.
- **Recommended action**: Restrict the override to development/staging builds (guard on `process.env.NODE_ENV`), enforce an allowlist of approved license hosts in `normalizeLicenseApiUrl`, remove the localhost→http downgrade for prod, and confirm the license endpoint and its data fields are recorded as an approved subprocessor in the ROPA.

**[Policy Gap] No audit trail for endpoint redirection**
- **File**: `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts` (L15–19); `src/figmaStorage/LicenseApiUrlOverrideProperty.ts` (new)
- **Implication**: Changing where user PII is transmitted produces only a transient `figma.notify` toast — no log, no consent/notice, no tamper record. Fails SOC2 change/audit-trail expectations for a security-relevant configuration event.
- **Recommended action**: Emit an auditable log event on every override set/reset (old→new host, timestamp, actor context); consider requiring explicit confirmation.

**[Access Control] Override applied without message-origin validation**
- **File**: `src/app/asyncMessageHandlers/startup.tsx` (L32–46, `handleSandboxOverride` `window.addEventListener('message', …)`)
- **Implication**: The UI applies a license-URL override from any `message` event with `data.pluginMessage.type === '__LICENSE_URL_OVERRIDE__'`, with no origin/authenticity check. Low risk given Figma's iframe messaging model, but the magic-string trigger is unauthenticated.
- **Recommended action**: Validate message origin/source before applying, or gate the listener to dev builds.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **2 finding(s) across 2 file(s)**

| 🟡 MEDIUM | Access Control | `src/plugin/controller.ts` | Hidden runtime command redirects license API (carrying userId/email) to arbitrary host |
| 🟢 LOW | Policy Gap | `src/utils/licenseApiUrl.ts` | Override accepts any http/https host with no allowlist or audit trail |

<details>

**[Access Control] Hidden runtime command redirects license API (carrying userId/email) to arbitrary host**
- **File**: `src/plugin/controller.ts` (lines ~104-136); related `src/app/internal/registerLicenseApiConsoleCommands.ts`, `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts`, `src/figmaStorage/LicenseApiUrlOverrideProperty.ts`
- **Implication**: New hidden, non-enumerable developer commands (`__setTokensStudioLicenseApiUrl` etc.) let anyone with DevTools access persist (in clientStorage) an override of the license API base URL. All three license calls (`validateLicense`, `getLicenseKey`, `removeLicense`) transmit personal data — `userId`, `userName`, `licenseKey` — and `validateLicense` returns `email`. Because the destination host is now runtime-configurable to any origin, personal data could be routed to an unintended/third-party endpoint, and this new privileged surface is not gated by role or audit logging. This is an access-control / data-flow (subprocessor destination) change relevant to GDPR Art. 28/32 and SOC2 CC6.
- **Recommended action**: Restrict overrides to an internal allowlist (e.g. `*.tokens.studio`), gate behind a non-production/build flag so the commands are stripped from release builds, and log override changes for audit. Confirm no ROPA/subprocessor destination is affected in production.

**[Policy Gap] Override accepts any http/https host with no allowlist or audit trail**
- **File**: `src/utils/licenseApiUrl.ts` (lines 1-46)
- **Implication**: `normalizeLicenseApiUrl` accepts any `http(s)` (or localhost) host with no domain allowlist; the only user-facing signal is a transient `figma.notify`. There is no persistent audit record of when/where the license endpoint (which carries personal data) was repointed.
- **Recommended action**: Add an allowlist check and emit an auditable log/telemetry event when the override is set or cleared.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **1 finding across 1 file**

| 🟡 MEDIUM | Policy Gap | `src/plugin/controller.ts` | Hidden runtime override can redirect the license API (carrying userId / license key / name) to an arbitrary URL |

<details>

**[Policy Gap] Hidden runtime override for the license API endpoint**
- **File**: `packages/tokens-studio-for-figma/src/plugin/controller.ts` (lines ~104-136), with supporting logic in `src/app/internal/registerLicenseApiConsoleCommands.ts`, `src/utils/licenseApiUrl.ts`, `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts`, and `src/figmaStorage/LicenseApiUrlOverrideProperty.ts`
- **Implication**: Adds undocumented sandbox/console commands (`__setTokensStudioLicenseApiUrl`, `__resetTokensStudioLicenseApiUrl`, `__getTokensStudioLicenseApiUrl`) that persist an override of the license API base URL to client storage. The license endpoints (`validateLicense`, `getLicenseKey`, `removeLicense`) transmit personal data — `userId`, `userName`, and the license key — so an override allows those PII-bearing requests to be redirected to an arbitrary external host at runtime. This is a data-flow control with no audit trail or user-facing disclosure. The default endpoint (`https://licence.tokens.studio`) is unchanged, and no hardcoded secret is introduced.
- **Recommended action**: Confirm the override hooks are dev-only and cannot be invoked in shipped/production builds (guard behind a build flag), document the mechanism, and verify that permitting redirection of a PII-bearing endpoint is acceptable under Hyma's data-flow / transparency policy before merging.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **5 finding(s) across 6 file(s)**

| 🟡 MEDIUM | Access Control | `src/plugin/controller.ts` | Ungated global console commands repoint the license API in production builds |
| 🟡 MEDIUM | Subprocessor | `src/utils/licenseApiUrl.ts` | License API host becomes runtime-configurable to any http/https origin |
| 🟡 MEDIUM | Secrets | `src/utils/validateLicense.ts` | License key and userId sent as query params to a now-configurable host |
| 🟢 LOW | ROPA | `src/figmaStorage/LicenseApiUrlOverrideProperty.ts` | New `licenseApiUrlOverride` persisted in Figma clientStorage |
| 🟢 LOW | Policy Gap | `src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts` | Endpoint change produces a toast but no audit record |

<details>

**[Access Control] Ungated global console commands repoint the license API in production builds**
- **File**: `packages/tokens-studio-for-figma/src/plugin/controller.ts` (lines 104–136) and `packages/tokens-studio-for-figma/src/app/internal/registerLicenseApiConsoleCommands.ts` (lines 1–47)
- **Implication**: `registerSandboxDevCommands()` runs unconditionally at controller init, and `registerLicenseApiConsoleCommands()` is called unconditionally from `startup()`. There is no `NODE_ENV`/dev-build guard, no authentication and no confirmation step, so `__setTokensStudioLicenseApiUrl` ships to every end user. Any code or actor with console access in the Figma sandbox or the plugin UI iframe can silently redirect all license traffic, and the value persists across plugin reloads via `clientStorage`. This is a logical-access control change that takes effect on deploy (SOC2 CC6).
- **Recommended action**: Gate both registrations behind a dev/internal build flag before merging. If the commands must ship in production, require an explicit confirmation and restrict targets to an allowlist.

**[Subprocessor] License API host becomes runtime-configurable to any http/https origin**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` (lines 1–46)
- **Implication**: `normalizeLicenseApiUrl` accepts any value that resolves to `http://` or `https://` — there is no host allowlist. `getLicenseApiUrl()` is now the single source for `validateLicense`, `getLicenseKey` and `removeLicense`, which transmit the license key, Figma `userId` and `userName` and receive the user's **email** in the response. Redirecting these calls sends personal data to an entity that is not on Hyma's subprocessor register and has no DPA in place.
- **Recommended action**: Restrict `normalizeLicenseApiUrl` to an explicit allowlist (`licence.tokens.studio`, `api-staging.tokens.studio`, localhost). Separately, confirm the compiled default host is already recorded on the subprocessor register and disclosed in the privacy policy.

**[Secrets] License key and userId sent as query params to a now-configurable host**
- **File**: `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` (lines 11–14) and `packages/tokens-studio-for-figma/src/utils/getLicenseKey.ts` (line 10)
- **Implication**: `licenseKey`, `userId` and `userName` remain in the URL query string; this diff changes the destination of those URLs from a build-time constant to a runtime-mutable value. Query strings are routinely written to proxy and server access logs, so combined with the override a live licence credential and user identifiers can be persisted in logs outside Hyma's control. Any credential that reaches a non-production host is exposed.
- **Recommended action**: Move `licenseKey` into a POST body or an `Authorization` header rather than the query string. Rotate any license key known to have been sent to a non-production host during testing.

**[ROPA] New `licenseApiUrlOverride` persisted in Figma clientStorage**
- **File**: `packages/tokens-studio-for-figma/src/figmaStorage/LicenseApiUrlOverrideProperty.ts` (lines 1–6); read into the startup payload in `packages/tokens-studio-for-figma/src/utils/plugin/startup.ts` (lines 33–95)
- **Implication**: A new value is persisted client-side and returned in every `STARTUP` message. It is configuration rather than personal data, but it is a new persisted item with no retention or clearing path other than the manual `__resetTokensStudioLicenseApiUrl()` command.
- **Recommended action**: Add to the plugin local-storage section of the Studio Platform ROPA entry, or record a note confirming it is out of scope as non-personal configuration data.

**[Policy Gap] Endpoint change produces a toast but no audit record**
- **File**: `packages/tokens-studio-for-figma/src/plugin/asyncMessageHandlers/setLicenseApiUrlOverride.ts` (lines 7–25)
- **Implication**: The only signal that license traffic has been redirected is a transient `figma.notify` toast. No durable audit entry is emitted, so there is no way after the fact to establish whether a given session sent license and identity data to a non-production endpoint (SOC2 CC7 monitoring).
- **Recommended action**: Emit a Sentry breadcrumb or analytics event whenever the override is set or cleared, including the resulting effective host.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **4 finding(s) across 4 file(s)**

| 🟡 MEDIUM | Subprocessor | `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` | New primary license API host `api-production.tokens.studio` hardcoded as first destination for all license traffic |
| 🟡 MEDIUM | ROPA | `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` | License key, Figma user id and user name are now sent to a second host, in URL query parameters |
| 🟡 MEDIUM | Subprocessor | `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` | `figmaId` added to the analytics fallback event; the doc comment says it is hashed, the code sends it raw |
| 🟢 LOW | ROPA | `packages/tokens-studio-for-figma/src/utils/removeLicense.ts` | Detach failover can leave the seat attached on the legacy host, so a removal is not applied consistently |

<details>

**[Subprocessor] New primary license API host hardcoded as the first destination for all license traffic**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` (lines 24–45), also `packages/tokens-studio-for-figma/.env.example` and `.env.production.example` (lines 1–6)
- **Implication**: `PRIMARY_LICENSE_API_URL` is a literal `https://api-production.tokens.studio`, and `LICENSE_API_HOSTS` puts it ahead of the existing `LICENSE_API_URL`. Every `get-license`, `validate-license` and `detach-license` call now reaches a host that was not previously contacted by the plugin. Unlike the legacy host it is not environment-configurable, so it cannot be pointed elsewhere per deployment. If this endpoint terminates on different infrastructure (region, hosting provider, or operator) than `licence.tokens.studio`, that is a change of processing location for licence and account data and needs to be reflected in the ROPA and, where a third party operates it, in the subprocessor list.
- **Recommended action**: Before merging, confirm the hosting region and operator of `api-production.tokens.studio` and record it in the ROPA transfer/hosting section; add it to the subprocessor register if it is operated by a party not already listed. Consider making the primary host configurable via env like the legacy one, so non-production builds do not reach production licence infrastructure.

**[ROPA] License key, Figma user id and user name are now sent to a second host, in URL query parameters**
- **File**: `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` (lines 1–27); same pattern in `getLicenseKey.ts` (lines 1–16) and `removeLicense.ts` (lines 1–23)
- **Implication**: `fetchLicenseApi` replays the identical request path against each host in turn. The validate path carries `licenseKey`, `userId` and `userName` as query-string parameters, and the response carries `email`. A failover therefore discloses those identifiers to the legacy host as well as the primary — two sets of access logs, TLS terminators and any intermediate proxies now hold them, where previously there was one. Query-string parameters are the most log-exposed part of a request and are routinely retained in web-server and CDN logs; carrying a licence key and a user name there is a pre-existing weakness that this change duplicates rather than introduces.
- **Recommended action**: After merging, confirm both hosts' access-log retention is covered by the ROPA retention entry for licence data. Separately, consider moving `licenseKey`/`userId`/`userName` from the query string into a POST body or headers so they stop landing in access logs on either host.

**[Subprocessor] `figmaId` added to the analytics fallback event; the doc comment says it is hashed, the code sends it raw**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` (lines 30–40 for the doc comment, lines 100–120 in `reportFallback`)
- **Implication**: The `figmaId` option is documented as "Hashed before it is sent", but `reportFallback` spreads it into the `track('License API fallback', …)` payload unchanged: `...(figmaId ? { figmaId } : {})`. `getLicenseKey`, `validateLicense` and `removeLicense` all pass the real Figma user id. This sends a stable per-user identifier to the analytics destination (Mixpanel, per `MIXPANEL_ACCESS_TOKEN` in `.env.example`) on every failover — a new category of personal data to an existing processor, and a control that the code claims to have but does not. The accompanying test asserts the raw value (`figmaId: 'figma-user-1'`), so the gap is locked in by the suite rather than caught by it.
- **Recommended action**: Before merging, either implement the hashing the comment promises (and update `licenseApi.test.ts` to assert the digest rather than the raw id) or drop `figmaId` from the analytics payload and rely on the Sentry breadcrumb, which carries no identifier. If the raw id is genuinely needed to trace migration gaps, record the added identifier against the analytics processing activity in the ROPA and confirm it is in scope for the existing analytics retention and deletion path.

**[ROPA] Detach failover can leave the seat attached on the legacy host**
- **File**: `packages/tokens-studio-for-figma/src/utils/removeLicense.ts` (lines 1–23)
- **Implication**: The code comment states the case plainly: while both systems are live, a detach the primary accepts leaves the seat attached on the legacy host. `fetchLicenseApi` stops at the first clean answer, so a successful detach on the primary never reaches the legacy record. A user who removes their licence is told it succeeded while the association persists in the other system, which is a divergence between what the product reports and what is actually stored.
- **Recommended action**: After merging, track the delta-sync/reconciliation step the comment already calls for as an explicit migration exit criterion alongside the `migrationGap` telemetry, so the legacy host is not switched off with stale attachments. If detach is ever used to service an erasure or account-closure request, confirm the reconciliation runs before the request is reported as complete.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **6 finding(s) across 5 file(s)**

| 🟡 MEDIUM | Subprocessor | `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` | New external licence host `api-production.tokens.studio` introduced as the primary endpoint |
| 🟡 MEDIUM | ROPA | `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` | Licence key, Figma user ID and user name transmitted as URL query parameters to a new host |
| 🟡 MEDIUM | Subprocessor | `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` | New `License API fallback` analytics event sends `figmaId` and licence-state metadata to the analytics processor |
| 🟡 MEDIUM | Access Control | `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` | Plan and entitlement resolution now fails over across two independent licence datastores |
| 🟡 MEDIUM | Access Control | `packages/tokens-studio-for-figma/src/utils/removeLicense.ts` | Seat detach fails over, so a licence can remain attached on the host that did not process it |
| 🟢 LOW | Policy Gap | `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` | Per-call error reporting narrowed to the "all hosts unreachable" case |

<details>

**[Subprocessor] New external licence host `api-production.tokens.studio` introduced as the primary endpoint**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` (lines 23–45)
- **Implication**: Every licence read and write from the plugin is now sent to a new external host that is not derived from `LICENSE_API_URL` — it is hardcoded and cannot be disabled by configuration. This is a new outbound destination for licence and user-identifying data and a new processing location that the subprocessor register and the ROPA transfer section do not currently describe.
- **Recommended action**: Before merging, confirm `api-production.tokens.studio` is covered by the same controller/processor arrangement and hosting region as `licence.tokens.studio`, and add or amend the subprocessor register entry accordingly.

**[ROPA] Licence key, Figma user ID and user name transmitted as URL query parameters to a new host**
- **File**: `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` (lines 11–22); also `packages/tokens-studio-for-figma/src/utils/getLicenseKey.ts` (lines 7–10)
- **Implication**: `licenseKey`, `userId` and `userName` are placed in the query string, and the validate response carries `email`. Query strings are routinely captured in edge, proxy and web-server access logs, so this personal data is now duplicated into the log estate of a second host whose retention and deletion schedule has not yet been recorded.
- **Recommended action**: Before merging, confirm the retention and deletion schedule for request logs on the new host and record the additional processing location in the ROPA, and after merging consider moving these parameters into request headers or a request body.

**[Subprocessor] New `License API fallback` analytics event sends `figmaId` and licence-state metadata to the analytics processor**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` (lines 100–118)
- **Implication**: A new event category is emitted to the existing analytics subprocessor containing a pseudonymised user identifier alongside `endpoint`, `primaryStatus`, `primaryReason`, `fallbackStatus`, `resolvedBy` and `migrationGap`. The code notes `figmaId` is hashed by `track()`, but hashed identifiers remain pseudonymous personal data, and `migrationGap` is explicitly designed to be traceable back to an individual account, which expands the data categories held by that processor.
- **Recommended action**: Before merging, extend the analytics subprocessor ROPA entry with this event and its fields, and set an explicit end date for the event tied to the migration cutover.

**[Access Control] Plan and entitlement resolution now fails over across two independent licence datastores**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` (lines 124–149); `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` (lines 13–21)
- **Implication**: A non-200, a timeout, an unparseable body, or a 200 the `isComplete` guard rejects on the primary now causes the legacy host to decide the user's plan and entitlements. Entitlement decisions therefore have two authorities during the migration window, and a stale or un-revoked record on the legacy host can grant access the primary has already withdrawn.
- **Recommended action**: Before merging, confirm the legacy host's revocation state is kept in step with the primary for the migration window, and after merging track the `migrationGap` metric to a documented cutover decision.

**[Access Control] Seat detach fails over, so a licence can remain attached on the host that did not process it**
- **File**: `packages/tokens-studio-for-figma/src/utils/removeLicense.ts` (lines 5–17)
- **Implication**: The PR itself documents that a `detach-license` accepted by the primary leaves the seat attached on the legacy host. That is a de-provisioning gap: a user or organisation that removes a licence may still be counted as entitled on the other system until a delta sync runs, which weakens the removal-of-access control expected under SOC 2 CC6.
- **Recommended action**: Before merging, agree and record the delta-sync step as a required, owned task with a completion date rather than a note in the PR description.

**[Policy Gap] Per-call error reporting narrowed to the "all hosts unreachable" case**
- **File**: `packages/tokens-studio-for-figma/src/utils/validateLicense.ts` (lines 26–28 removed); `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` (lines 152–157)
- **Implication**: The per-helper `captureException` is replaced by a single `captureMessage` raised only when every host fails at transport level. Non-transport failures on the primary — including the "incomplete payload" case that would otherwise silently downgrade a paying user — no longer produce an error record, so the only remaining signal for those cases is an analytics event rather than the monitoring pipeline.
- **Recommended action**: After merging, add an alert on the `License API fallback` event with `primaryReason: 'incomplete'` so entitlement-affecting failures remain visible in monitoring.

</details>

anathaniel-TS:
🛡️ **Hyma Compliance Check**

⚠️ **2 finding(s) across 2 file(s)**

| 🟡 MEDIUM | Access Control | `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` | License credentials + user PII now sent to a new primary host with legacy failover |
| 🟢 LOW | Subprocessor | `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` | New "License API fallback" analytics event emits hashed figmaId + migration metadata |

<details>

**[Access Control] License credentials + user PII now sent to a new primary host with legacy failover**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApiUrl.ts` (lines ~24-45); consuming logic in `packages/tokens-studio-for-figma/src/utils/licenseApi.ts`, `validateLicense.ts`, `getLicenseKey.ts`, `removeLicense.ts`
- **Implication**: License/entitlement validation now routes to a new primary endpoint `api-production.tokens.studio` first, falling back to the legacy `licence.tokens.studio`. Both hosts run separate datastores and receive `licenseKey`, `userId`, and `userName` (and `validate-license` returns `email`/`plan`/`entitlements`). This is an endpoint/host failover that changes where credentials and personal data are sent, and takes effect on deploy. The new host/datastore processing license-holder PII should be reflected in the ROPA and any subprocessor/DPA records before go-live.
- **Recommended action**: Confirm `api-production.tokens.studio` and its datastore are covered in the ROPA register and (if operated/hosted by a third party) on the subprocessor list + privacy policy; verify license-holder PII (userId/userName/email) processing on the new host is documented and the legacy host is decommissioned per the migration plan.

**[Subprocessor] New "License API fallback" analytics event emits hashed figmaId + migration metadata**
- **File**: `packages/tokens-studio-for-figma/src/utils/licenseApi.ts` (lines ~96-120, `reportFallback`)
- **Implication**: A new `track('License API fallback', …)` event sends endpoint, primary/fallback status, a `migrationGap` flag, and a hashed `figmaId` to the existing analytics subprocessor (Mixpanel). This is an expansion of data sent to an existing subprocessor; `figmaId` is passed through the existing `track()` hashing but still links a migration gap to an account.
- **Recommended action**: Confirm the new event and the (hashed) figmaId field are within the scope of the existing Mixpanel processing entry/ROPA; no new subprocessor appears to be introduced.

</details>

PR #3959, opened 2026-08-22
fix: styles marked changed and re-publishable after a no-op re-export

six7:
## Summary
- Re-exporting styles (e.g. from a theme) with no token changes was marking color/effect/typography styles as modified in Figma, so the Publish dialog kept showing them as "ready to publish" even though nothing actually changed.
- Root cause: `unbindVariableFromTarget` unconditionally reassigned a color style's `paints` on every export — even when there was no bound variable to remove and the value already matched — via `setColorValuesOnTarget`, before the existing `isPaintEqual` guard even ran.
- Applied the same fix to the other unconditional-write spots found in the same code paths: `setEffectValuesOnTarget` (effects array, using `isEffectEqual` plus a `boundVariables` check), `tryApplyTypographyCompositeVariable` (fontSize/lineHeight/letterSpacing/etc.), `setFontStyleOnTarget` (fontName), and the `description` field writes in the color/text/effect style setters.
- Verified live against the real plugin in a debug Figma instance: before the fix, re-exporting an unchanged theme flipped several styles from `CURRENT` to `CHANGED` in Figma's own publish-status API even though the paint values were byte-for-byte identical; confirmed fixed after these changes.

## Test plan
- [x] `yarn lint` on all touched files
- [x] `yarn jest src/plugin` (85 suites / 506 tests) and `yarn jest src/utils` (87 suites / 442 tests) — all passing, no regressions
- [x] Added regression tests asserting a second export with an unchanged token does not reassign `paints`/`effects`/typography properties (reference-identity / write-count checks)
- [x] Manually reproduced the original bug end-to-end in a live Figma instance (export → publish → export again with zero changes → Figma reported changes) and confirmed the fix resolves it

### DIFF
--- a/.changeset/fix-styles-marked-changed-on-reexport.md
+++ b/.changeset/fix-styles-marked-changed-on-reexport.md
@@ -2,4 +2,4 @@
 "@tokens-studio/figma-plugin": patch
 ---
 
-Fix styles being marked as changed and needing republish in Figma after exporting again with no actual token changes. The main cause was `unbindVariableFromTarget` unconditionally rewriting a color style's paint on every export, even when there was no bound variable to remove and the value already matched. Effect, typography, and style-description writes now skip the same way when the resolved value hasn't changed.
+Fix styles showing up as changed in Figma's Publish dialog after exporting again, even when nothing was actually changed. Re-exporting styles (e.g. from a theme) should be a no-op when your tokens haven't changed.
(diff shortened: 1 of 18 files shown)


anathaniel-TS:
🛡️ **Hyma Compliance Check**

9 file(s) scanned · 0 findings

PR #3960, opened 2026-08-27
fix(typography): match numeric weights to family styles

mck:
## Summary

Numeric font-weight `400` was applied as `Regular`/`Normal` only. Families like Gotham SSm have **Book** instead, so apply failed with `Error setting font family/weight combination for GothamSSm/400`.

- Invert the sd-transforms weight map (plus Roman/Buch) into Figma style candidates.
- On apply, `listAvailableFontsAsync()` ∩ those candidates (case-insensitive, keep Figma’s canonical style string).
- Fallback: try candidates via `loadFontAsync` if the font list has no hit.

Companion Platform PR: https://github.com/The-Phoenix-Will-Fly/studio-on-rails/pull/4867

## Test plan

- [x] `jest src/plugin/figmaTransforms/fontWeight.test.ts src/plugin/setFontStyleOnTarget.test.ts`
- [ ] Apply typography with Gotham SSm + weight 400 → Book, no toast
- [ ] Apply Inter + 400 → Regular
- [ ] Weight token `Book` still applies


anathaniel-TS:
🛡️ **Hyma Compliance Check**

5 file(s) scanned · 0 findings

PR #3477, opened 2025-07-30
Fix number tokens display in json

akshay-gupta7:
### Why does this PR exist?

Closes #3465 

### What does this pull request do?

Currently number tokens are displayed as strings in the json file export/editor
According to DTCG format, it should be displayed in numeric format without quotes, this PR addresses that concern and will now be displayed with appropriate format.

### Testing this change

- Create a numeric token
- Go to JSON editor and check how is it being displayed

### Additional Notes (if any)

Screenshot 2025-07-31 at 11 55 52 AM

PR #3964, opened 2026-09-02
fix: serialize number and fontWeights tokens as JSON numbers, not strings

### LINKED ISSUE #3465: Number Tokens are Incorrectly Saved as Strings When Created via the Plugin
**Describe the bug**
When a number token is created using the plugin, it is incorrectly saved as a ⁠string in the JSON output `⁠"value": "16"`. According to the (DTCG) specification, number tokens should be stored as ⁠number types `"value": 16`. If a token is however manually added to the JSON file without quotation marks, the system correctly interprets it as a number. 

**To Reproduce**

**Expected behavior**
Number tokens need to be stored and pushed as `Numbers` and not as `Strings`

Creating number tokens within the plugin:
Image

Adding them via the JSON:
Image

con-cis:
### Why does this PR exist?

Tokens Studio serializes `number` and `fontWeights` tokens with a literal numeric value
(e.g. `"700"`) as JSON strings instead of numbers, violating the DTCG spec. #3465 was
originally addressed by #3477, but that fix only ever covered `TokenTypes.NUMBER` — the
`fontWeights` half was never in scope, not a regression. Separately, `processNumberValue()`
still returns the raw empty string for an empty/whitespace-only value instead of `undefined`,
so an empty number-typed value gets serialized back as `""`.

Closes #3465

- `stringifyTokens.ts`: `processTokenValue()` now applies numeric coercion to
  `TokenTypes.FONT_WEIGHTS` as well as `TokenTypes.NUMBER`.
- `processNumberValue.ts`: returns `undefined` instead of `''` when the trimmed value is
  empty, so `JSON.stringify` drops the key instead of emitting `"value": ""`.

### Testing this change

`yarn test` (runs `turbo run test`) — updated `processNumberValue.test.ts` and
`stringifyTokens.test.ts` cover both fixes; the new/updated assertions fail against the
pre-fix code and pass with this change.

### Additional Notes (if any)

The `undefined`-for-empty-value behavior is a judgment call — open to feedback if
maintainers would rather keep the raw (empty) value untouched instead of dropping the key.


anathaniel-TS:
🛡️ **Hyma Compliance Check**

PR #3969, opened 2026-09-03
feat: add motion, duration, and cubicBezier token types

akshay-gupta7:
### Why does this PR exist?

Closes #0000 

Adds first-class support for the DTCG motion token family — `duration`, `cubicBezier`, and `transition` (the composite formerly called `motion` in early drafts) — so token sets that describe animation timing round-trip cleanly between JSON, Studio, and Figma. Duration and cubicBezier become native Figma **TIMING** and **EASING** variables where the runtime allows it, with a graceful fallback to STRING/FLOAT for older Figma builds.

**New token types (all users, DTCG-aligned)**
- Adds `TokenTypes.DURATION` (`"duration"`), `TokenTypes.CUBIC_BEZIER` (`"cubicBezier"`), and `TokenTypes.MOTION` (wire value `"transition"`, per DTCG canonical).
- Value shapes: `TokenDurationValue = { value, unit: 'ms' | 's' }`, `TokenCubicBezierValue = [x1,y1,x2,y2]`, `TokenMotionValue = { property?, duration, timingFunction, delay? }`.
- Corresponding `SingleXToken` types, `is/` guards, and union wiring in `SingleToken`.
- DTCG import normalization accepts `duration`, `cubicbezier`, `motion`, and `transition`.
- Resolver + alias paths handle the composite; `TokenResolver.calculateTokenValue` skips math eval for cubicBezier (previously `"0.4, 0, 0.2, 1"` collapsed to `1` via the comma sequence operator).
- Reference-token picker allows DURATION↔NUMBER and self-refs for the other two.
- Composition token value union widened.

**Figma variable export**
- New **Motion** checkbox in the Manage Styles & Variables options (i18n added for all 6 locales).
- `variablesMotion` setting plumbed through Redux, notifiers, message types, uiSettings, setUi, defaults `true`.
- Gate wired into `checkIfTokenCanCreateVariable`; the three `shouldCreateVariables` / `checkSetting` short-circuits updated so motion-only exports aren't dropped.
- `tokenTypesToCreateVariable` allowlist widened.
- `DURATION → TIMING`, `CUBIC_BEZIER → EASING` in `convertTokenTypeToVariableType`.
- New `figmaTransforms/motion.ts` (`timingToSeconds`, `toFigmaEasing`, `defaultFigmaEasing`) — ported from studio-on-rails so both plugins share one canonicalization.
- New `setMotionValuesOnVariable.ts` writers.
- Runtime feature-flag probe: creates+removes a throwaway EASING variable per collection and caches the result — falls back to STRING (`"0.4, 0, 0.2, 1"`) or FLOAT (ms) when the current Figma build hasn't opened up plugin motion variable creation.
- Type-mismatch handling: for TIMING/EASING, removes and recreates existing variables of the wrong type (Figma disallows in-place `resolvedType` changes; earlier STRING placeholders would otherwise be stuck).

**Studio compatibility**
- Wire value for the composite is `"transition"` (DTCG canonical, matches Studio's `tokenscript_registry`).
- Inbound normalizers accept both `motion` and `transition`; legacy JSON silently migrates on save.

**Dependency**
- Bumps `@figma/plugin-typings` `^1.96.0 → ^1.135.0` for `EASING` / `TIMING` / `MotionEasing`. Two unrelated typings-fallout fixes (SlotNode `@ts-expect-error` gone stale; ShaderEffect added to the `Effect` union has no `boundVariables`).

### Testing this change

**Unit tests (44 pass)**
```
yarn workspace @tokens-studio/figma-plugin jest \
  src/plugin/figmaTransforms/motion.test.ts \
  src/plugin/figmaTransforms/duration.test.ts \
  src/utils/convertTokenTypeToVariableType.test.ts \
  src/utils/checkIfTokenCanCreateVariable.test.ts \
  src/utils/__tests__/uiSettings.test.ts
```

**Manual — end-to-end variable export**
1. Load a token set with:
   ```json
   {
     "fast":    { "$type": "duration",    "$value": "200ms" },
     "slow":    { "$type": "duration",    "$value": "0.4s" },
     "easeOut": { "$type": "cubicBezier", "$value": "0.4, 0, 0.2, 1" }
   }
   ```
2. Open the Export → Options modal; confirm the new **Motion** checkbox is present and on by default.
3. Run "Create variables".
4. In Figma's Variables panel:
   - `fast` → **Number (Time)** = `200ms`
   - `slow` → **Number (Time)** = `400ms`
   - `easeOut` → **Easing** = `CUSTOM_CUBIC_BEZIER(0.4, 0, 0.2, 1)`

   If your Figma build hasn't opened plugin-side motion variables yet, you'll instead see `fast/slow` as **Number** in ms and `easeOut` as **String** `"0.4, 0, 0.2, 1"`.
5. Untick the Motion checkbox and re-export a motion-only set: verify **no** variables are created (previous behavior silently no-op'd because other-type checkboxes short-circuited the whole path).
6. Change `easeOut`'s value in the JSON and re-export: verify the existing EASING variable updates (or, if you started on the STRING fallback path and later gained EASING support, the variable is transparently deleted and recreated as EASING).


packages/tokens-studio-for-figma/src/constants/TokenTypes.ts
```diff
@@ -40,3 +46,10 @@ export const ExportNumberVariablesTokenTypes = [
   TokenTypes.PARAGRAPH_SPACING,
   TokenTypes.OPACITY,
 ];
+
+// Motion-related token types gated by the "Motion variables" export setting.
+// Duration -> FLOAT, CubicBezier -> STRING. Motion itself is composite and skipped.
+export const ExportMotionVariablesTokenTypes = [
+  TokenTypes.DURATION,
+  TokenTypes.CUBIC_BEZIER,
+];
```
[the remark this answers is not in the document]
akshay-gupta7: resolved in https://github.com/tokens-studio/figma-plugin/pull/3969/commits/315b2c409e5fdf118219ef07678a1510be50a06a

packages/tokens-studio-for-figma/src/app/store/utils/updateAliasesInState.ts
```diff
@@ -19,7 +19,13 @@ export function updateAliasesInState(tokens: Record<string, AnyTokenList>, data:
           // Update if token is of type array, e.g. box shadows
           if (Array.isArray(newToken.value)) {
```
six7, line 20: **Array-valued `cubicBezier` tokens get wiped to `[{},{},{},{}]`**

This branch assumes every array element is a box-shadow object. A DTCG cubicBezier stored as `[0.4, 0, 0.2, 1]` is an array of numbers, and `Object.entries(0.4)` is `[]`, so the value is rewritten to `[{}, {}, {}, {}]`. `updateAliases` runs over every token in every set on any rename, so renaming an unrelated token is enough to trigger it.

`trimValue.ts` has the same flaw in its array branch, so saving the cubicBezier token after only changing its description corrupts it too.

The bug predates this PR, but the PR makes the array tuple a first-class value shape (`TokenCubicBezierValue`), so it undercuts DTCG round-tripping.

**Repro (local storage)**
1. In the JSON view, paste:
   ```json
   {
     "easing": { "standard": { "value": [0.4, 0, 0.2, 1], "type": "cubicBezier" } },
     "spacing": { "sm": { "value": "4", "type": "spacing" } }
   }
   ```
2. Rename `spacing.sm` to `spacing.small`.
3. The JSON view now shows `easing.standard.value: [{}, {}, {}, {}]`. Exporting variables logs `Could not parse cubicBezier value`.

Suggested fix: only run the object-reduce on array elements that are objects, and pass primitives through (or map them through `replaceReferences(String(v))`). Apply the same change in `trimValue`.

packages/tokens-studio-for-figma/src/app/store/remoteTokens.tsx
```diff
@@ -250,9 +250,18 @@ export default function useRemoteTokens() {
...
                     ...context,
                     changeSetId: remoteData.metadata.changeSetId,
+                  };
+                  dispatch.uiState.setApiData(nextCredentials);
+                  // Also persist to Figma pluginData so subsequent pushes have
+                  // the changeSetId available after a plugin reload — Redux
+                  // alone would drop it. Without this, the very first push
+                  // after reload hits 422 "change_set_id is required".
+                  AsyncMessageChannel.ReactInstance.message({
+                    type: AsyncMessageTypes.CREDENTIALS,
```
six7, line 263: **This save doesn't do anything for OAuth credentials**

`updateCredentials` (`src/utils/credentials.ts`) returns early when `context.provider === StorageProviderType.TOKENS_STUDIO_OAUTH`, so this `CREDENTIALS` message never saves `changeSetId`. After a plugin reload, `changeSetId` is still missing. The new early return in `pushToTokensStudioOAuth` then skips the first push and shows the "Cannot sync" toast, which is the case the comment above says this change fixes.

Either save `changeSetId` through a path that handles OAuth credentials, or drop this message and the comment so they don't suggest it's covered. These `changeSetId` changes are also unrelated to motion tokens, so a separate PR might be cleaner.

six7 (approved):

Issue #3791, opened 2026-02-26
Themes JSON variable order and group name inconsistency after creating new tokens and variables


akshay-gupta7:
**Describe the bug**
When a user creates new tokens in the plugin and subsequently creates variables in Figma, the themes.json file is updated to include the new variable IDs as expected. However, the order of existing variables in the theme JSON is sometimes shuffled, and in some cases, the group names are also unexpectedly modified. This can lead to confusion and difficulties in tracking theme changes over time.

**To Reproduce**
- Open the plugin and create a set of new tokens.
- Export these token, leading to new variables in Figma
- Observe that themes.json correctly receives the new variable IDs, but check the order and grouping of other variables.
- Notice that previously existing variables may have changed order and/or group names.

**Expected behavior**
Adding new tokens and then variables should update themes.json only by adding the new variable IDs, without reordering other existing variables or altering their group names. The original order and grouping should be preserved.

**Screenshots or Screencasts**
User is using Azure sync(could be sync provider specific issue as well)
https://tokens-studio.slack.com/archives/C02DEJMS9AP/p1772024430042829?thread_ts=1764701050.847729&cid=C02DEJMS9AP

**Figma file (optional)**
Send us the link to a Figma file containing a reproducible example

**JSON (optional)**
If possible, attach the JSON that leads to this bug.

PR #3977, opened 2026-09-17
Akshay/fix themes file issue 3791

akshay-gupta7:
### Why does this PR exist?

Closes #0000


anathaniel-TS:
🛡️ **Hyma Compliance Check**

akshay-gupta7:
# Code Review — PR #3977

**Target:** `fix: preserve themes.json variable order on export (#3791)`
**Effort:** xhigh · **Scope:** `main...pr/3977` (3 files, +60/−9)

---

## 1. Stale token entries persist in `$figmaVariableReferences` — correctness

`packages/tokens-studio-for-figma/src/app/store/models/reducers/tokenState/assignVariableIdsToTheme.ts:11`

Merging without pruning leaves references to deleted/renamed tokens — and dangling pointers to removed Figma variables — in `themes.json` forever. The previous wholesale replace scrubbed them on every export.

**Failure scenario**

A theme has tokens `{colors.red, colors.old}` and the setting `removeStylesAndVariablesWithoutConnection` is enabled. The user deletes `colors.old` (or renames it to `colors.crimson`) and re-exports:

1. `updateVariables.ts:112-120` removes the underlying Figma variable.
2. `setValuesOnVariable` produces an incoming map containing only `{colors.red}` (or `{colors.red, colors.crimson}`).
3. The reducer merges that on top of the old map — so `themes.json` still contains `"colors.old": "<key of deleted variable>"`, a dangling pointer.

In the rename case the file ends up holding **both** the stale `colors.old` and the new `colors.crimson`.

**Suggested fix** — preserve order *and* prune, by walking the previous map first and keeping only keys still present in `incoming`:

```ts
const prev = theme.$figmaVariableReferences ?? {};
const merged: Record<string, string> = {};
Object.keys(prev).forEach((key) => {
  if (key in incoming) merged[key] = incoming[key];
});
Object.keys(incoming).forEach((key) => {
  if (!(key in merged)) merged[key] = incoming[key];
});
```

This keeps existing entries in their original position (the point of #3791) while dropping tokens that no longer exist.

---

## 2. `Object.keys().forEach` can be a single spread — simplification

`packages/tokens-studio-for-figma/src/app/store/models/reducers/tokenState/assignVariableIdsToTheme.ts:11`

`{ ...prev, ...incoming }` produces identical key ordering and values, so the explicit loop and the separate `merged` declaration are redundant.

**Cost** — four extra lines doing what `{ ...(theme.$figmaVariableReferences ?? {}), ...incoming }` says in one, with the same insertion-order semantics (duplicate keys keep their earlier position; the later value wins). Maintenance noise, not runtime.

> Note: this is moot if finding #1 is applied, since the pruning version needs an explicit loop anyway.

---

## 3. Comment references issue #3791 — conventions

`packages/tokens-studio-for-figma/src/app/store/models/reducers/tokenState/assignVariableIdsToTheme.ts:9`

`~/.claude/CLAUDE.md` states:

> Don't reference the current task, fix, or callers ("used by X", "added for the Y flow", "handles the case from issue 123"), since those belong in the PR description and rot as the codebase evolves.

The comment ends with `(issue #3791)` — an exact match for the disallowed pattern.

**Suggested fix** — trim to the invariant only:

```ts
// Merge so existing entries keep their original key order and only new tokens append.
```

The issue link already lives in the changeset and the PR description.

---

## Not flagged

- **Skipped-theme path** (`if (!variableIds[theme.id]) return theme;`) is already covered — the existing test's `dark` theme has no matching entry and is asserted unchanged.
- **`?? {}` fallback** is legitimate: a freshly created theme genuinely has no `$figmaVariableReferences` yet.
- **Group-name mutation** from the original issue report — no code path writes `theme.group` during export; `createNecessaryVariableCollections.ts:21` renames the Figma collection, not the theme.

anathaniel-TS:
🛡️ **Hyma Compliance Check**

PR #3980, opened 2026-09-23
fix: token value tooltips late or missing on large token sets

### LINKED ISSUE #3955: Token value tooltips intermittently fail to appear on hover in projects with large token sets
**Describe the bug**
Hovering a token doesn't show its value tooltip. It's intermittent, and it correlates
with the size of the token set — not reproducible on a small set, consistently
reproducible on a large one.

This is separate from #3891, which was fixed in 2.11.10. Confirmed as a distinct issue
with Akshay. Two customers have reported the residual behaviour since that fix:

- Adam Lowe, 23 Jul, immediately after 2.11.10 shipped: "It looks to largely have fixed
  it. There seems to be some instances when it won't show it. But not anything that I
  can replicate consistently." (screen recording in thread)
- Erwin Kreuze, 11 Aug, on v2.11.12: still present, screen recording in thread

Reproduced internally on 17 Aug: no occurrence in a file with a handful of tokens, but
consistent in a file with ~6,000 tokens. The token count appears to be the variable
rather than anything specific to a customer's setup.

**To Reproduce**
1. Create a file with a large token set (~6,000 tokens reproduces it reliably; a set of
   a few dozen does not)
2. Open the plugin and hover over tokens in the list
3. The value tooltip fails to appear on a proportion of hovers

**Expected behavior**
The value tooltip appears on hover regardless of how many tokens are in the set.

Internal repro (17 Aug, ~6,000 tokens): [attach Screen Recording 2026-08-17 at 6.01.30 PM.mov]

https://github.com/user-attachments/assets/e630c7e8-e76e-443a-beab-4b7e29c33ede

Customer reports:
- Erwin, v2.11.12: https://tokens-studio.slack.com/archives/C02DLQXNY6Q/p1786448659639149
- Full thread including Adam's post-2.11.10 report:
  https://tokens-studio.slack.com/archives/C02DLQXNY6Q/p1784028015951099

**Figma file (optional)**
n/a — reproduces on any file with a large enough token set

**JSON (optional)**
Reproduced with a generated 6,352-token set; happy to attach if useful.

Plugin version: 2.11.12 (customer), reproduced internally on the current build.

six7:
### Why does this PR exist?

Closes #3955

On large token sets (~6k tokens) the value tooltip often shows up late or not at all. Two things re-render the entire token list:

1. **Hover (since Dec 2023).** Each token has its own Radix `Tooltip.Root`, and all of them share the app-wide `Tooltip.Provider` in `startup.tsx`. Radix 1.0.7 keeps the provider's skip-delay flag (`isOpenDelayed`) in React state and flips it on the first open and again 300 ms after the last close. Each flip re-renders every mounted tooltip, including the ones in collapsed listings (they're `display: none` but still mounted). With ~7k tooltips that blocks the main thread for ~400 ms (up to 740 ms), twice per hover session. Radix moved this flag to a ref in 1.2.0.
2. **Selection updates (regression from #3784, shipped in 2.11.1).** Every `MoreButton` subscribes to `mainNodeSelectionValues` to pick the property to deselect on click. The plugin re-sends selection after every token apply (`setNodeData` → `sendSelectionChange`) and on canvas selection changes, so each of those now re-renders all token buttons and their tooltips: ~1.37 s per update vs ~0.30 s on 2.11.0. The hover right after an apply is the one that misses.

The watchdog from #3919 isn't involved; removing it made latency worse.

- `TokenTooltip` wraps each token tooltip in its own `Tooltip.Provider`, so a skip-delay flip only re-renders that one tooltip. Token tooltips use `delayDuration={0}` and non-hoverable content, so the shared provider's skip-delay and grace-area coordination did nothing for them. The document-level `tooltip.open` event still keeps a single tooltip open at a time, and the watchdog keeps working.
- `MoreButton` reads `mainNodeSelectionValues` from the store when the token is clicked instead of subscribing to it. The deselection behaviour from #3784 is unchanged.
- Tests:
  - `TokenTooltip.test.tsx`: opening one token tooltip must not re-render another token's trigger.
  - `MoreButton.test.tsx`: a selection change that doesn't involve the token doesn't re-render it, and a click uses the latest selection.
  - The two re-render tests fail on `main` and pass here.

No dependency changes. Upgrading `@radix-ui/react-tooltip` to 1.2+ would also fix the first issue, but `@tokens-studio/ui` 0.9.0 resolves its own copy, and both would have to dedupe to the same version or the provider context splits.

### Testing this change

Measured with the browser preview build (`--PREVIEW_ENV=browser`, production mode) in headless Chrome, a generated 6,352-token set and real mouse input over CDP (60 hovers per run, resting 250–900 ms on each token):

| Scenario | `main` | This PR |
|---|---|---|
| Hovering only: hovers with no tooltip | 8% / 15% (two seeds) | 2% / 0% |
| Hovering only: long tasks per run | 68–80, ~420 ms each (max 740 ms) | 17–23, ~60 ms each (max 100 ms) |
| Main thread blocked per selection update | ~1.37 s | ~0.36 s |
| Hover → apply → hover: hovers with no tooltip | 40% | 0% |

For reference, 2.11.0 (Feb 2026) measured 12% misses when only hovering, 5% in the hover → apply loop, and ~0.30 s per selection update.

Manual check in Figma:
1. Open a file with a few thousand tokens.
2. Select a layer and hover tokens: tooltips should appear right away.
3. Click a token to apply it and immediately hover the next one: its tooltip should appear without a pause.

`yarn test` passes (296 suites, 1,749 tests), eslint is clean and `tsc -p tsconfig.build.json` reports no errors.

### Additional Notes (if any)

- The numbers come from headless Chrome on an M-series Mac, not from Figma. Absolute times inside Figma will be higher; the before/after gap should hold. The measurement scripts aren't part of this PR.
- Possible follow-ups: one shared tooltip for the token list instead of a Radix root per token, and not mounting collapsed token-type listings.


anathaniel-TS:
🛡️ **Hyma Compliance Check**

2 file(s) scanned · 0 findings

akshay-gupta7 (approved):

PR #3981, opened 2026-09-23
chore(ci): remove SonarQube scanning

six7:
### Why does this PR exist?

There's no tracking issue for this. The SonarQube check fails on every recent PR, for example #3911, #3977, #3979 and #3980, before it analyses any code. The scanner log says:

> Not authorized or project not found. Please check the 'SONAR_TOKEN' environment variable, the 'sonar.projectKey' and 'sonar.organization' properties

The team decided to drop SonarQube rather than fix the token. Until this lands, every PR shows a red check that has nothing to do with its code.

### What does this pull request do?

Removes SonarQube from the repo:

- Deletes the `sonarqube` job from `.github/workflows/node.js.yml`. Nothing depends on it, since no workflow under `.github/` uses `needs:`.
- Deletes `.github/workflows/sonarqube.yml`, the manually triggered "Sonarqube manual" workflow.
- Deletes `sonar-project.properties`.

`git grep -i sonar` now returns nothing, lockfiles included.

This only touches CI and doesn't change the published plugin package, so there's no changeset.

Removing the check doesn't affect merging. `main` has no required status checks, and the "Force checks to pass" ruleset, id 62640, is disabled.

### Testing this change

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/node.js.yml'))"` loads the workflow without errors.
- Compared with `main`, the parsed workflow has lost only the `sonarqube` job. The other five jobs and the top-level `name`, `on` and `env` keys are identical.
- The new `node.js.yml` matches the first 135 lines of the old one byte for byte, so the diff is a pure deletion.
- The checks on this PR should show no SonarQube entry, because `pull_request` runs use the workflow file from the PR branch.

### Additional Notes (if any)

Follow-ups for a repo admin. They live outside the codebase, so this PR doesn't do them:

- [ ] Delete the `SONAR_TOKEN` repository secret.
- [ ] Remove or unbind the project in SonarQube Cloud. The org is `tokens-studio` and the project key is `tokens-studio_figma-plugin`.
- [ ] Uninstall the SonarQube Cloud GitHub App if it's installed on this repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3982, opened 2026-09-23
test: fix flaky ImportedTokensDialog and branch-creation tests

six7:
Two tests fail intermittently in Node.js CI on PRs that don't touch related code, such as #3981, #3975 and #3977. In both cases the reported test isn't the one that's broken. This PR fixes each at the root. It doesn't raise any timeouts or loosen any assertions.

## 1. `ImportedTokensDialog`: failures leaking in from earlier tests

Six tests in the file called `waitFor(...)` without `await`. Three of those assertions could never pass:

- **"should create all tokens":** expects `headline` without `$extensions`.
- **"should update a token":** expects `light` without `$extensions`.
- **"should update all tokens":** expects both tokens without `$extensions`.

The reducers always write those `$extensions`: `createMultipleTokens` and `editMultipleTokens` via `updateTokenPayloadToSingleToken(…, uuidv4())`, and `editToken` too.

Since nothing awaited those promises, the tests themselves passed right away. Each `waitFor` kept polling and then rejected after its 1000 ms timeout. jest-circus attaches an unhandled rejection to whichever test is running at that moment (`state.currentlyRunningTest.errors.push(…)` in its `error` event handler).

- **Locally** the file runs in about 350 ms. It finishes before the timeouts fire, and jsdom teardown clears the timers.
- **On CI** it took about 5 s, so a rejection lands in one of the last two tests.

Both failed attempts on #3981 show the same `headline` diff, thrown from `ImportedTokensDialog.test.tsx:166:61` (the "create all tokens" `waitFor`). One attempt reported it as "should not switch…", the other as "should switch…".

**Fix:** await all six `waitFor` calls, and correct the three stale expectations to match what the reducers produce. The already-awaited sibling tests ("should ignore a new token", "should ignore an existing token") already expect these `$extensions` on the same code paths.

I also checked whether fixtures get mutated in place when ids are assigned. They don't: `updateTokenPayloadToSingleToken`, `addIdPropertyToTokens` and `removeTokenId` all copy, and `createMultipleTokens` clones through JSON. Deep-freezing every fixture passed to `createMockStore` raises no errors.

## 2. Cypress "Branch switcher › successfully create a new branch": asserting on a transient list

This turned out not to be a race between branch creation and reopening the popover. I timestamped the fetches and DOM changes, and creation always finishes first. The open create-branch dialog is a modal, which sets `pointer-events: none` on `<body>`, so Cypress can't click the trigger until `onSuccess` has closed it. After that:

1. The popover opens with the list that `onSuccess` set: `[main, development, new-branch]`.
2. Opening the popover calls `fetchBranches` (`BranchSelector.handleOpenChangePopover`) and replaces the list with the response.
3. The mocked `GET …/branches` always returned `[main, development]`. About 20 ms later `new-branch` was gone, and it stayed gone.

So the assertion only passed if Cypress queried the DOM inside that window, which depends on how the mock's round-trip time compares with Cypress's command latency. Locally the unchanged test failed 7/7. With 250 ms of fetch latency added, which widens the window, the same unchanged test passes.

Waiting for creation to finish doesn't help on its own. With the old mock, explicitly waiting for `@createRef`, the trigger label and the dialog to close still failed 3/3.

**Fix:**

- **`MockEnv`:** the mocked branch list is now stateful. `POST /git/refs` adds the created branch and `GET /branches` returns the current list, as GitHub does. The state resets per test because `MockEnv()` runs in `beforeEach`.
- **The test:** it waits until the trigger shows `new-branch`, which signals that creation finished. Then it reopens the popover, waits for the refetch, and asserts on the list it settles on.

## Verification

**Jest**

- The file alone: 5× with coverage, plus 5 `--randomize` seeds.
- The full suite passes (295 suites, 1746 tests).
- The CI failure reproduces reliably by stalling the second-to-last test for 1.1 s. With the fix, the same stall passes.
- A trailing 3 s canary test catches no leaked errors on the fixed file; on the original file it reports all three leaks.

**Cypress** (`build:cy` + `serve`, Electron)

- The fixed spec passes 10/10 (the original failed 0/7 locally), and 3/3 with every CPU core saturated.
- The full suite passes twice (27/27).
- The fixed test still passes with a 300 ms gap before the assertion, and with 400 ms of latency on every mocked GitHub call.
- Run against the old static branch list, the fixed test fails, so the assertion does check that the new branch is listed.

No changeset, because this is a test-only change.


anathaniel-TS:
🛡️ **Hyma Compliance Check**

0 file(s) scanned · 0 findings

akshay-gupta7 (approved):

